PR #5, opened 2017-12-22
Rework package.json to support non-bash (Ubuntu) environments.

### LINKED ISSUE #4: Build fails on ubuntu 17.10
Looks like there are a few bash-ism in the scripts, mostly of the form: `node-sass --output-style compressed $(< .build/node-sass-args) > .build/opensphere.min.css` which doesn't work if the shell isn't bash (any recent Ubuntu: https://wiki.ubuntu.com/DashAsBinSh). 

Any easy replacement is something like: ```node-sass --output-style compressed `cat .build/node-sass-args` > .build/opensphere.min.css```

Working on a PR now.

bradh:
Should resolve #4.

Tested build on "Bash on Ubuntu on Windows", which doesn't have chrome, but build worked fine up to that point.


wallw-bits:
Closed in favor of https://github.com/ngageoint/opensphere/commit/45f3158583cf212f10e4eb746509db19f14a68da using POSIX compliant `cat` but keeping the newer `$(...)` syntax.

bradh:
Can I ask why? Is there some style guide for the project that I missed?

wallw-bits:
Because it fixed more than was busted, and, as I said, I was already working on it with the Travis build so it just went in with all that. All of our projects use the `$(...)` syntax, and - while we do not have an official style guide for shell commands - we would prefer that it stay that way.

See http://mywiki.wooledge.org/BashFAQ/082 for information on why that syntax is preferred.

We are looking forward to you contributing to the project! Note that the markdown documentation is still in the works (so let me know if you find anything odd) and the API documentation is going up soon, along with the website and a live demo version.

PR #13, opened 2017-12-30
Update Cesium (1.40.0), OpenLayers (4.6.2) and OL-Cesium to match.

bradh:


wallw-bits:
@bradh, any particular reason this was closed rather than merely updated?

bradh:
Just that it was getting messy. I'll clean up and resubmit.

bradh:
@wallw-bits Do you recall what the purpose of the [attribution subclass](https://github.com/ngageoint/opensphere/blob/master/src/os/control/attribution.js) was?

The OL code changed a lot in https://github.com/openlayers/openlayers/commit/2dd8fdb5b7cf99f799df8d1c4db2d84d46d0de5a#diff-cdbb5958468189257bbf4e0956ac62e9 and this isn't going to be trivial to update.

wallw-bits:
The attribution control is the thing that provides attribution to map providers in the corner of the map. See http://openlayers.org/en/latest/examples/attributions.html. In OpenSphere it shows in the bottom-left corner.

bradh:
I know what the attribution thing is, I just wondered about why we needed the subclass here.

wallw-bits:
Killing the icons was probably the only reason.

bradh:
It looks like there was also a change to avoid checking attribution on a per-tile basis. 

In master, with the default tile sets, you always get attribution for both tile sets  (if enabled) even if there is nothing to show at that zoom level for one of the sets. 

Showing both sets of attribution separately (Source and Sources in the attribution text) is pretty ugly IMO - will see if that can be easily fixed here. Otherwise will raise a ticket...

PR #24, opened 2018-01-04
fix(measure): Use three decimal places for bearing.

### LINKED ISSUE #23: Measure tool has excess bearing precision
The measure tool (range and bearing) has three decimal digits for length, but five for bearing, which seems excessive.

Two or three for both would be more appropriate.

bradh:
Resolves #23.


wallw-bits:
@bradh, any particular reason this was closed rather than merely updated?

bradh:
Just that it was getting messy. I'll clean up and resubmit.

bradh:
Replaced by https://github.com/ngageoint/opensphere/pull/28

PR #34, opened 2018-01-05
ci(travis): add deployment for tags and branches to surge.sh

wallw-bits:


wallw-bits:
We're going to need a better way of doing credentials before we attempt automated PR deployments. Until then, this should handle master and releases.

PR #37, opened 2018-01-06
refactor(measure): Update measure tool UI implementation.

bradh:
Replaces deprecated Action / ActionManager implementation.

### DIFF
--- a/src/os/ui/measurebutton.js
+++ b/src/os/ui/measurebutton.js
@@ -3,10 +3,11 @@ goog.provide('os.ui.measureButtonDirective');
 
 goog.require('os.interaction.Measure');
 goog.require('os.metrics.keys');
-goog.require('os.ui.MenuButtonCtrl');
 goog.require('os.ui.Module');
-goog.require('os.ui.action.Action');
-goog.require('os.ui.action.ActionManager');
+goog.require('os.ui.menu.Menu');
+goog.require('os.ui.menu.MenuButtonCtrl');
+goog.require('os.ui.menu.MenuItem');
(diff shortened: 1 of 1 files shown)


src/os/ui/measurebutton.js
```diff
@@ -163,7 +174,7 @@ os.ui.MeasureButtonCtrl.prototype.getMeasureInteraction_ = function() {
 
 
 /**
- * @param {os.ui.action.ActionEvent} evt The event from the menu
+ * @param {os.ui.menu.MenuEvent<Array<number>>} evt The event from the menu
```
wallw-bits: Just for consistency, please make the handler type match the type provided by the menu. In this case, `os.ui.menu.MenuEvent<undefined>`.

schmidtk: The generic type here should be `undefined` to match the menu type (describes the type of the menu context).

bradh:
I see in #40 that the implementation was moved into `src/os/ui/menu/`. 

Should that change also be made here? That is, should `src/os/ui/measurebutton.js` be moved to be `src/os/ui/menu/measurebutton.js`?

wallw-bits:
You can if you like but I don't really have an opinion on it either way. Our package structure is a gigantic mess due to some moves/merges of other projects in the past.

bradh:
Incorporated change to type, did not move at this time. Force pushed an updated tree with the changes.

schmidtk:
@bradh I moved the query area import menu because it was an isolated menu under `os.action` ("actions" are the old menu API). Yours is part of a controller/UI component so it doesn't need to move.

PR #42, opened 2018-01-09
Update Cesium (1.40.0), OpenLayers (4.6.2) and OL-Cesium to match.

bradh:
The changes to `src/os/mixin/mixin.js`, `src/os/ol/mixin/rendermixin.js` and `src/plugin/heatmap/heatmapsource.js` are pretty minor - just adapting to minor implementation changes.

The attribution changes are larger, since the OL implementation was reworked in openlayers/openlayers@2dd8fdb#diff-cdbb5958468189257bbf4e0956ac62e9 

This change includes a minor cleanup such that the two attribution lists are now properly merged, so you don't see

`Sources: Esri, DeLorme, NAVTEQ, USGS, NRCAN, METI, iPC, TomTom Source: Esri, DigitalGlobe, GeoEye, Earthstar Geographics, CNES/Airbus DS, USDA, USGS, AeroGRID, IGN, and the GIS User Community `

(I hadn't really noticed the Source: in the middle before, but now can't un-see it.)

We previously had an override that ignored whether the layer was potentially rendered (for zoom level / resolution). That could make for a very long list, and wasn't accurate. It potentially was faster though. This version provides a tuning knob to enable / disable that, defaulted to doing the check (definite accuracy over potential speed).


bradh:
Reopening to trigger new test run.

bradh:
Build failure associated with:

`09 01 2018 06:25:45.002:ERROR [launcher]: Cannot start ChromeHeadless`

wallw-bits:
@bradh rebase on master and that should be fixed. See https://github.com/travis-ci/travis-ci/issues/8836#issuecomment-356362524 for details.

bradh:
@wallw-bits Thanks - I don't like to rebase branches that have been "out there" for a while, since it sometimes causes "argh!" type reactions for later changes. I prefer it for stuff that I'm reasonably confident hasn't been cloned though.

bradh:
Merged in master (using lazy github button), build seems OK.

schmidtk:
I updated a few of our Openlayers mixins so they're in parity with the original functions from 4.6.2.
- Canvas hit detection instructions changed.
- The functions on `ol.Map` were moved to `ol.PluggableMap`.
- `ol.View#fitExtent` was replaced awhile ago and I never removed the mixin for it.

PR #45, opened 2018-01-10
fix(source): fix mismatched return type

wallw-bits:
There was an inferred union type here `{Array<!(ColumnDefinition\|FeatureTypeColumn)>}` that the compiler is not detecting as a mismatch with the declared type. I opened https://github.com/google/closure-compiler/issues/2783 with the compiler. This issue was causing filters programmatically created from histogram bins to fail.

I also removed a superfluous, empty interface.

### DIFF
--- a/src/os/data/columndefinition.js
+++ b/src/os/data/columndefinition.js
@@ -1,7 +1,5 @@
 goog.provide('os.data.ColumnDefinition');
 goog.require('os.IPersistable');
-goog.require('os.parse.IColumnDefinition');
-
 
 
 /**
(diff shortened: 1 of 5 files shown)


src/os/data/columndefinition.js
```diff
@@ -1,6 +1,6 @@
 goog.provide('os.data.ColumnDefinition');
 goog.require('os.IPersistable');
-goog.require('os.parse.IColumnDefinition');
+goog.require('os.data.ColumnDefinition');
```
schmidtk: Guessing this was from search/replace, but this file is now requiring itself.

src/os/data/columndefinition.js
```diff
@@ -27,9 +26,9 @@ os.data.ColumnDefinition = function(opt_name, opt_field) {
 
   /**
    * The field
-   * @type {string}
+   * @type {!string}
```
schmidtk: `!` is implied for primitives. You also removed the `|| ''`, which would allow this value to be null/undefined if `opt_name` and `opt_field` aren't provided.

PR #49, opened 2018-01-11
Add documentation for yarn.

schmidtk:
See [README.md](https://github.com/schmidtk/opensphere/blob/yarn-docs/README.md) and [DEVELOPING.md](https://github.com/schmidtk/opensphere/blob/yarn-docs/DEVELOPING.md) for rendered markdown with these changes.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -9,19 +9,26 @@ Start by [forking](https://github.com/ngageoint/opensphere/fork) the OpenSphere 
 * git
 * Java 1.7.0+
 * Node 6+ and npm
+* [Yarn](https://yarnpkg.com/en/) (optional)
 * Python (optional)
 
 Ensure that the executables `git`, `node`, and `java` are in your `PATH`.
 
-[Clone|https://help.github.com/articles/cloning-a-repository/] your repository, and change directory into the clone.
+### Workspace Setup
 
-Install the dependencies by running
+#### Yarn
 
-    $ npm install
+Clone [opensphere-yarn-workspace](https://github.com/ngageoint/opensphere-yarn-workspace), and change directory into `opensphere-yarn-workspace/workspace`.
+
+Clone your fork, then run `yarn install`.
+
+#### NPM
+
+Clone your fork, then change directory to the clone. Run `npm install`.
 
 ### Linking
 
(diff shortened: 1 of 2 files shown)


DEVELOPING.md
```diff
@@ -9,19 +9,26 @@ Start by [forking](https://github.com/ngageoint/opensphere/fork) the OpenSphere
...
-[Clone|https://help.github.com/articles/cloning-a-repository/] your repository, and change directory into the clone.
+### Workspace Setup
 
-Install the dependencies by running
+#### Yarn
 
-    $ npm install
+Clone [opensphere-yarn-workspace](https://github.com/ngageoint/opensphere-yarn-workspace), and change directory into `opensphere-yarn-workspace/workspace`.
+
+Clone your fork, then run `yarn install`.
```
wallw-bits: I was under the impression this needed to be done in the workspace root. If so, perhaps make that more clear?

README.md
```diff
@@ -19,8 +18,21 @@ what we've done with Openlayers and Cesium but want your own UI? You can do that
...
+The build will not run natively in Windows. Perhaps try Cygwin or Linux via a VM or Docker container. It should run great on OS X and typical Linux distributions.
+
+### Yarn
+
+[Yarn](https://yarnpkg.com/en/) is recommended for dependency management over npm, particularly when using plugins and configuration projects. It will automatically link related projects and dependencies in your workspace.
+
+* Clone [opensphere-yarn-workspace](https://github.com/ngageoint/opensphere-yarn-workspace)
+* Clone OpenSphere to the `workspace` directory
+* `yarn install`
+* `npm run build`
```
wallw-bits: If we are recommending `yarn`, then why not use it to run scripts as well? `yarn run build`

DEVELOPING.md
schmidtk: That was my initial assumption, but I later realized yarn will walk up the directory structure, find the workspace root, and run install for the workspace. It doesn't matter where you run it from.

README.md
schmidtk: Either works for me. We're assuming npm is installed and those two commands are a horse apiece, so you can run either one. All of our scripts use `npm run`, so I went with that.

Issue #58, opened 2018-01-15
Convert state actions to menu API.


schmidtk:
`os.ui.state.action.manager` needs to be converted to use the `os.ui.menu.Menu` API.

bradh:
Resolved by https://github.com/ngageoint/opensphere/pull/65 ?

schmidtk:
Yep, resolved by #65. Thanks!

Issue #56, opened 2018-01-15
Convert column actions to menu API.


schmidtk:
Refactor `os.ui.slick.createColumnActions` to return a `os.ui.menu.Menu` instance, and update `slickgrid.js` to use the new menu.

schmidtk:
Resolved by #65.

Issue #50, opened 2018-01-11
Convert layer actions to menu API.


schmidtk:
`os.action.layer.manager` needs to be converted to use the `os.ui.menu.Menu` API, including all plugins that contribute to it.

bradh:
Resolved by https://github.com/ngageoint/opensphere/pull/65 ?

schmidtk:
Yep, resolved by #65. Thanks!

PR #65, opened 2018-01-17
Refactor remaining action managers to menu API.

schmidtk:
Refactors the following action managers to the menu API:
- Layer window context menu.
- Column context menu (I believe this is only used by external plugins currently).
- State button drop down.
- Import button drop down.
- Add Filter button drop down.
- Windows button drop down.


wallw-bits:
@schmidtk, does this address all of #50, #56, and #58?

schmidtk:
@wallw-bits Yes I'll close all of those now.

PR #81, opened 2018-01-24
docs(cookbook): Add logging example.

bradh:


bradh:
I'm happy for requests to refine these - they're partly an attempt to document and partly an attempt to learn, so feedback helps everyone.

wallw-bits:
For this one specifically, one handy thing to avoid copying/typing the long `LOGGER_` name every time is to assign it to `this.log_` in the constructor and then use that, which is much shorter. You'll see that pattern occasionally in OpenSphere.

However, I didn't feel that necessary to include in the cookbook example.

PR #83, opened 2018-01-25
docs(rtd): Minor guide cleanups.

bradh:
The guides are important (so want to be close to the top), but they need more introduction. 

The `json` -> `layers` change is just so you don't have to change that line later to match.

### DIFF
--- a/docs/guides/plugin_server_guide/parsing.rst
+++ b/docs/guides/plugin_server_guide/parsing.rst
@@ -34,7 +34,7 @@ The last layer is a vector format, which we will ignore in this guide. The first
   :caption: ``src/plugin/tileserver/tileserver.js``
   :linenos:
   :language: javascript
-  :emphasize-lines: 3, 5, 7, 10-13, 58-67, 71-125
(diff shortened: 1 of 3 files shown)


docs/guides/plugin_guide.rst
```diff
@@ -25,6 +18,13 @@ Data providers such as ArcGIS_, Geoserver_, and others are implemented as plugin
 .. _Geoserver: https://github.com/ngageoint/opensphere/tree/master/src/plugin/ogc
 .. _os.data.IDataProvider: https://github.com/ngageoint/opensphere/blob/master/src/os/data/idataprovider.js
 
+.. toctree::
+  :maxdepth: 1
+  :caption: Step-by-step Guides
+
+  plugin_file_type_guide/index
+  plugin_server_guide/index
+
```
wallw-bits: I'm not sure I like this move. With the File Guide, we simply removed the quick-and-dirty section on it from this page. We should probably move this back and remove the "New Server/Provider Types" section.

bradh: I'll back out that part, and have another attempt (probably bigger edit) in a separate PR.

PR #85, opened 2018-01-25
docs(cookbook): Add simple OSM example.

bradh:


docs/cookbook/osm/index.rst
```diff
@@ -0,0 +1,71 @@
...
+  {
+    "providers": {
+      "basemap": {
+        "maps": {
+          "osm": {
+            "title": "OpenStreetMap",
+            "type": "BaseMap",
+            "baseType": "XYZ",
+            "provider": "OSM",
+            "url": "//c.tile.openstreetmap.org/{z}/{x}/{y}.png",
```
wallw-bits: Does OSM have `//[a-c].tile.openstreetmap.org/`? If so, we should use more than just `c`.

I think single URL templating is currently broken in OpenSphere (and we should write an issue for that), but you can do:

```
"urls": [
  "//a.tile.openstreetmap.org/{z}/{x}/{y}.png",
  "//b.tile.openstreetmap.org/{z}/{x}/{y}.png",
  "//c.tile.openstreetmap.org/{z}/{x}/{y}.png"
]
```

bradh: Yeah, it does. I'll test templating (didn't see any code to support it), and raise the ticket as appropriate.

Will update the PR to include multiple URLs.

bradh: Actually, it looks like URL array is also a little tricky (#86). Will see if I can fix that up.

bradh: Urgh. `urls`, not `url`....

PR #92, opened 2018-01-27
fix(xyz): Add support for URLs containing server ranges.

### LINKED ISSUE #86: Tile provider URL should support templated server names and arrays of servers
Often we have a range of tile providers (e.g. OSM provides 3 - [a-c].tile.openstreetmap.org].

Ideally we'd be able to provide just the template.

bradh:
Includes updates to OSM cookbook docs.

Resolves #86.

### DIFF
--- a/docs/cookbook/osm/index.rst
+++ b/docs/cookbook/osm/index.rst
@@ -94,6 +94,8 @@ If your provider supports multiple URLs (which is the case for most OSM styles),
     }
   }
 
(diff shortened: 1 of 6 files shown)


bradh:
The PR does contain some tests: https://github.com/ngageoint/opensphere/pull/92/commits/f4a4470b93457b0fc3284ee9ec63b646984153be#diff-af52b62eeed6298ac3ed43c46cb36042

I will look to add more.

wallw-bits:
Ahh, my bad. I was expecting tests for the abstract class, since most of the additions were there.

bradh:
Should I move the tests across? Seemed like most of the similar tests were in the plugin area, and I wanted to make sure there wasn't some issue that hit later in the layer creation process.

bradh:
Didn't move the tests, but added a second commit to improve test coverage.

PR #95, opened 2018-01-28
Pelias search

### LINKED ISSUE #21: Consider removing mapzen search plugin
Will be unusable as of 20180201 - https://mapzen.com/blog/shutdown/

bradh:
Three parts:
 
- adding boundary constraints at small extent, along with a zoom-dependent focus point option.
- renaming mapzen places (not sure why places - maybe from google places?) to be pelias geocoder - resolves #21.
- adding a cookbook part for pelias search


docs/cookbook/pelias/index.rst
```diff
@@ -0,0 +1,49 @@
...
+
+.. code-block:: json
+  :caption: ``config/settings.json``
+
+  {
+    "admin": {
+      "plugin": {
+        "pelias": {
+          "geocoder": {
+            "url": "https://api.geocode.earth/v1/search?api_key=REPLACE_THIS&text={s}",
```
wallw-bits: While this will work perfectly fine, we always recommend hiding your API key behind a proxy that you control so that it is not visible to end users. That way it is less likely to be abused in some fashion.

bradh: Is this something you were planning for the deployment guide (#99)? Might it be worth a separate cookbook example?

I can see a simple pass-through handler on the server side (e.g. Apache with mod_proxy / mod_rewrite), but is this something that `os.net.ProxyHandler` can make transparent?

wallw-bits: That is a job for a simple proxypass server-side.

The purpose of `ProxyHandler` is to pass failed requests through a generic proxy to avoid mixed content and CORS issues. URLs going through it are still created client side and the user can see everything in the request payload if they are so inclined.

src/plugin/pelias/geocoder/search.js
```diff
@@ -47,32 +47,49 @@ goog.inherits(plugin.mapzen.places.Search, os.search.AbstractUrlSearch);
...
+        + '&boundary.rect.max_lon=' + extent[2];
+      url += boundaryText;
     }
   }
 
+  // Add the focus.point modifiers to the URL if enabled and we are zoomed in to at least zoom level 4.
+  var focuspoint = /** @type {boolean} */ (os.settings.get(['plugin', 'pelias', 'geocoder', 'focusPoint']));
+  if (focuspoint) {
+    var threshold = /** @type {number} */ (os.settings.get(['plugin', 'pelias', 'geocoder', 'focusPointMinZoom'], 4.0));
+    var currentZoom = /** @type {number} */ (os.MapContainer.getInstance().getMap().getView().getZoom());
```
wallw-bits: What's the reason for the cast here?

src/plugin/pelias/geocoder/search.js
```diff
@@ -47,32 +47,49 @@ goog.inherits(plugin.mapzen.places.Search, os.search.AbstractUrlSearch);
...
+    if (currentZoom >= threshold) {
+      var centre = /** @type {Array<number>} */ (os.MapContainer.getInstance().getMap().getView().getCenter());
```
wallw-bits: And also here?

wallw-bits:
Also, GitHub is showing a file conflict. I didn't have to resolve anything when I had a look locally. Maybe get it up to date with master?

src/plugin/pelias/geocoder/search.js
bradh: This one is needed:
```
/home/bradh/opensphere-yarn-workspace/workspace/opensphere/src/plugin/pelias/geocoder/search.js:88: ERROR - actual parameter 1 of ol.proj.toLonLat does not match formal parameter
found   : (Array<number>|null|undefined)
required: (Array<number>|null)
      var centreLonLat = ol.proj.toLonLat(centre, os.map.PROJECTION);
                                          ^^^^^^
```

src/plugin/pelias/geocoder/search.js
bradh: Appear to be an error on my part.

src/plugin/pelias/geocoder/search.js
wallw-bits: I'd recommend this instead, as the cast doesn't actually prevent an error:
```
var center = os.MapContainer.getInstance().getMap().getView().getCenter();
if (center) {
 ...
}
```

In other words, if the coordinate can actually be `null` or `undefined`, the cast only satisfies the compiler (you just soothe it with a little white lie). The app itself can still return those values and break your code (the `null` or `undefined` coordinate case is more common in the 3D view).

bradh: Thanks. Found another problem in the `getExtent` call above too (where the 2D map returns > 180 or < -180). Maybe can use the view there too.

docs/cookbook/pelias/index.rst
bradh: I think I've addressed the reverse proxy needs (which took way longer than it should have) in https://github.com/ngageoint/opensphere/pull/95/commits/fa373208cf5e6b0161a683a5c40881cd54f28a15.

PR #96, opened 2018-01-28
refactor(warnings): fix all missing override warnings

### LINKED ISSUE #32: Fix missingOverride warnings.
The compiler update added a new `missingOverride` check, which was set as a warning in `opensphere-build-resolver`. This warning was disabled in opensphere's `package.json`, but we should enable it and fix the warnings.

wallw-bits:
There was a code path in `os.MapContainer.addFeature()` that was accepting a 'feature' that was really an `ol.Coordinate`. The function definition does not include `ol.Coordinate` anywhere, so that path has been removed.

I also removed `externs/svg.js` in favor of the one from the compiler project.

Resolves #32


src/os/ui/ol/imap.js
```diff
@@ -31,21 +32,23 @@ os.ui.ol.IMap.prototype.addFeature;
...
  * @param {!Array<!ol.Feature>} features The features to add
+ * @param {Object=} opt_style Optional feature style to apply to all features
+ * @param {ol.layer.Layer=} opt_layer Optional layer to which to add the features
  * @return {!Array<!ol.Feature>}
  */
 os.ui.ol.IMap.prototype.addFeatures;
 
 
 /**
- * Removes a feature from the drawing layer
```
schmidtk: There is still a little variation between this interface and the implementations in `os.MapContainer` and `os.ui.ol.OLMap`. All of these functions were originally intended to abstract the drawing layer from code that needed to manage temporary features on the map.

It looks like the `opt_layer` arguments were added to reuse these functions for the search layer when that was split out from the drawing layer. I'm not very fond of that personally, and I'd prefer if `os.ui.search.FeatureResultCardCtrl` managed features directly on the search layer/source instead of routing that through the map container.

src/os/ui/search/featureresultcard.js
```diff
@@ -78,18 +97,15 @@ os.ui.search.FeatureResultCardCtrl.prototype.disposeInternal = function() {
...
-    if (features.length === 0) {
-      searchLayer.setRemovable(true);
-      mm.removeLayer(os.MapContainer.SEARCH_ID);
-    }
+  var source = this.layer.getSource();
+  source.removeFeature(this.feature);
+
+  if (!source.getFeatures().length) {
+    this.layer.setRemovable(true);
+    mm.removeLayer(os.ui.search.FeatureResultCardCtrl.SEARCH_LAYER_ID);
```
schmidtk: This works fine, but if you have a reference to the layer already in `this.layer` you may as well provide it here to avoid the lookup.

src/os/ui/search/featureresultcard.js
```diff
@@ -201,7 +219,7 @@ os.ui.search.FeatureResultCardCtrl.prototype.over = function() {
   this.feature.set(os.style.StyleType.HIGHLIGHT, os.style.DEFAULT_HIGHLIGHT_CONFIG);
   os.style.setFeatureStyle(this.feature);
   os.style.notifyStyleChange(
-      os.MapContainer.getInstance().getLayer(os.MapContainer.SEARCH_ID),
+      os.MapContainer.getInstance().getLayer(os.ui.search.FeatureResultCardCtrl.SEARCH_LAYER_ID),
```
schmidtk: You should be able to pass `this.layer` here too.

src/os/ui/search/featureresultcard.js
```diff
@@ -215,7 +233,7 @@ os.ui.search.FeatureResultCardCtrl.prototype.out = function() {
   this.feature.set(os.style.StyleType.HIGHLIGHT, undefined);
   os.style.setFeatureStyle(this.feature);
   os.style.notifyStyleChange(
-      os.MapContainer.getInstance().getLayer(os.MapContainer.SEARCH_ID),
+      os.MapContainer.getInstance().getLayer(os.ui.search.FeatureResultCardCtrl.SEARCH_LAYER_ID),
```
schmidtk: And here as well.

bradh:
Looks like travis is wedged.

PR #102, opened 2018-02-04
Upgrade OL to 4.6.4, Cesium to 1.41

bradh:
ol-cesium upgraded to match (1.34).

Includes minor attribution fixes (not previously noted) as separate commit.


src/os/olcs/imageryprovider.js
```diff
@@ -291,7 +291,7 @@ os.olcs.ImageryProvider.createCreditForSource = function(source) {
   }
 
   return (goog.isDef(imageUrl) || text.length > 0) ?
-         new Cesium.Credit(text, imageUrl, link) : null;
+         new Cesium.Credit({'text': text, 'imageUrl': imageUrl, 'link': link}) : null;
```
schmidtk: In the externs, create a `@typedef` (something like `Cesium.CreditOptions`) that defines which properties are available in this options object. Then cast the object here and remove the quoted properties. This allows the compiler to type check the values passed, while preventing them from being minified.

src/os/olcs/abstractterrainprovider.js
```diff
@@ -22,7 +22,7 @@ os.olcs.AbstractTerrainProvider = function(options) {
 
   var credit = options.credit;
   if (goog.isString(credit)) {
-    credit = new Cesium.Credit(credit);
+    credit = new Cesium.Credit({'text': credit});
```
schmidtk: The typedef mentioned in `imageryprovider.js` should be used here as well.

bradh:
@schmidtk I think I understood what you wanted - added in the latest commit. If I've got it wrong, I may need some coaching on this one...

PR #109, opened 2018-02-08
chore(sass-lint): Fix selectors single-line warnings.

bradh:
This includes some minor cleanups to make the animations more readable, which unfortunately increases the complaints about `translateY` casing (`Mixin 'translateY' should be written in lowercase with hyphens  mixin-name-format`).

### DIFF
--- a/scss/os/overrides/_bootstrap.scss
+++ b/scss/os/overrides/_bootstrap.scss
@@ -338,7 +338,8 @@ td > input {
 }
 
 @media (max-height: 800px) {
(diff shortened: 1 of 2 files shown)


schmidtk:
We're definitely open to suggestions on changes to the SASS linter rules. From the pile of linter warnings we have, it's clearly not something we've nailed down at this point but it's certainly on the agenda. The `mixin-name-format` rule could absolutely change since it conflicts with Compass mixins we rely on.

bradh:
Added some per-line suppressions for the mixin-name-format.

PR #112, opened 2018-02-08
Initial support for KML tours.

schmidtk:
Supports FlyTo, SoundCue, TourControl, and Wait elements for KML 2.2 w/ gx extension and KML 2.3.

I did not implement support for `AnimatedUpdate` yet. That's a much larger undertaking, and seems to be often used to control balloon visibility which we don't yet support.

Other changes:
- Updated the map container and Cesium camera `flyTo` calls to support both looking at the specified coordinate (KML `LookAt` element), and positioning the camera _at_ the specified coordinate (KML `Camera` element).
- Enabled tile loading on animation in 2D mode to avoid a black screen while tour FlyTo executes.
- Changed 3D camera FOV to 60 degrees to match Google Earth and Cesium default.
- Fixed an exception when calling `plugin.places.getPlacesRoot` to test if a KML layer node is the Saved Places node, when the KML tree was empty.


src/plugin/file/kml/tour/toursoundcue.js
```diff
@@ -0,0 +1,141 @@
...
+
+  return goog.Promise.resolve();
+};
+
+
+/**
+ * Play the audio file.
+ * @private
+ */
+plugin.file.kml.tour.SoundCue.prototype.playAudio_ = function() {
```
bradh: Could this just use the existing AudioManager support? Or could AudioManager be extended instead?

schmidtk: The AudioManager is for shared sounds in the application and is intended to be used for notifications. We use it in external plugins to notify things like data arriving from a streaming service. KML SoundCue's typically point to music or voice overs that relate to the tour and didn't seem to fit with the AudioManager.

schmidtk: I simplified SoundCue quite a bit by adding asynchronous execution support to `plugin.file.kml.tour.Wait` and extending that for the timeout behavior. Now it simply manages the audio clip and leaves the rest to Wait.

bradh: OK.

wallw-bits: That's fine that the sounds are not added, but I'd expect it to still obey the overall mute setting in there.

schmidtk: Thanks I hadn't considered that. Fixed and tests updated.

PR #120, opened 2018-02-09
Fix annoyances

wallw-bits:
- Fix punctuation in attribution list
- Added `os.fn.mapLayerToSource` and also adjusted `os.fn.reduceExtentFromLayers` to not care so much about the type of layer
- Fixed mismatched inferred union type
- Fixed the interface detection in `os.layer.Tile.prototype.getStyle`
- Moved generation of histogram data down to the sources so multiple types of sources can implement in various ways


src/os/layer/imageoverlay.js
```diff
@@ -505,11 +505,6 @@ os.layer.Image.prototype.supportsAction = function(type, opt_actionArgs) {
         return true;
       case os.action.EventType.RENAME:
         return !!opt_actionArgs && goog.isArrayLike(opt_actionArgs) && opt_actionArgs.length === 1;
-      case os.action.EventType.EXPORT:
-      case os.action.EventType.DISABLE_TIME:
-      case os.action.EventType.ENABLE_TIME:
-      case os.action.EventType.MOST_RECENT:
-        return false;
```
bradh: This removal (and the following part) is unrelated to the actual fix? That is, its just coincidental cleanup?

wallw-bits: Those are unused in this case and are only there because someone copied them from the vector layer.

src/os/ui/timeline/abstracttimelinectrl.js
```diff
@@ -876,34 +876,32 @@ goog.exportProperty(
...
-              label: label,
-              tooltip: tip + label,
-              icons: ['<i class="fa fa-fw fa-bars"></i>'],
-              sort: sort++,
-              handler: this.onMenuEvent.bind(this)
-            });
-          }
+    // clear the previous data actions
+    var dataGroup = menu.getRoot().find('Data');
+    goog.asserts.assert(!!dataGroup, 'Group "Data" should exist!');
```
schmidtk: This throws an exception for the Load menu, which I'm guessing is why the if check was added long ago.

I would suggest adding a Data group to the load menu (so you can change the load range to match a layer/all layers), then replace the assertion with an if block (so extending classes don't hit the same exception if they add their own menu).

schmidtk: @wallw-bits The data menu still has issues. The first time its opened throws an exception, then you get a duplicate menu item to load time for a layer because line 882 assumes there is an "All Data" item in the Data group and drops all children but the first.

scss/os/overrides/_openlayers.scss
```diff
@@ -125,3 +125,11 @@
     color: #000;
   }
 }
+
+.ol-attribution li:not(:first-child):before {
+  content: ", "
```
schmidtk: The Source/Sources label is the first child, so this is adding an extra comma before the first source.

PR #126, opened 2018-02-12
fix(test): Try to make QueryEntriesState test more robust.

bradh:
This is a blind fix, trying to overcome occasional 2 != 3 results.


bradh:
For an example of this, see https://travis-ci.org/ngageoint/opensphere/builds/344033485

PR #127, opened 2018-02-12
feat(docs): Add cookbook example of audio manager notifications.

bradh:


docs/cookbook/audionotification/index.rst
```diff
@@ -0,0 +1,61 @@
...
+A programmatic approach is shown below:
+
+.. code-block:: javascript
+
+  var audioManager = os.audio.AudioManager.getInstance();
+  audioManager.addSound(os.ROOT + 'sounds/cowbell.wav', 'label');
+  audioManager.play('label');
+
+User upload uses the normal Import Data dialog. Note that it only works if the target is a HTTP or HTTPS URL.
+
```
wallw-bits: This is true only if the application is served via web. If you are using something like Electron.io to package an app for desktop installation, `file://` should also work.

bradh: I thought I saw some code that checked for this. Will double check.

bradh:
If you hold off on this for now, I'll do another pass over it on the weekend.

schmidtk:
What's the source on your cowbell audio clip? I know the repo needs more cowbell, but only if it's open source cowbell. :)

bradh:
@schmidtk I created it using Audacity (i.e. its original work).

ObDisclosure: It doesn't really sound like a literal cow bell. Its more like this computer-generated four-tone sequence... However the Audacity menu option (`Beat sound:` in https://manual.audacityteam.org/man/rhythm_track.html) for this was "Cowbell", which seemed like a good name, especially since we can upgrade it to real cowbell without changing the docs...

Issue #115, opened 2018-02-08
Support reading KML Track/MultiTrack


schmidtk:
KML [Track](https://developers.google.com/kml/documentation/kmlreference#gxtrack) and [MultiTrack](https://developers.google.com/kml/documentation/kmlreference#gxmultitrack) defines how an object moves around the world over time. The Openlayers KML format supports parsing these, and OpenSphere should support displaying them and animating them via the timeline.

This should support both the KML 2.2 gx extension, and the formalized KML 2.3 spec.

PR #128, opened 2018-02-12
Add track plugin

schmidtk:
Adds a plugin that supports animating tracks comprised of many coordinates, each with a timestamp.

Initial track support for #115.


src/os/feature/dynamicfeature.js
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Updates the feature to display for the given timestamp.
+ */
+os.feature.DynamicFeature.prototype.initDynamic = function() {
+  this.initFn(this);
+};
+
+
+/**
+ * Updates the feature to display for the given timestamp.
```
jsalankey: Comments here look copy pasted.

src/os/feature/dynamicfeature.js
```diff
@@ -0,0 +1,92 @@
...
+};
+
+
+/**
+ * @inheritDoc
+ * @suppress {accessControls} To allow direct access to feature metadata.
+ */
+os.feature.DynamicFeature.prototype.clone = function() {
+  var clone = new os.feature.DynamicFeature(undefined, this.initFn, this.disposeFn, this.updateFn);
+  clone.setProperties(this.values_, true);
```
jsalankey: This could be done with a `getProperties()` to avoid the suppress tag.

src/os/feature/dynamicfeature.js
schmidtk: Missed that, thanks. Fixed.

src/os/feature/dynamicfeature.js
schmidtk: It's done this way for performance reasons, to avoid an unnecessary copy of the properties. We did the same thing in our mixin for `ol.Feature#clone`.

schmidtk:
I'll work on some tests to verify the dynamic feature behavior in vector source. We also need tests for `track.js` but I'll write that up separately.

bradh:
@schmidtk: Compile errors from Travis. Looks like its insisting on an Object.

src/os/feature/dynamicfeature.js
jsalankey: Gotcha. Figured this wasn't called that much so the usual performance caveat wouldn't apply.

PR #131, opened 2018-02-12
chore(hooks): prevent issue collisions between github and bitbucket

wallw-bits:
Our internal bitbucket server references `pull request #X in WV/...` in merge commit messages, but it doesn't use the `#` to actually reference anything. The `#X` reference is causing unrelated references in Github. Therefore, this hook rewrites that commit message in a way that keeps all the information but does not cause the unrelated references.


wallw-bits:
Nevermind, this is really the wrong place for that and there's no way husky gets installed server-side. I'll investigate some other solutions.

Issue #135, opened 2018-02-15
Improve Cesium multi-geometry translation


schmidtk:
Functions like `ol.geom.MultiLineString#getLineStrings` and `ol.geom.MultiPolygon#getPolygons` reconstitute the sub geometries from flat coordinates on each call. Since new geometries are created each time, the Cesium feature converter dumps the old primitive(s) representing the multi geometry and creates new ones. This happens on every change to the feature, including style changes, and is very expensive. We should determine a better approach to synchronizing multi geometries that promotes reuse of Cesium primitives.

It would be ideal to avoid calling `getLineStrings`, etc, and instead using the flat coordinates, stride, and ends to create Cesium primitives. This would avoid creating unnecessary Openlayers objects that are going to be immediately GC'ed.

Consider track animation while working this. MultiTracks and tracks crossing the date line are split into a MultiLineString, and while animating they may show `0..n` of the sub-geos.

wallw-bits:
Are you also concerned about Array creation in those? If not, a happy medium might be to call `getCoordinates()` rather than `getLineStrings()` or `getPolygons()`, as those call into the proper `ol.geom.flat.inflate.coordinatesX()` method. That's what we're doing in `os.interpolate`, but that file also intends to change the coordinates array rather than simply read it.

schmidtk:
The update will happen on each timeline frame while animating tracks, so I'd prefer to use the flat coordinates directly to avoid as much unnecessary GC as possible.

PR #136, opened 2018-02-15
Add KML track support

schmidtk:
This adds support for parsing both KML 2.2 and 2.3 Track and MultiTrack elements into a `os.feature.DynamicFeature` that can be animated via the timeline.

Other changes:
* `plugin.track.createTrack` API was changed to take an options argument, instead of the old parameter train. Choo choo!
* `plugin.track.addToTrack` was renamed to `plugin.track.addFeaturesToTrack` for clarity.
* Tracks will now use `ol.geom.LineString` if possible (not a multi-track, and doesn't cross the date line) as a performance enhancement. This is particularly useful for 3D mode, since our feature converter recreates all line primitives representing a multi-line on each update. #135 will address the greater issue.

### DIFF
--- a/src/os/feature/feature.js
+++ b/src/os/feature/feature.js
@@ -693,9 +693,15 @@ os.feature.getColor = function(feature, opt_source, opt_default) {
     }
 
     if (!color) {
-      var featureConfig = /** @type {Object|undefined} */ (feature.values_[os.style.StyleType.FEATURE]);
+      var featureConfig = /** @type {Array<Object>|Object|undefined} */ (feature.values_[os.style.StyleType.FEATURE]);
       if (featureConfig) {
-        color = os.style.getConfigColor(featureConfig) || undefined;
+        if (Array.isArray(featureConfig)) {
+          for (var i = 0; !color && i < featureConfig.length; i++) {
+            color = os.style.getConfigColor(featureConfig[i]) || undefined;
+          }
+        } else {
+          color = os.style.getConfigColor(featureConfig) || undefined;
+        }
       }
     }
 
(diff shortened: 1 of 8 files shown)


schmidtk:
Marked as WIP because I'm waiting on some other track changes to come in.

welchyd:
Would it be possible to add an export option for rotation here as well?

schmidtk:
@welchyd I wrote #154 to address current position heading on tracks. That will cover KML import/export of heading, and generating tracks from data with known heading.

bradh:
@schmidtk Is this ready for test?

schmidtk:
Yes, all good now. You should be able to load KML's with Track/MultiTrack elements (2.2 or 2.3), animate them with the timeline, and export them to KML's that will work in Google Earth.

bradh:
Built and ran locally for me with a KML 2.2 track. I did note a UI issue which is probably more general than this (filed at #159), and should not block merging.

With https://raw.githubusercontent.com/mapbox/togeojson/master/test/data/multitrack.kml I see some UI artifacts that look like styles and Document properties being interpreted as features. I would expect to see the tour and the two placemarks, but not the other leaf elements in the tree.

image

A KML 2.3 file from the OGC KML 2.3 test suite worked OK, although there was no feature info displayed. Not sure if that is intended or not.

schmidtk:
We haven't seen that style node issue before, but those should not be in the tree. It's an issue on the surge master deployment as well (https://master-branch-opensphere-ngageoint.surge.sh/), so not introduced here but I'll take a look at it when I get back on Monday. Thanks Brad!

bradh:
I added a ticket for the unexpected nodes (since its independent of this PR) as #163

PR #138, opened 2018-02-17
fix(sass): Fix remaining sass-lint warnings and upgrade configuration.

### LINKED ISSUE #121: Tune sass-lint configuration
After the current sass-lint warnings are fixed or suppressed, we should tune the configuration to make what are currently warnings into errors.

bradh:
Resolves #121.

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -26,90 +26,90 @@ rules:
     - 0
     - max-depth: 1
   border-zero:
-    - 1
+    - 2
   brace-style:
-    - 1
+    - 2
     - allow-single-line: true
   class-name-format:
     - 0
     - convention: hyphenatedlowercase
   clean-import-paths:
-    - 1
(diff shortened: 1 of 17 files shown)


schmidtk:
Thanks for cleaning these up! We should probably have a follow-on ticket to go through the `eslint-disable` lines and see if we can fix that SCSS to use a recommended approach. Did you look into those at all?

bradh:
I added #140 and #141. 

I did look at fixing some of the stuff that was previously disabled, but I wanted to keep this change to be fairly clean (no behavioural changes) since its already fairly large. Also, it wasn't clear to me what browser / OS combinations need to be tested if we're reverting fallbacks.

bradh:
@schmidtk - assume you meant `sass-lint: disable` rather than `eslint-disable` here, just multiple lint things happening at the same time?

PR #142, opened 2018-02-19
refactor(require-jsdoc): upgraded lint plugin and fixed issues it exp…

wallw-bits:
And fix lint issues with new version.

### DIFF
--- a/src/os/mixin/renderloop.js
+++ b/src/os/mixin/renderloop.js
@@ -9,6 +9,12 @@ goog.require('os.olcs.RenderLoop');
  */
 (function() {
   var origEnable = olcs.AutoRenderLoop.prototype.enable;
+
+  /**
+   * Overridden to listen to <code>os.olcs.RenderLoop.REPAINT</code> events in addition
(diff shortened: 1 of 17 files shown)


src/os/mixin/renderloop.js
```diff
@@ -9,6 +9,12 @@ goog.require('os.olcs.RenderLoop');
  */
 (function() {
   var origEnable = olcs.AutoRenderLoop.prototype.enable;
+
+  /**
+   * Overridden to listen to <code>os.olcs.RenderLoop.REPAINT</code> events in addition
```
schmidtk: I've been using backticks instead of `<code>` in docs for brevity. Either works though, just a side note.

wallw-bits: Oh, does `js-dossier` support markdown in comments or just the backticks?

schmidtk: Yes, it supports markdown. Some details here:
https://github.com/jleyba/js-dossier#formatting

PR #143, opened 2018-02-20
chore(tests): Add tests for os.geo.normalizeLongitude_().

bradh:
I originally wrote these for my own normalisation function, before discovering the one in [os.geo](https://github.com/ngageoint/opensphere/blob/master/src/os/geo/geo.js#L1791).


wallw-bits:
Thanks! In the future, the `test(geo)` is a better way of categorizing that change. We currently don't have any scripts that care about that category, so it is fine for now.

PR #144, opened 2018-02-20
WIP: Ffmpegencoder

welchyd:
This is in no way complete, but I wanted to get the PR out there in case someone else wanted to pick it up. Leverages an asmjs port of ffmpegencoder to make better videos. Better meaning smaller, higher quality, and faster to produce.
Where I'm stuck is at getting an array of images into the asm port ... all of the examples seem to work off of converting from webm, so we may need another port entirely (could mean making our own)
See https://github.com/muaz-khan/Ffmpeg.js

### DIFF
--- a/src/os/capture/ffmpegencoder.js
+++ b/src/os/capture/ffmpegencoder.js
@@ -64,7 +64,7 @@ os.capture.ffmpegEncoder.SCRIPT_URL = os.ROOT + os.capture.ffmpegEncoder.SCRIPT_
  */
 os.capture.ffmpegEncoder.prototype.processInWebWorker = function() {
   var ffmpegPath = window.location.href + os.capture.ffmpegEncoder.SCRIPT_LOCATION;
-  var blob = URL.createObjectURL(new Blob(['importScripts("' + ffmpegPath + '");' + 
+  var blob = URL.createObjectURL(new Blob(['importScripts("' + ffmpegPath + '");' +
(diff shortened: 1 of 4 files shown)


src/os/capture/ffmpegencoder.js
```diff
@@ -0,0 +1,353 @@
...
+goog.require('os.capture.AbstractVideoEncoder');
+goog.require('os.capture.GifEncoder');
+goog.require('os.capture.WebMEncoder');
+goog.require('os.defines');
+
+
+
+/**
+ * A ffmpeg video encoder.
+ * @extends {os.capture.AbstractVideoEncoder}
```
wallw-bits: I know we don't have a lint rule for this, but classes/constructors should begin with an uppercase letter to be consistent with the rest of the code.

src/os/capture/ffmpegencoder.js
```diff
@@ -0,0 +1,353 @@
...
+ * @constructor
+ */
+os.capture.ffmpegEncoder = function() {
```
wallw-bits: I know we don't have a lint rule for this, but classes/constructors should begin with an uppercase letter to be consistent with the rest of the code.

wallw-bits:
To be honest, I've always pushed for the capture stuff to not be included. Users should be using some other tool for taking screenshots or making videos, all of which are significantly better at doing that job than our capture plugin.

PR #146, opened 2018-02-21
Cesium Image Layer Synchronizer

wallw-bits:
Our previous image layer synchronizer was a hack in the sense that it only supported KML Ground Overlay layers.

This addition should support any image layer generically, including ground overlays, heatmap rasters, and others.

Caveats (open to suggestions):
- This isn't perfect if the source is changing its image at a high speed (i.e. with the timeline). It can attempt to render an item that the underlying source has already cleaned up because it lags a bit.
- I also couldn't figure out anything to support layer opacity (at least in Cesium itself)

### DIFF
--- a/externs/Cesium.externs.js
+++ b/externs/Cesium.externs.js
@@ -1493,6 +1493,7 @@ Cesium.Polygon.prototype.material;
  * @typedef {{
  *   color: (Cesium.Color|undefined),
  *   horizontal: (boolean|undefined),
+ *   image: (string|undefined),
  *   repeat: (number|undefined),
  *   evenColor: (Cesium.Color|undefined),
  *   oddColor: (Cesium.Color|undefined)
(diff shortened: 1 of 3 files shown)


schmidtk:
There are a few projection issues here:
- Heatmaps don't work in 3D mode in projections other than 4326. They don't work on master either, but I'm hoping that's an easy fix.
- GroundOverlays don't work in 2D mode in projections other than 4326. This was an existing bug, but it seems your changes do make them work in 3D mode in other projections. It looks like the KML parser needs to transform the extent it provides to the layer/source when creating the ground overlay. Again, existing problem but hopefully easy to address.

wallw-bits:
I propose addressing the KML Ground Overlay since it is just a matter of putting the given extent in the correct projection. The heatmap has other projection issues that are already fixed on my heatmap-performance branch.

schmidtk:
Sounds good. No reason to address it twice.

wallw-bits:
Had a quick look. Layers which can generate their rasters in different projections (such as heatmap), we can support. However, KML GroundOverlays are given in `EPSG:4326`. As long as `enableReprojection` is `false` in config, we should not show that image in any other projection except the one in which it is originally given.

Currently, this just means that the layer does not show. We need to give a message stating that the image is in a different projection. We can do that either by triggering a projection switch when the layer is added (I prefer this since it is consistent), or by doing something a little less intrusive.

schmidtk:
Should we investigate raster reprojection in 2D mode (not here, just in general)? Openlayers has client-side reprojection examples for TileWMS, XYZ, Image, etc layers. I don't like triggering a projection switch for this case because any vector data in the KML can and will be transformed automatically. Unless nothing in the added layer can be viewed in the current projection, I don't think we should switch.

wallw-bits:
Upon even further inspection, the KML Ground Overlay implementation is incomplete. We don't support the `<rotation>` tag on the `<LatLonBox>`, so even if we can display it (CORS for 3D, etc.), it is probably wrong. This was likely skipped because OpenLayers does not support rotation on `ol.Image` (and image sources and layers). However, it is very doable to simply draw the rotated image to an intermediate canvas and then pass that in.

Also, the spec does not define a projection for the image, which potentially makes us violate the `enableReprojection` setting, which is intended to keep all imagery in the view in the same projection (customer rule). While a ground overlay does not have to be imagery (could be transparent overlays), the projection still matters for proper pixel transforms.

I vote we make a separate issue for it since it is turning into a can of worms. I can move the heatmap PR somewhere public and use that as the major test for this if you like.

schmidtk:
Sounds like there are too many issues with GroundOverlay to creep into this issue. I'll second creating a new one.

schmidtk:
@wallw-bits This needs a rebase. The camera change can be skipped (I pulled it into the Cesium plugin refactor), and the image synchronizer changes need updated with the new `plugin.cesium` namespace.

wallw-bits:
Closed in favor of #644

PR #149, opened 2018-02-22
fix(configdescriptor): fix null error

wallw-bits:


wallw-bits:
Closing because Kevin's fix was more thorough.

PR #152, opened 2018-02-23
feat(electron): Support file:// URL's in Electron.

schmidtk:
Electron [adds a `path` property](https://electronjs.org/docs/api/file-object) to the File API to expose the file system path. This PR detects that and loads the file directly from the file system. The file is treated as a remote URL so it will not be added to file storage. Support for `file://` URL's is disabled by default, and can be enabled by setting `fileUrls` to `true` in the app configuration.

Layers with `file://` URL's will always be ignored when creating state files and by request URL parameter edit. They _can_ be imported into the audio manager.

### DIFF
--- a/src/os/file/file.js
+++ b/src/os/file/file.js
@@ -2,7 +2,6 @@ goog.provide('os.file');
 goog.provide('os.file.File');
 
 goog.require('goog.async.Deferred');
-goog.require('goog.crypt.base64');
 goog.require('goog.fs.FileReader');
 goog.require('goog.net.jsloader');
 goog.require('os.IPersistable');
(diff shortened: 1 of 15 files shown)


src/os/file/file.js
```diff
@@ -313,27 +323,52 @@ os.file.createFromContent = function(fileName, url, originalFile, content) {
...
+
+
+/**
+ * Creates a `local://` url used by file storage.
+ * @param {string} fileName The file name to use in generating the url.
  * @return {string}
  */
 os.file.getLocalUrl = function(fileName) {
-  return os.file.File.URL_SCHEME + '://' + goog.crypt.base64.encodeString(fileName);
+  return os.file.FileScheme.LOCAL + '://' + goog.crypt.base64.encodeString(fileName);
```
wallw-bits: I know this doesn't have anything to do with your change, but was there a reason that we were using `goog.crypt.base64.encodeString()` over the native `window.btoa()`?

schmidtk:
Switched to using `btoa` and removed some extra references to `goog.crypt.base64` (looked like copy/pasted `goog.require` statements).

PR #153, opened 2018-02-23
fix(dragbox): Don't update if start/end coord are missing.

### LINKED ISSUE #150: Fix drawing tools escape
If a user hits the escape button while drawing to cancel drawing something that they did not intend to draw, it breaks the drawing tools and map menus.

schmidtk:
Fixes #150.


welchyd:
Looks good to me ... just can't approve :/

PR #156, opened 2018-02-26
chore(deps): Remove unused dependency on jasmine-core.

schmidtk:
Jasmine comes bundled with [karma-jasmine](https://github.com/karma-runner/karma-jasmine) 0.1.0, and our dependency on `jasmine-core` is both unused and misleading (we're not using Jasmine 2). Removing it.


bradh:
Looks good, tests pass in Travis. Recommend merge.

PR #158, opened 2018-03-01
WIP: Fix implicit dependency issues

schmidtk:
Implicit dependencies were discovered when compiling a library that depends on OpenSphere.

Most of these revolved around the use of `os.MapContainer.getInstance()`, which isn't required in most cases. The `os.map.IMapContainer` interface was added to allow creating custom map containers that will still work with much of OpenSphere. These references haven't been fixed across the board, but this PR handles a number of them.


bradh:
Looks like some relevant test failures in Travis CI.

schmidtk:
This is why you should never open PR's at the end of the day. I'll be out for a few days, so marking this as WIP for now.

schmidtk:
Closing in favor of the less invasive #168.

PR #162, opened 2018-03-01
Index ellipse and lob shapes in the r-tree

### LINKED ISSUE #107: Ellipses and LOBs disappear when the origin is out of sight in 2D mode
In 2D mode, when I draw an ellipse or lob, and zoom in far enough so that I can see part of the ellipse or line but not the origin of a feature, the ellipse/line disappears.

@wallw-bits pointed out this could be tied to forEachFeatureIntersectingExtent in vector.js which does an intersection test with feature.getGeometry

wallw-bits:
Fixes style shapes that change geometry (ellipse and LOB). Previously they would not display in the OpenLayers view if the original (point) geometry was off the viewport. This is done by updating the spatial index with the combined extent of the new geometry with the original for each feature.

When the style is turned off, the index reverts to only the original geometry extent.

Resolves #107 

@welchyd

### DIFF
--- a/src/os/source/vectorsource.js
+++ b/src/os/source/vectorsource.js
@@ -898,20 +898,50 @@ os.source.Vector.prototype.getGeometryShape = function() {
 /**
  * Sets the geometry shape used by features in the source.
  * @param {string} value
+ * @suppress {accessControls}
  */
 os.source.Vector.prototype.setGeometryShape = function(value) {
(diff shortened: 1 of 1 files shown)


welchyd:
Sweet! The lob style does have an additional option to add an ellipse, so will need to get both in that case.

wallw-bits:
Changing this to WIP because I want to clean it up a bit.

PR #166, opened 2018-03-03
chore(d3): Remove d3 vendor, switch to npm.

bradh:
Bumps d3 to 3.5.17 (last of the 3 series).


bradh:
I did a quick check with timeline (tooltip shows fine), but couldn't test the other graphs - need to do that with the downstream project.

wallw-bits:
The timeline is the only place we are using it. I'll give it a look.

PR #167, opened 2018-03-04
fix(kml): Handle KML file that has Schema without name attribute

### LINKED ISSUE #163: Unexpected nodes in layer tree (KML)
[Originally from review of #136, but actually occurs without that change].

With https://raw.githubusercontent.com/mapbox/togeojson/master/test/data/multitrack.kml I see some UI artifacts that look like styles and Document properties being interpreted as features. I would expect to see the tour and the two placemarks, but not the other leaf elements in the tree.

image

bradh:
Resolves #163

### DIFF
--- a/src/plugin/file/kml/kmlparser.js
+++ b/src/plugin/file/kml/kmlparser.js
@@ -1431,7 +1431,14 @@ plugin.file.kml.KMLParser.prototype.parseSchema_ = function(el) {
   var schemas = os.xml.getChildrenByTagName(el, 'Schema');
   for (var i = 0, n = schemas.length; i < n; i++) {
     var name = schemas[i].name || schemas[i].getAttribute('name');
-    this.kmlThings_.push(name);
+    if (!name) {
+      continue;
+    }
+
+    if (this.kmlThings_.indexOf(name) === -1) {
+      this.kmlThings_.push(name);
+    }
+
     var fields = os.xml.getChildrenByTagName(schemas[i], 'SimpleField');
 
     /**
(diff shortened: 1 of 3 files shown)


src/plugin/file/kml/kmlparser.js
```diff
@@ -1432,7 +1432,9 @@ plugin.file.kml.KMLParser.prototype.parseSchema_ = function(el) {
   var schemas = os.xml.getChildrenByTagName(el, 'Schema');
   for (var i = 0, n = schemas.length; i < n; i++) {
     var name = schemas[i].name || schemas[i].getAttribute('name');
-    this.kmlThings_.push(name);
+    if ((name != null) && (name != '')) {
```
schmidtk: `if (name)` is sufficient to test that the value is defined and not an empty string. If you also want to defend against whitespace-only strings, `goog.string.isEmptyOrWhitespace(goog.string.makeSafe(name))` is a good (but unfortunately verbose) way to do that.

schmidtk: After reviewing our current schema support, it seems you should skip the schema entirely if it doesn't provide a name. We don't support `SchemaData schemaUrl="#someSchemaId"` right now, so there is nothing further to do. I would recommend simply doing:
```
if (!name) {
  continue;
}
```

bradh:
@schmidtk I did something similar in an earlier attempt, but there is a test for the schema-as-inheritence support that will break. Should I just remove that?

schmidtk:
@bradh Both the existing schema test and the one you added in your PR should still work. The existing test has name fields, and yours should skip the schema but still parse the feature. Is there another test that would break? I was looking at a change like this and nothing jumped out:

```
var name = schemas[i].name || schemas[i].getAttribute('name');
if (!name) {
  continue;
}

if (this.kmlThings_.indexOf(name) === -1) {
  this.kmlThings_.push(name);
}
```

We also created a Slack workspace for OpenSphere if you'd like to join that. Should make communication a lot easier. Invitation link is in the Google Group here:

https://groups.google.com/forum/#!topic/opensphere/_fLb9m99WUk

PR #169, opened 2018-03-05
Stop using goog.dom.getTextContent and getRawTextContent.

schmidtk:
The Closure functions fail when parsing multi-line XML with CDATA to denote the value. For example, the following would be parsed as whitespace:
```
<name>
  <![CDATA[Some Value]]>
</name>
```

This caused problems reading values from formatted KML files, but these are resolved by using `ol.xml.getAllTextContent` instead. This function is also guaranteed to return a string, so I streamlined some following `trim` calls as well.

### DIFF
--- a/conformance_config.textproto
+++ b/conformance_config.textproto
@@ -121,6 +121,20 @@ requirement: {
   value: 'Number.MIN_VALUE'
 }
 
+requirement: {
+  type: BANNED_NAME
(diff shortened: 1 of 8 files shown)


schmidtk:
Tests are passing locally. I know the Job one is known to randomly fail, but I haven't seen that tour failure yet. Running the build again.

PR #171, opened 2018-03-05
fix(main): fix race condition on osasm load

wallw-bits:
This was only occuring on very slow networks and did not have much of an impact other than a couple of errors in the console. Subsequent calls to the function worked fine after the load completed.

### DIFF
--- a/src/os/main.js
+++ b/src/os/main.js
@@ -66,6 +66,9 @@ os.Module.config(os.Module.configureModule);
     return !!osasm.geodesicInverse;
   });
 
+  /**
+   * After osasm loads, kick off the rest of the application
+   */
   appWait.onSuccess = function() {
     if (os.isMainWindow()) {
       try {
--- a/src/os/main.js
+++ b/src/os/main.js
@@ -1,6 +1,7 @@
 goog.provide('os.Module');
 goog.provide('osmain');
 
+goog.require('goog.async.ConditionalDelay');
 goog.require('os');
 goog.require('os.MainCtrl');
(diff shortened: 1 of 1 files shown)


wallw-bits:
It occurred to me right after I submitted this that the `os.AppStartDelay` is actually way more generic than its name implies. I can move it if you like.

src/os/appstartdelay.js
```diff
@@ -0,0 +1,38 @@
+goog.provide('os.AppStartDelay');
```
schmidtk: I started creating one of these in the past, but it's not really necessary since `onSuccess` and `onFailure` are public on `goog.async.ConditionalDelay` and can be replaced directly on the instance. See [this example](https://github.com/ngageoint/opensphere/blob/master/src/os/storage/storage.js#L53).

src/os/main.js
```diff
@@ -56,30 +57,37 @@ os.Module.configureModule = function($routeProvider) {
...
+        // set up request handlers
+        os.net.addDefaultHandlers();
 
-    // set up request handlers
-    os.net.addDefaultHandlers();
+        // initialize settings for this app
+        var settingsInitializer = new os.config.SettingsInitializer();
+        settingsInitializer.init();
+      }
+    });
```
schmidtk: This should have an error callback that throws an error, so if anyone encounters the issue they have something useful to tell support (assuming they checked the console).

bradh:
Travis failed eslint check: js-doc for the added `onSuccess` function.

PR #177, opened 2018-03-06
fix(units): Show 0 in base units

### LINKED ISSUE #173: Default altitude measurement should be metres
If you have altitude of 0:

```
    <Placemark id="mountainpin1">
      <name>New Zealand's Southern Alps</name>
      <styleUrl>#style0</styleUrl>
      <Point>
        <coordinates>170.144,-43.605,0</coordinates>
      </Point>
    </Placemark>
```

then the altitude is shown in cm:
image

It could also be worth aligning precision in the header (longitude / latitude shown with 5 decimal places) with precision in the attribute display (3 decimal places).

bradh:
For example, 0 in metres is 0.00 m, not 0.00 cm

Resolves #173.

### DIFF
--- a/test/os/unit/unitmanager.test.js
+++ b/test/os/unit/unitmanager.test.js
@@ -78,6 +78,9 @@ describe('os.unit.UnitManager', function() {
     expect(um.formatToBestFit('distance', 1000, 'm', um.getBaseSystem(), 0)).toBe('1 km');
     expect(um.formatToBestFit('distance', 999, 'm', um.getBaseSystem(), 0)).toBe('999 m');
 
+    // Test zero to default
+    expect(um.formatToBestFit('distance', .0, 'm', um.getBaseSystem(), 2)).toBe('0.00 m');
+
     // Test the boundaries (mi to ft)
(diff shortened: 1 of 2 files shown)


test/os/unit/unitmanager.test.js
```diff
@@ -78,6 +78,9 @@ describe('os.unit.UnitManager', function() {
     expect(um.formatToBestFit('distance', 1000, 'm', um.getBaseSystem(), 0)).toBe('1 km');
     expect(um.formatToBestFit('distance', 999, 'm', um.getBaseSystem(), 0)).toBe('999 m');
 
+    // Test zero to default
+    expect(um.formatToBestFit('distance', .0, 'm', um.getBaseSystem(), 2)).toBe('0.00 m');
```
schmidtk: Out of curiosity, why `.0` instead of `0`?

bradh: Its my C# / C++ background - conceptually `double` instead of `int`.

schmidtk: Ohh, roger that. Carry on.

Issue #77, opened 2018-01-21
Task: Review vendor/ and migrate to npm where possible


bradh:
Some of [vendor](https://github.com/ngageoint/opensphere/tree/master/vendor) is pretty old, and it might be useful to migrate to newer versions, ideally via npm. I'd see this done in two stages - see what is now available, and progressively migrate.

Example: note that npm offers bootstrap x.y.z, raise a separate issue for the migration.

PR #178, opened 2018-03-06
Use npm for zip.js library.

schmidtk:
Replaces `vendor/zip` with `zip-js@0.0.2` from NPM, in support of #77.

This is marked as WIP because it depends on ngageoint/opensphere-build-resolver/pull/11. When that is merged, I'll update the required version of `opensphere-build-resolver` in `opensphere/package.json`.


docs/guides/app_package_guide/index.rst
```diff
@@ -20,6 +20,7 @@ The ``build`` property is used to instruct the resolver during the build. For th
 * ``index``: The index file used by opensphere-build-index_.
 * ``scss``: The root SCSS file for the application.
 * ``gcc``: Additional instructions for the `Google Closure Compiler`_.
+* ``moduleDefines``: ``goog.define`` properties that should be resolved to a ``node_module`` path in uncompiled mode.
```
wallw-bits: I'm ecstatic that the guide was updated along with the code. However, shouldn't this be documentation as part of the resolver project?

schmidtk: If the resolver had RTD guides, I would say yes. The documentation in the readme is very minimal and this seemed to be the better place for it.

wallw-bits: I agree and will open a ticket for some more extensive documentation there.

PR #180, opened 2018-03-08
Lazy load Cesium the first time 3D mode is activated.

schmidtk:
This PR defers loading Cesium until the first time 3D mode is activated.

Summary of changes:
- Moved terrain activation to the map menu/settings to be consistent with sunlight controls. I left the terrain descriptor for now, but made it tell the user how to activate terrain.
- If terrain can't be loaded (`errorEvent` is raised by the provider), an alert is fired and terrain is completely removed from the UI. Previously the globe would turn black with no user-facing indication why.
- Interactions were creating Cesium objects in the constructor. I refactored these to create the objects the first time they're needed. This affects all drawing controls (box, circle, polygon, line), measure, and drag zoom.
- Cesium mixins are now loaded in a function, which can be called after Cesium is loaded.
- `os.olcs.WMSImageryProvider` extends a Cesium class, which required actually setting it up in a function call.
- `os.olcs.VectorContext` extended `Cesium.PrimitiveCollection`, which was really a lazy way of not having to manage the collection separately. It now manages a collection as a property.


welchyd:
Would it be possible to provide a way to load things for offline use .. ie I'm planning to use this offline and want to make sure I cache enough of the page to be able to use it.

schmidtk:
The library will be cached by the browser if you use it once online. If you want an offline version of the app, see [opensphere-electron](https://github.com/ngageoint/opensphere-electron).

csgibson:
I get what @welchyd is saying.
A user is online but is getting set up to use the app (potentially offline) later. They don't want to find out that something that they expect to work isn't there because their connection is down or too far away. Especially if they have a critical need. 
In this case it seems like 3D is something that they would do when they are getting set up while online, so not that big of a deal if it is lazily loaded. But we want to be mindful of those types of users when we decide to lazily load something.
@welchyd had an idea: ideally a bunch of things could be lazily loaded based on some sort of user settings. "I'm never going to use that" -> lazy load it. Or vice versa.

welchyd:
For the record - I'm all for lazy loading, we just realized that there is some offline use going on as well.
I've opened an issue for allowing some control over lazy loading - #181

wallw-bits:
I agree with Kevin. Every remote template in the application will fail if it has not been loaded and the user is offline (resulting in a blank window). Every uncached map tile, terrain tile, and data request will also fail (even some cached ones may fail if they require checks for new content from the server).

The electron package addresses offline use as all the code is local to the machine. However, it does not address the map tile issue. You still need a connection to a server or have a lot of imagery loaded locally on the machine. The electron-aware version of the GeoPackage plugin should help address that.

That being said, let's keep the rest of this discussion related directly to the PR. Feel free to continue it in chat or on the forum.

PR #191, opened 2018-03-13
chore(select2): Switch to npm.

bradh:
Towards #77.


jsalankey:
Just checked, we do not. Next worry would be if there are any breaking changes that they didn't account for in the versioning. 3.2 is waaay old and I have trouble believing that they made absolutely no breaking changes in that time.

welchyd:
We should get this commit in

jsalankey:
Yep, we're good to merge.

PR #192, opened 2018-03-13
Inform the user when WebGL is being initialized.

schmidtk:
I noticed two problems with the async Cesium load:
- On slow connections or in debug mode when loading unminified Cesium, the 5 second timeout for `goog.net.jsloader` can be reached.
- It's unclear to the user that something is happening, other than the map going black.

To resolve these, I made the following changes:
- Show a loading spinner with updated tooltip on the map mode control while Cesium is being loaded/initialized.
- Hide the Openlayers canvas when initializing 3D mode on app startup. It seemed odd to show the 2D map then switch to 3D when the user hadn't done anything yet.
- Don't hide the Openlayers canvas during initialization if user toggled. The loading spinner provides feedback that their button click did something and we're working on it, and avoids displaying a blank screen while that happens.
- Increase the load timeout to 30 seconds by default, configurable with `cesium.loadTimeout`.

### DIFF
--- a/src/os/control/mapmodecontrol.js
+++ b/src/os/control/mapmodecontrol.js
@@ -25,21 +25,34 @@ os.control.MapMode = function(opt_options) {
   var textClass = options.textClass != null ? options.textClass : 'ol-mapmode-text';
 
   /**
-   * @type {Element}
+   * Default tooltip to display on the button.
+   * @type {string}
+   * @protected
+   */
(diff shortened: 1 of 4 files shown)


bradh:
Tried building and running this locally, seems OK. The spinner isn't very obvious - probably a good compromise on faster loading when you don't have much time to look around.

I didn't try adjusting the timeout, my load is relatively fast anyway.

PR #194, opened 2018-03-15
feat(descriptor): config desc should support icons

wallw-bits:


bradh:
Add docs to https://github.com/ngageoint/opensphere/blob/master/docs/guides/settings_guide.rst ?

wallw-bits:
That's not the correct spot in the guides, but https://opensphere.readthedocs.io/en/latest/guides/plugin_server_guide/parsing.html is.

PR #197, opened 2018-03-16
feature(layer): allow override of the layer description.

### LINKED ISSUE #198: Support alternative layer descriptions
This should come from the base descriptor, but allow override.

bradh:
Resolves #198.

### DIFF
--- a/src/os/data/basedescriptor.js
+++ b/src/os/data/basedescriptor.js
@@ -1,6 +1,8 @@
+
 goog.provide('os.data.BaseDescriptor');
 
 goog.require('goog.async.nextTick');
+goog.require('goog.date.UtcDateTime');
 goog.require('goog.events.Event');
 goog.require('goog.events.EventTarget');
 goog.require('os.data.ColumnDefinition');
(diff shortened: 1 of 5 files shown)


src/os/data/basedescriptor.js
```diff
@@ -698,6 +699,40 @@ os.data.BaseDescriptor.prototype.restore = function(from) {
 };
 
 
+/**
+ * @inheritDoc
+ */
+os.data.BaseDescriptor.prototype.getDescriptionMessage = function() {
```
schmidtk: I think `getHTMLDescription` would be a little more fitting.

The original code from the layer menu also appears to be mostly copy/pasted from `os.ui.data.AddDataCtrl.prototype.getInfo`, which could also benefit from getting this directly from the descriptor. I updated that function recently to omit the Provider/Type lines if they're undefined/empty, so those changes should be brought in here. This does add Layer Name from the Add Data description, but I'm fine with that being in the Add Data window as well.

bradh:
Applied @schmidtk recommended changes, added unit tests for new method.

Issue #141, opened 2018-02-19
Review scss/ directory for disabled linting and fix


bradh:
There are some disabled lint checks scattered through the `scss/` directory, easily findable by searching for `sass-lint` in the tree. Those should be reviewed for alternative approaches.

bradh:
@wallw-bits Is there a definition of browsers (and versions) that we need to support in OS?

wallw-bits:
Latest Chrome or Firefox is always the recommended choice.

We have _no requirement_ to support IE _or_ Edge (this was a recent drop on our side after metrics for performance, time spent supporting bugs on the platform, lack of proper web API support, etc. were given to our customers). IE11 also had some client cert bugs that made it a non-starter in many cases.

Technically, our current minimum browser versions are listed in the `old.html` page that the build produces from `old-template.html`. We develop against basically latest-1 though, and attempt to support the old ones if any issues come up during testing.

Warning: Firefox in the 30s has a bug that can corrupt your IndexedDB storage. The only fix we've found for that has been to get users to update to 45+ (generally our users are on the extended releases, so 31, 38, 45, 52, etc.). 

We don't specifically support browsers like Safari, Opera, etc but we try to do feature detection rather than browser detection (you'll probably still see some browser-specific code in there occasionally for when we did not have a choice at the time or where spec interpretations differ). We've had a guy accidentally do a demo in Safari and it went fine.

Also, we've noted in the past that Chrome tends to be about 3x faster on average in the compiled version than Firefox. However, we haven't done any serious performance comparisons between those two in the last 18 months. There are certainly a couple of notable exceptions in Canvas operations where Firefox is significantly faster (I think there might be an issue logged with Chrome about that too, I'll see if I can find it.).

schmidtk:
When addressing `no-vendor-prefixes` violations, please consider if the prefixed styles can be replaced by [Compass mixins](http://compass-style.org/reference/compass/css3/). I would expect all of these to be covered by a mixin that `node-sass` will replace with vendor prefixes meeting the supported browser configuration. If the defaults for Compass don't suit us, we can [tune them](http://compass-style.org/help/documentation/tuning-vendor-prefixes/) to fit our browser requirements.

bradh:
[...]

PR #202, opened 2018-03-20
chore(css): Remove vendor prefixes for translate.

bradh:
This is supported in all modern browsers (Chrome from 36, Firefox from 16, even IE 10, Safari 9), see https://caniuse.com/#feat=transforms2d

This is part of the work in #141.


bradh:
OK, looks like we actually need Chrome from 35, so this is bad change.

PR #203, opened 2018-03-21
chore(css): Remove vendor prefix for translate and transition.

bradh:
Chrome 35 is a supported browser, and `translate` isn't available (without prefix) until 36, so that stays for now.

Part of the work for #141

### DIFF
--- a/scss/ui/partials/ui/onboarding/_ngonboarding.scss
+++ b/scss/ui/partials/ui/onboarding/_ngonboarding.scss
@@ -32,13 +32,9 @@
 }
 
 .onboarding-popover.onboarding-centered {
-  // sass-lint:disable no-vendor-prefixes
-  -moz-transform: translate(-50%, -50%);
-  -ms-transform: translate(-50%, -50%);
-  -webkit-transform: translate(-50%, -50%);
+  @include translate(-50%, -50%);
   left: 50%;
   top: 50%;
-  transform: translate(-50%, -50%);
 }
 
 .onboarding-arrow,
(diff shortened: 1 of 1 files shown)


scss/ui/partials/ui/onboarding/_ngonboarding.scss
```diff
@@ -32,10 +32,8 @@
 }
 
 .onboarding-popover.onboarding-centered {
-  // sass-lint:disable no-vendor-prefixes
-  -moz-transform: translate(-50%, -50%);
-  -ms-transform: translate(-50%, -50%);
-  -webkit-transform: translate(-50%, -50%);
+  // This can be removed after we drop support for Chrome before version 36.
+  -webkit-transform: translate(-50%, -50%); // sass-lint:disable-line no-vendor-prefixes
```
schmidtk: We should be able to replace this with the [Compass translate mixin](http://compass-style.org/reference/compass/css3/transform/#mixin-translate), which will add vendor prefixes as configured by the SCSS build.

scss/ui/partials/ui/onboarding/_ngonboarding.scss
```diff
@@ -185,9 +183,6 @@
 }
 
 .onboarding-focus {
-  // sass-lint:disable no-vendor-prefixes
-  -moz-transition: none !important;
-  -webkit-transition: none !important;
```
schmidtk: Same for these. There isn't anything wrong with having vendor prefixes, we just disallow them in the linter because they should be provided via a single mixin line from Compass.

bradh: I removed these because they're overridden by the following transition line which is supported for our required browsers.

schmidtk: That's fine. I was thinking we should replace all vendor prefixes (and their non-prefixed counterparts) with Compass mixins then rely on [tuning vendor prefix support](http://compass-style.org/help/documentation/tuning-vendor-prefixes/) to control which prefixes make it into the compiled CSS.

bradh:
PR updated, in Travis build.

PR #207, opened 2018-03-21
fix(build): Ensure onboarding is resolvable.

### LINKED ISSUE #206: Build system should use a .build args file for onboarding, not assume files are local
Something similar to the views directory approach would be better.

Also see https://github.com/ngageoint/opensphere-build-resolver/issues/14

bradh:
Resolves #206 

Note that https://github.com/ngageoint/opensphere-build-resolver/pull/15 needs to go in first.


bradh:
Updated dependencies to match required version of build-resolver.

PR #211, opened 2018-03-22
feat(wms): Added caseSensitive option for params

wallw-bits:
I was working with a provider who uses all caps in WMS spec params (`SERVICE`, `VERSION`, etc.) and all lower case in non-spec proprietary params (`provider`, `query`, etc.). So I added a `caseSensitive` option that prevents the WMS layer config from assuming that all the parameters can be upper-cased.

EDIT:
I should probably note that this is a little janky. The underlying `ol.source.TileWMS` class will still attempt to bake in the default parameters in upper case. So you can't add case-sensitive parameters for `SERVICE`, `VERSION`, `REQUEST`, `FORMAT`, `TRANSPARENT`, `WIDTH`, `HEIGHT`, `STYLES`, `CRS` (`SRS` for 1.1.1), or `BBOX`. However, anything else will be preserved as sent in the param options.

### DIFF
--- a/src/plugin/ogc/wms/wmslayerconfig.js
+++ b/src/plugin/ogc/wms/wmslayerconfig.js
@@ -110,7 +110,7 @@ plugin.ogc.wms.WMSLayerConfig.prototype.getSource = function(options) {
     var keys = this.params.getKeys();
     for (var i = 0, n = keys.length; i < n; i++) {
       var key = keys[i];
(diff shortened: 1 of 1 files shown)


wallw-bits:
I'm aware of what the spec dictates. However, in practice, many of the services we hit are merely WMS-like rather than fully spec-compliant. Therefore, it doesn't hurt to have an opt-in option for case sensitive parameters.

PR #212, opened 2018-03-22
refactor(datetimemapping): use RecordField.TIME const instead of 'rec…

wallw-bits:
Pretty sure this was an oversight and not intentional.


bradh:
Should this `goog.require('os.data.RecordField');`?

Issue #209, opened 2018-03-22
Refactor Filter API


wallw-bits:
I'm about to bash a bunch of stuff, but I either wrote it or it was copied from something else I wrote. So here goes.

The filter API is clunky because the filter builders save off a string of OGC Filter XML to the `FilterEntry`, rather than a data structure that matches the filter tree that was created. While this is fine and dandy for WFS and OGC Filter queries, it is extraordinarily annoying when working with other query languages (Arc, ECQL, Elastic, etc.). This terrible design decision has also crept into the API for `IFilterFormatter` and `ISpatialFormatter` and their use by `QueryHandler.createFilter()`.

Fix it.

* The data structure is a tree, and as such should extend `os.structs.TreeNode` or perhaps even `os.ui.slick.SlickTreeNode` for UI ease
* The logic in `QueryHandler.createFilter()` should be changed to create the full tree structure from the query entries and then that full structure should be passed to the filter formatter
* Formatter of structure to/from JSON (for use internally)
* Formatter of structure to/from OGC Filter spec
* Formatter of structure to Arc (`plugin.arc`)
* Migrate old filters and filter storage to new structure using new `OGCFilterFormatter.parse()` and `JSONFilterFormatter.format()`
* Filter Builder updates to read/write the new structure
* Filter export/import to file can be same format (using `OGCFilterFormatter`)
* State file export should still be OGC Filter (ditto)

@schmidtk and @jsalankey, thoughts?

PR #213, opened 2018-03-23
Add ECQL support

wallw-bits:
Adds support for ECQL filter format. See http://docs.geoserver.org/stable/en/user/filter/ecql_reference.html for more details.

### DIFF
--- a/test/os/filter/impl/ecql/exclusionformatter.test.js
+++ b/test/os/filter/impl/ecql/exclusionformatter.test.js
@@ -26,7 +26,7 @@ describe('os.filter.impl.ecql.ExclusionFormatter', function() {
     expect(formatter.wrapMultiple(result)).toBe(
       '(' +
       '(DISJOINT(geometry,POLYGON((0 1,0 3,2 3,2 1,0 1))))' +
-      ' OR ' +
+      ' AND ' +
       '(DISJOINT(geometry,POLYGON((4 5,4 7,6 7,6 5,4 5))))' +
       ')');
   });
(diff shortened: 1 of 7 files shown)


test/os/filter/impl/ecql/exclusionformatter.test.js
```diff
@@ -0,0 +1,33 @@
...
+    var formatter = new os.filter.impl.ecql.ExclusionFormatter();
+    var f = new ol.Feature(ol.geom.Polygon.fromExtent([0, 1, 2, 3]));
+    var result = formatter.format(f);
+    var f2 = new ol.Feature(ol.geom.Polygon.fromExtent([4, 5, 6, 7]));
+    result += formatter.format(f2);
+
+    expect(formatter.wrapMultiple(result)).toBe(
+      '(' +
+      '(DISJOINT(geometry,POLYGON((0 1,0 3,2 3,2 1,0 1))))' +
+      ' OR ' +
```
bradh: If I'm reading this right, if we have a geometry within one of the exclusion areas (say `POINT(1.5 2.5)` which is within `f`) then this filter will return true.

There is probably no right answer (users being users...) but I would have assumed that given two exclusion areas, it would be appropriate to exclude anything that intersects either of them. Right now, no points will be excluded (because no point can be disjoint to both).

At least I think so - this double negative stuff is doing my head in..

wallw-bits: That's not really that function's concern. If there's a bug with the exclusion condition, then it is in `os.ui.filter.QueryHandler.prototype.wrap()`.

Your confusion is why I wrote #209.

PR #226, opened 2018-04-04
feat(addlayer): Adds a page that adds layers from links

wallw-bits:
Added a separate `addlayer.html` page that is built separately from the main page and takes a URI-encoded JSON string in the fragment and sends it over to OpenSphere (or navigates to it if it is not up). The JSON is just a layer config or array of layer configs.

`os.MainCtrl` was modified to not init the XT peer until after plugins are loaded.

`os.xt.Peer` was modified to check for messages it can handle on startup. This allows the "drop a message and then navigate to the app" flow.

The compile portion is admittedly a little hacky since we don't have any notion of separate compiles for different pages within an app. It adds ~7 seconds to the build. It also does not have a debug compile (so any changes to `plugin.openpage.Page` require at least `yarn run compile:gcc:addlayer`; any changes to `addlayer-template.html` requires `yarn run build:debugindex`).

Compile advantages over static JS:
- Don't have to recreate any of `os.xt.Peer` (which we could admittedly export or even just use direct `localStorage` calls since there's really only two)
- The `os.NAMESPACE` define is set properly. This allows the page to work properly for plugins that override the value (which I suppose we could otherwise search/replace in the build)


wallw-bits:
Agreed. I'm also experimenting with skipping the extra compile in favor of simply replacing the values in the script.

src/os/xt/peer.js
```diff
@@ -481,6 +484,26 @@ os.xt.Peer.prototype.init = function() {
...
+ */
+os.xt.Peer.prototype.processInitialMessages = function() {
+  var priv = ['xt', this.group_, this.id_, ''].join('.');
+  var pub = ['xt', this.group_, 'public', ''].join('.');
+
+  var storage = this.storage_;
+  var i = storage.length;
+  while (i--) {
+    var key = storage.key(i);
+    if (key && (key.startsWith(pub) || (key.startsWith(priv) && os.xt.Peer.notThese_.indexOf(key) === -1))) {
```
schmidtk: The `os.xt.Peer.notThese_.indexOf(key) === -1` check here should be looking for the last part of the key in that array, not the entire key. Currently that condition will always be true and extra messages will be processed.

PR #227, opened 2018-04-08
chore(cesium): Update Cesium to 1.43.0, and ol-cesium to match (1.35.0).

bradh:
No changes to OpenLayers for ol-cesium 1.35.

The changes to Cesium are in https://github.com/AnalyticalGraphicsInc/cesium/blob/1.43/CHANGES.md#143---2018-03-01 - the breaking changes are either not in code we use, or are independent of the code (like deprecated sources). Downstream users of OS need to check those though.

I've smoke tested this - seems fine. There is one additional console logging message, about deprecated STK Terrain. That seems to be a statement about the data rather than about the code though, per the change log.


schmidtk:
WRT STK Terrain, we'll have to remove that by September 1, 2018 when they get rid of that provider completely. They're replacing it with [Cesium World Terrain](https://cesium.com/blog/2018/03/01/introducing-cesium-world-terrain/), which is part of their new Cesium ion subscription based terrain/3D tile platform.

bradh:
Should we have a separate ticket for the terrain migration (which I think we'll hit even if we don't apply this change)?

schmidtk:
Yes, added #229 to remove the deprecated terrain provider and look into Cesium World Terrain support. The latter will eventually be subscription based, so we'll have to add support for configuring a Cesium Ion account key.

bradh:
If I understood what you wanted, it should now be incorporated.

src/os/mixin/cesiummixin.js
```diff
@@ -172,11 +172,10 @@ os.mixin.cesium.loadCesiumMixins = function() {
     image.src = url;
   };
 
-
   /**
    * @suppress {accessControls|duplicate}
    */
-  Cesium.loadImage.defaultCreateImage = Cesium.loadImage.createImage;
+  Cesium.Resource._Implementations.createImage = Cesium.loadImage.createImage;
```
schmidtk: One minor additional request. Can you set the above function to a local var and remove the suppression in the docs? This code will otherwise break when `Cesium.loadImage` is removed. `Cesium.loadImage.createImage` is no longer used internally by Cesium and we don't use it either, so there is no longer a need to replace the function.

schmidtk:
Realized the trivial change I was requesting was just as easy to submit back myself. Sorry about that!

PR #228, opened 2018-04-11
feat(storage): add persistent storage permission check

wallw-bits:
Browsers have internal limits for storage that may be cleared when under storage pressure. However, a recently-added permission allows the storage to be permanent.

This currently asks for the permission on every load (which might be overly aggressive).

See https://developers.google.com/web/updates/2016/06/persistent-storage for details.

## Testing
1. Serve from `http://` (not localhost) and verify that there are messages in the logs stating that persistent storage is not permitted in an unsecure environment
1. Firefox: Serve from `https://` (not localhost) and verify that the user is prompted for a persistent storage permission
    * Verify that declining the permission produces a warning in both the toast alerts and in the log
    * Verify that accepting the permission produces a successful log message and that `navigator.storage.persist()` resolves to `true` afterward
1. Chrome: Serve from `https` on a hostname that you don't hit often
    * Verify that it asks for notification permissions, and that accepting the permissions results in `navigator.storage.persist()` resolving to `true`


justin-bits:
This build is failing due to `.travis.yml` referencing a file that doesn't exist yet on this branch.  Get the latest from master to resolve.

src/plugin/storage/persistplugin.js
```diff
@@ -0,0 +1,131 @@
+goog.module('plugin.storage.PersistPlugin');
+goog.module.declareLegacyNamespace();
+
+const browser = goog.require('goog.labs.userAgent.browser');
+const log = goog.require('goog.log');
+const AlertEventSeverity = goog.require('os.alert.AlertEventSeverity');
+const AbstractPlugin = goog.require('os.plugin.AbstractPlugin');
+const Metrics = goog.require('os.metrics.Metrics');
+const {launchConfirm} = goog.require('os.ui.window');
```
schmidtk: This line will break when `os.ui.window` is converted to a module, because `launchConfirm` is in a different file (`confirm.js`). It currently only works because `goog.require` returns the namespace if requiring a `goog.provide` file.

We may want to avoid using the return value of `goog.require` for old code, both for this reason and because it's clear at a glance what remaining deps are not modules.

src/plugin/storage/persistplugin.js
```diff
@@ -0,0 +1,131 @@
...
+        } else {
+          log.info(LOGGER_, 'Storage is persistent');
+          metrics.updateMetric(PERSISTENCE_SUCCEEDED, 1);
+        }
+      });
+    } else {
+      if (window.isSecureContext === false) {
+        msgs.push('The application is not being served in a secure context. Persistent storage permission ' +
+            'will be automatically declined.');
+      } else if (browser.isChrome()) {
```
schmidtk: I think this check needs to be in the `if (!userDecision)` block. In Chrome, `navigator.storage.persist()` will be called, but Chrome makes that decision without user input.

src/plugin/storage/persistplugin.js
```diff
@@ -0,0 +1,131 @@
...
+        msgs.push('Chrome automatically manages the persistent storage permission based on the following ' +
+          'criteria (<a href="https://developers.google.com/web/updates/2016/06/persistent-storage" ' +
+          'target="_blank">source</a>):' +
+          '<ul><li>The site is bookmarked (and the user has 5 or less bookmarks)</li>' +
+          '<li>The site has high site engagement [not quantified]</li>' +
+          '<li>The site has been added to home screen</li>' +
+          '<li>The site has push notifications enabled</li></ul> Try bookmarking the application or ' +
```
schmidtk: Should we automatically request enabling push notifications? Users may not know how to enable them manually, and doing so could help the storage issue.

wallw-bits:
@schmidtk I added some detection on whether the result of the promise was a user decision or automatically resolved based on timing. Also added request for Notification permissions in Chrome.

wallw-bits:
`navigator.storage` isn't writable, so mocking it for tests is difficult. I could add code to allow passing in a mock `StorageManager` for tests. Would that be preferable or would we rather avoid the test-only code?

PR #240, opened 2018-04-26
fix(capture): Scale renderers to match map DPI.

### LINKED ISSUE #239: Capture does not scale canvases appropriately on high DPI displays
When capturing screenshots/recordings, OpenSphere should normalize input canvases with varying pixel ratios so the final image represents what is displayed on screen.

Current canvas pixel ratios are as follows:
- OpenLayers canvases (map and overlay) both use `devicePixelRatio`.
- Cesium canvas has a fixed pixel ratio of 1 and uses `image-rendering` to scale the image, when supported by the browser (Chrome/Firefox support it, IE/Edge do not).
- Timeline canvas created from the SVG has a pixel ratio of 1.
- Legend canvas has a fixed pixel ratio of 1.

Which causes the following problems:
- Overlay is not sized correctly when capturing with 3D mode active.
- Timeline/legend are not sized/positioned correctly when capturing with 2D mode active.

schmidtk:
Fixes #239.

### DIFF
--- a/src/os/capture/capture.js
+++ b/src/os/capture/capture.js
@@ -370,3 +370,13 @@ os.capture.isTainted = function(canvas) {
   }
   return false;
 };
+
+
+/**
+ * Get the pixel ratio for the canvas created by capture. Override this function in capture plugins to replace the
+ * default behavior.
+ * @return {number} The pixel ratio for the output canvas.
+ */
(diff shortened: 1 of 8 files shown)


wallw-bits:
Capture output looks good.

PR #244, opened 2018-05-07
chore(cesium): Update Cesium to 1.45 and fix deprecated API's.

schmidtk:
This is the last `ol-cesium` update we'll get unless we update OpenSphere to support OpenLayers 5.x and its ES6 modules. All future `ol-cesium` releases will also use ES6 modules.


schmidtk:
@jsalankey @wallw-bits - rebased to resolve conflicts with the Cesium plugin changes, and fix tests breaking due to lolex. All good now.

PR #254, opened 2018-06-06
feat(featureType): Add and integrate interface for featureType

welchyd:
@schmidtk  - this is the interface we discussed. Let me know what you think!


src/plugin/ogc/ui/ogclayernodeui.js
```diff
@@ -45,15 +45,15 @@ plugin.ogc.ui.OGCLayerNodeUICtrl = function($scope, $element) {
...
+  this.descriptor_ = os.dataManager.getDescriptor(this.getLayerId());
 
   var chooseTime = false;
 
-  if (this.descriptor_) {
+  if (os.implements(this.descriptor_, os.ui.ogc.IFeatureTypeDescriptor.ID)
+      && os.implements(this.descriptor_, os.ui.ogc.wms.IWMSLayer.ID)) {
     var featureType = this.descriptor_.getFeatureType();
     if (featureType) {
       chooseTime = this.descriptor_.hasTimeExtent() && featureType.getTimeColumns().length >= 2 &&
```
schmidtk: I suspect the test for `hasTimeExtent` is really just making sure the start or end date column name isn't null ([this line](https://github.com/ngageoint/opensphere/blob/master/src/plugin/ogc/ogclayerdescriptor.js#L817) in the OGC descriptor).

I think we can replace that with the conditions in the linked line, and remove the `os.implements(this.descriptor_, os.ui.ogc.wms.IWMSLayer.ID)` test above.

src/os/ui/ogc/ifeaturetypedescriptor.js
```diff
@@ -0,0 +1,30 @@
+goog.provide('os.ui.ogc.IFeatureTypeDescriptor');
+
+
+
+/**
+ * Interface for OGC feature type.
```
schmidtk: Update this to something like:
```
Interface for descriptors providing an `os.ogc.IFeatureType`.
```

src/os/ui/ogc/ifeaturetypedescriptor.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @type {string}
+ * @const
+ */
+os.ui.ogc.IFeatureTypeDescriptor.ID = 'os.ui.ogc.IFeatureTypeDescriptor';
+
+
+/**
+ * Get the WFS FeatureType.
```
schmidtk: This isn't restricted to WFS, so you can just say "FeatureType". Arc is using this interface as well.

welchyd:
Thanks! Added those changes.

PR #257, opened 2018-06-13
Cesium should use OpenLayers tile stack

wallw-bits:
- The `ImageryProvider` now extends the underlying `OLImageryProvider` from ol-cesium, preventing a lot of re-work.
- `ImageryProvider.prototype.requestImage` now uses the OpenLayers tile loading stack. This eliminates the duplicated load counting and tile filter functions from this class.
- `TileGridTilingScheme` was added. This calls into the source's `TileGrid` instance for all the relevant information needed to tile a Cesium imagery layer. Note: Cesium does not support `TileGrid`'s ability to have multiple origins.

All of this enables highly custom raster imagery plugins, such as GeoPackage, to work in Cesium without two separate implementations for OpenLayers and Cesium.

# Testing
I initially attempted a unit test for this but it was subject to some rounding error that was not occurring during in-app tests.

Therefore, pop up the dev console and run `cesiumImageryTest()` (even in compiled mode). This will copy out all of your base maps to the old imagery provider and put them on top with a 50% opacity, allowing a visual comparison. Test high zoom levels and tiles near the end of the layer's extent (generally the antimeridian for base maps). You're looking for them to line up perfectly everywhere.

Also test:
- Tiled layers with limited extents
- Animated tile layers in the timeline


wallw-bits:
@schmidtk, before we merge this I want to take out the old imagery provider and the test code that duplicates the base maps.

schmidtk:
Yep great idea. I'll take another look and let you do the cleanup/merge when it's ready.

PR #265, opened 2018-07-08
perf(cesium): Use flat coords for multi-geom translation

### LINKED ISSUE #135: Improve Cesium multi-geometry translation

wallw-bits:
All geom types now read from flat coordinates using the given offset, stride, and ends. Scratch coordinate arrays are used to prevent array creation during geometry translation. The greatest performance enhancement comes from Multi geoms, which previously called `multi.getGeometries()` which re-inflates the sub geos from the flat coordinates. Since we now use the flat coords directly, it is both faster and more memory-efficient.

I'm not overly happy with the fix, largely because `FeatureConverter` could use a design refactor in the first place.

Resolves #135

## Testing
* For every geometry type (including `MultiPoint`, `MultiLineString`, and `MultiPolygon`)
    * Ensure it is rendered identically to master
    * Ensure it can be picked/highlighted
    * Bring up the timeline and ensure it is rendered correctly and can be picked/highlighted
* Test KML track animation
* Test Hide/Show feature both in and out of the timeline


wallw-bits:
I'm leaving this as is until I can figure out why the build is passing locally. The error in Travis is legit

src/plugin/cesium/sync/featureconverter.js
```diff
@@ -624,6 +620,43 @@ plugin.cesium.sync.FeatureConverter.prototype.updatePolyline = function(feature,
...
+
+  var positions = new Array((end - offset) / stride);
+  var count = 0;
+  for (var i = offset; i < end; i += stride) {
+    for (var j = 0; j < stride; j++) {
+      coord[j] = flats[i + j];
+      transformedCoord[j] = coord[j];
+    }
+
+    transform(coord, transformedCoord, stride);
```
schmidtk: Need a null check here. `getTransformFunction` will return null when in 4326.

src/plugin/cesium/sync/featureconverter.js
```diff
@@ -826,47 +859,130 @@ plugin.cesium.sync.FeatureConverter.prototype.olEllipseGeometryToEllipsoid = fun
...
+      for (var j = 0; j < stride; j++) {
+        coord[j] = flats[i + j];
+        transformedCoord[j] = coord[j];
+      }
 
-  for (var i = 0; i < rings.length; ++i) {
-    var olPos = rings[i].getCoordinates();
-    var csPos;
-    var extent = rings[i].getExtent();
+      transform(coord, transformedCoord, coord.length);
```
schmidtk: Need a null check on `transform` here as well.

schmidtk:
Load the following file:
[geometrytest.txt](https://github.com/ngageoint/opensphere/files/2178317/geometrytest.txt)

Comparing to master, I noticed the following:

- When hovering the MultiPolygons the geometries flicker, sometimes very notably.
polygonflicker
- When hovering the MultiLineString, the middle line disappears and the left line remains red after hover. The flicker behavior is also present.
multilinehover

wallw-bits:
That last commit should only recreate the primitives from the flats if the geometry revision changes, which fixes the flickering on hover of the file you posted.

That still doesn't address the KML track animation though, I believe you said that the flats/ends are being modified without a geometry revision change.

wallw-bits:
@ksgolding I've updated this with the latest from master. The only outstanding issue was the aforementioned KML Track animation issue, which stems from changing the flats/ends on the geometry without updating the geometry revision. We should discuss that with @schmidtk to come up with a good solution for that case and then we can merge this.

wallw-bits:
Before this is merged, I'm going to put together some performance tests. I am concerned that the point performance may have suffered slightly, and that is far and away our top use case. The multi geoms all got a massive performance and memory boost.

PR #271, opened 2018-07-19
feat(layer): GOTO is now supported for tile layers

wallw-bits:
Go To is now enabled for Tile layers which have a reduceable extent which is contained by the overall projection extent and is less than 80% of the overall projection extent area.

... and it just occurred to me that this probably fails for extents crossing the anti-meridan.


wallw-bits:
I removed the contains check since that breaks over the anti-meridian.

Issue #157, opened 2018-02-28
Implement 2D mode contrast, saturation and brightness controls


jsalankey:
Since Openlayers does not natively support these functions, they were taken out a long time ago and only shown in 3D since Cesium does natively support them. 

Tile coloring is implemented through the `IFilterableTileSource` interface and mixins to `ol.source.TileImage`. We can implement these 3 features as filters run through that interface.

This should also be able to clean up some of the layer UI checks for whether the map is in 3D mode.

schmidtk:
Resolved in #273.

PR #273, opened 2018-08-02
fix(2D): Added saturation, brightness, and contrast to 2D mode

lilysikes:
resolves Issue #157

### DIFF
--- a/src/os/color.js
+++ b/src/os/color.js
@@ -740,6 +740,41 @@ os.color.colorize = function(data, tgtColor) {
   }
 };
 
+/**
+ * Applies a color transform to an array of image data. This transform takes a target color and adjusts the color
+ * for saturation color and brightness.
+ * @param {Array<number>} data The image data to colorize
+ * @param {number} brightness The target brightness. The range is -1 to 1.
+ * @param {number} contrast The target contrast. The range is 0 to 2.
+ * @param {number} saturation The target saturation. The range is 0 to 1.
+ */
+os.color.adjustColor = function(data, brightness, contrast, saturation) {
+  var intercept = (1 - contrast) / 2;
+  var sr = (1 - saturation) * 0.3086;
+  var sg = (1 - saturation) * 0.6094;
+  var sb = (1 - saturation) * 0.0820;
+  brightness = brightness * 225;
+
+  // color transform matrix for contrast and saturation, taken from online research https://docs.rainmeter.net/tips/colormatrix-guide/
+  var m = [
+    contrast * (sr + saturation), contrast * sr, contrast * sr, 0, 0,
+    contrast * sg, contrast * (sg + saturation), sg * contrast, 0, 0,
+    contrast * sb, contrast * sb, contrast * (sb + saturation), 0, 0,
+    0, 0, 0, 1, 0,
+    intercept + brightness, intercept + brightness, intercept + brightness, 1, 0
+  ];
+
+
+  for (var i = 0, n = data.length; i < n; i += 4) {
+    var r = data[i];
+    var g = data[i + 1];
+    var b = data[i + 2];
+    var a = data[i + 3];
+    data[i] = (r * m[0] + g * m[5] + b * m[10] + a * m[15] + m[20]);
+    data[i + 1] = (r * m[1] + g * m[6] + b * m[11] + a * m[16] + m[21]);
+    data[i + 2] = (r * m[2] + g * m[7] + b * m[12] + a * m[17] + m[22]);
+  }
+};
 
 /**
  * Applies a color transform to an array of image data. This transform takes a source and target color and blends
--- a/src/os/layer/tile.js
+++ b/src/os/layer/tile.js
@@ -276,6 +276,42 @@ os.layer.Tile.prototype.getColor = function() {
   return null;
 };
 
+/**
+ * Get the brightness for the tile layer.
+ * @return {number}
+ * @override
+ */
+os.layer.Tile.prototype.getBrightness = function() {
+  if (this.layerOptions_) {
+    return /** @type {number} */ (this.layerOptions_['brightness'] || 0);
+  }
+  return 0;
+};
+
+/**
+ * Get the brightness for the tile layer.
+ * @override
+ * @return {number}
+ */
+os.layer.Tile.prototype.getContrast = function() {
+  if (this.layerOptions_) {
+    return /** @type {number} */ (this.layerOptions_['contrast'] || 1);
+  }
+  return 1;
+};
+
+/**
+ * Get the saturation for the tile layer.
+ * @override
+ * @return {number}
+ */
+os.layer.Tile.prototype.getSaturation = function() {
+  if (this.layerOptions_) {
+    return /** @type {number} */ (this.layerOptions_['saturation'] || 1);
+  }
+  return 1;
+};
+
 
 /**
  * Get the whether the tile layer is being colorized.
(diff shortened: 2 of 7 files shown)


src/os/color.js
```diff
@@ -785,11 +801,20 @@ os.color.transformColor = function(data, srcColor, tgtColor) {
   ];
 
   for (var i = 0, n = data.length; i < n; i += 4) {
-    var r = data[i];
-    var g = data[i + 1];
-    var b = data[i + 2];
     var a = data[i + 3];
 
+    // Applies saturation, brightness and contrast to transformed color
+    var hslColor = goog.color.rgbArrayToHsl([data[i], data[i + 1], data[i + 2], data[i + 3]]);
```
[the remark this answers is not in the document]
schmidtk: You don't need to create a new array here. `goog.color.rgbArrayToHsl` is a wrapper for `goog.color.rgbToHsl`, which takes the RGB values as separate params. This should be:
`var hslColor = goog.color.rgbToHsl(data[i], data[i + 1], data[i + 2]);`

src/os/layer/tile.js
```diff
@@ -326,6 +362,46 @@ os.layer.Tile.prototype.setColor = function(value, opt_options) {
   }
 };
 
+/**
+ * Set the brightness for the tile layer.
+ * @param {?number} value The new color
+ * @param {Object=} opt_options The layer options to use
+ */
+os.layer.Tile.prototype.changeBrightness = function(value, opt_options) {
```
schmidtk: The set calls for brightness/contrast/saturation are defined by OpenSphere in `layerbasemixin.js` and are only used for tile layers. We put them in that mixin after OpenLayers removed the functions from their base layer class, but we never used them for anything but tile layers. It would be a nice code cleanup pass to:

1. Remove the mixin (get and set calls).
2. Use `set*` instead of `change*` here for consistency.
3. Move the related functions in `os.ui.layer.DefaultLayerUICtrl` to `os.ui.layer.TileLayerUICtrl`.

Thoughts on this @wallw-bits?

schmidtk:
A few things I noticed in testing:
- Brightness, contrast, and saturation are only applied if the default color is changed or Colorize is checked. This is due to the color filter being removed from the source unless at least one of those is true.
- The end result is different in 2D and 3D because `cesium.js` is still adjusting values on the Cesium layer, in addition to the transform done in the original tile layer. 2D and 3D tiles should look the same.
- Changing the values in 3D mode causes excessive flickering due to reloading the tile textures to the GPU. It would be preferred for Cesium to continue adjusting these values directly on the Cesium layer instead of transforming tiles, for performance reasons. 2D mode should do the transform, and the result should fairly closely match what we see in Cesium.
- Translating each pixel from RGB to HSL and back is responsible for 40%+ of the CPU trace when adjusting these tile controls. Can we use a transform matrix against the RGB values to avoid that step? We do that for the color transform, so maybe we can do something similar for brightness/contrast/saturation. I'm not an expert on color transforms, so I poked around a little and most searches point toward this guide: https://docs.rainmeter.net/tips/colormatrix-guide/

src/os/layer/tile.js
wallw-bits: I agree overall, but what about raster layers that don't use tiles? KML Ground Overlays come to mind (`os.layer.Image`).

schmidtk: I guess the base mixin can stay so all layers have those functions, and tile/image layers would have to override them for their specific implementation.

src/os/color.js
```diff
@@ -740,6 +740,37 @@ os.color.colorize = function(data, tgtColor) {
...
+    contrast * sg, contrast * (sg + saturation), sg * contrast, 0, 0,
+    contrast * sb, contrast * sb, contrast * (sb + saturation), 0, 0,
+    0, 0, 0, 1, 0,
+    intercept + brightness, intercept + brightness, intercept + brightness, 1, 0
+  ];
+
+
+  for (var i = 0, n = data.length; i < n; i += 4) {
+    data[i] = (data[i] * m[0] + data[i + 1] * m[5] + data[i + 2] * m[10] + data[i + 3] * m[15] + m[20]);
+    data[i + 1] = (data[i] * m[1] + data[i + 1] * m[6] + data[i + 2] * m[11] + data[i + 3] * m[16] + m[21]);
```
wallw-bits: Is `data[i]` on the right side of this expression intended to be the already-recalculated `data[i]` or the original `data[i]`?

Same for that and also `data[i + 1]` on the next line.

lilysikes: It should probably be the original I will fix that.

src/os/layer/tile.js
```diff
@@ -326,6 +362,56 @@ os.layer.Tile.prototype.setColor = function(value, opt_options) {
   }
 };
 
+/**
+ * Adjust the layer brightness.  A value of -1 will render the layer completely
+ * black.  A value of 0 will leave the brightness unchanged.  A value of 1 will
+ * render the layer completely white.  Other values are linear multipliers on
+ * the effect (values are clamped between -1 and 1).
+ *
+ * @param {number} value The brightness of the layer.
```
wallw-bits: The description at the top is good, but it would be better to also have a short version on the `value` parameter itself.

Also, consider having the function check that the value is in the proper range via `goog.assert` would be nice. Even though the underlying filter function clamps the values properly, a check like that will alert developers that they are sending something incorrect (like `[0, 100]` from a slider or spinner).

Ditto for the rest of the setters in this file.

conformance_config.textproto
```diff
@@ -28,13 +28,10 @@ requirement: {
   error_message: 'Assignment to Element.prototype.innerHTML is not allowed '
```
wallw-bits: This file should not be a part of this PR

externs/angular-os.js
```diff
@@ -4,7 +4,6 @@
  * @externs
```
wallw-bits: This file should not be a part of this PR

src/os/net/request.js
```diff
@@ -153,7 +153,6 @@ os.net.Request = function(opt_uri, opt_method) {
 };
```
wallw-bits: This file should not be a part of this PR

wallw-bits:
@schmidtk's comment needs to be addressed:
> Brightness, contrast, and saturation are only applied if the default color is changed or Colorize is checked. This is due to the color filter being removed from the source unless at least one of those is true.

That's occurring on `os.layer.Tile:440`. You'll need to add some expressions to that check to test for non-default brightness, contrast, and saturation as well.

wallw-bits:
@lilysikes check out `yarn test` on your branch. The tests you added have a couple of failures and also one other test is failing due to this ticket.

The failures related to the state XSD should be addressed by ngageoint/opensphere-state-schema#8.

As soon as those are addressed then I think this is good to merge.

src/os/layer/tile.js
```diff
@@ -364,6 +471,7 @@ os.layer.Tile.prototype.applyColors = function(data) {
         // transformColor blends between the src and target color, leaving densitization intact
         os.color.transformColor(data, srcColor, tgtColor);
```
schmidtk: We should skip transforming the color if the layer is using the default color. Changing brightness/contrast/saturation without changing the layer color should adjust those values against the original tiles. Currently it re-colors the tiles white, then makes the adjustment.

src/os/color.js
```diff
@@ -770,9 +770,24 @@ os.color.adjustColor = function(data, brightness, contrast, saturation) {
     var g = data[i + 1];
     var b = data[i + 2];
     var a = data[i + 3];
-    data[i] = (r * m[0] + g * m[5] + b * m[10] + a * m[15] + m[20]);
-    data[i + 1] = (r * m[1] + g * m[6] + b * m[11] + a * m[16] + m[21]);
-    data[i + 2] = (r * m[2] + g * m[7] + b * m[12] + a * m[17] + m[22]);
+    data[i] = Math.round((r * m[0] + g * m[5] + b * m[10] + a * m[15] + m[20]));
+    data[i + 1] = Math.round((r * m[1] + g * m[6] + b * m[11] + a * m[16] + m[21]));
+    data[i + 2] = Math.round((r * m[2] + g * m[7] + b * m[12] + a * m[17] + m[22]));
+    if (data[i] < 0) {
```
wallw-bits: These blocks can be simplified using `Math.min(max, Math.max(min, value))` or `goog.math.clamp(value, min, max)`

src/os/color.js
```diff
@@ -770,9 +770,24 @@ os.color.adjustColor = function(data, brightness, contrast, saturation) {
...
+      data[i] = 0;
+    } else if (data[i] > 225) {
```
wallw-bits: 225? Did you mean 255 (0xFF)?

src/os/color.js
```diff
@@ -781,9 +781,12 @@ os.color.adjustColor = function(data, brightness, contrast, saturation) {
     var g = data[i + 1];
     var b = data[i + 2];
     var a = data[i + 3];
-    data[i] = (r * m[0] + g * m[5] + b * m[10] + a * m[15] + m[20]);
-    data[i + 1] = (r * m[1] + g * m[6] + b * m[11] + a * m[16] + m[21]);
-    data[i + 2] = (r * m[2] + g * m[7] + b * m[12] + a * m[17] + m[22]);
+    data[i] = Math.round((r * m[0] + g * m[5] + b * m[10] + a * m[15] + m[20]));
+    data[i + 1] = Math.round((r * m[1] + g * m[6] + b * m[11] + a * m[16] + m[21]));
+    data[i + 2] = Math.round((r * m[2] + g * m[7] + b * m[12] + a * m[17] + m[22]));
+    goog.math.clamp(data[i], 0, 255);
```
schmidtk: Clamp doesn't modify the value in-place, it returns the clamped value. You'll want to do the clamp on the previous lines.

src/os/layer/tile.js
```diff
@@ -295,7 +295,9 @@ os.layer.Tile.prototype.getBrightness = function() {
  */
 os.layer.Tile.prototype.getContrast = function() {
   if (this.layerOptions_) {
-    return /** @type {number} */ (this.layerOptions_['contrast'] || 1);
+    if (this.layerOptions_['contrast'] != null) {
```
schmidtk: As minor cleanup, this if condition can be &&'d with the previous one. Same with the block in `getSaturation` below.

PR #276, opened 2018-08-06
Adds altitude to 'Add Place' menu

### LINKED ISSUE #183: Add altitude to places
Places supports fields for location, icon rotation, and even ellipse parameters so adding altitude seems like a natural fit. Would be useful for denoting something airborne or at a certain height

arwatts16:
resolves #183


src/os/ui/featureedit.js
```diff
@@ -980,8 +1010,20 @@ os.ui.FeatureEditCtrl.prototype.loadFromFeature_ = function(feature) {
       if (coordinate) {
         this['pointGeometry'] = {
           'lon': coordinate[0],
-          'lat': coordinate[1]
+          'lat': coordinate[1],
+          'alt': coordinate[2]
         };
+
+        var altU = this['options']['feature']['values_']['ALTITUDE_UNITS'];
```
schmidtk: `this['options']['feature']` is the same object as the `feature` param to this function. Use that instead, and retrieve the value of `altU` with `feature.get(os.Fields.ALT_UNITS)`.

src/os/ui/featureedit.js
```diff
@@ -980,8 +1010,20 @@ os.ui.FeatureEditCtrl.prototype.loadFromFeature_ = function(feature) {
...
+
+        // else if runs when the page is refreshed
+        if (coordinate[2] !== null && coordinate[2] !== undefined && altU !== null && altU !== undefined) {
```
schmidtk: `val != null` is functionally equivalent to `val !== null && val !== undefined`. The former uses loose equality, and `null == undefined` in JavaScript.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

src/os/fields/fields.js
```diff
@@ -8,6 +8,7 @@ goog.provide('os.fields');
 os.Fields = {
   ALT: 'ALTITUDE',
   ALT_UNITS: 'ALTITUDE_UNITS',
+  ALTITUDE_INPUT: 'ALTITUDE_INPUT',
```
schmidtk: The user altitude input will get saved to the geometry, and doesn't need to be saved again as a property on the feature. You can remove this field and all references to it.

src/os/ui/featureedit.js
```diff
@@ -317,6 +319,12 @@ os.ui.FeatureEditCtrl = function($scope, $element, $timeout) {
     os.math.Units.METERS
   ];
 
+  /**
+   * Units for altitude
+   * @type {string}
+   */
+  this['altOptions'] = goog.object.getKeys(os.math.Units);
```
schmidtk: Use the values of `os.math.Units` instead. The keys will be minified by the Closure compiler, resulting in confusing options for the user.

For an example of another UI with altitude units, see `src/os/ui/wiz/geometrystep.js`.

src/os/ui/featureedit.js
```diff
@@ -847,6 +862,21 @@ goog.exportProperty(
     'updatePreview',
     os.ui.FeatureEditCtrl.prototype.updatePreview);
 
+/**
+ * Updates the altitude units.
+ * @param {*} value
+ */
+os.ui.FeatureEditCtrl.prototype.updateAltUnits = function(value) {
```
schmidtk: Is this still used? I don't see any reference to it in the PR. If not, please remove it.

src/os/ui/featureedit.js
```diff
@@ -637,6 +651,7 @@ os.ui.FeatureEditCtrl.prototype.accept = function() {
     });
   }
 
+
```
schmidtk: Unnecessary extra blank line.

src/os/ui/featureedit.js
```diff
@@ -980,8 +1010,20 @@ os.ui.FeatureEditCtrl.prototype.loadFromFeature_ = function(feature) {
       if (coordinate) {
         this['pointGeometry'] = {
           'lon': coordinate[0],
-          'lat': coordinate[1]
+          'lat': coordinate[1],
+          'alt': coordinate[2]
```
schmidtk: Use `coordinate[2] || 0` to ensure you get a numeric value. If a coordinate does not specify an altitude, you can assume a value of 0.

src/os/ui/featureedit.js
```diff
@@ -1154,9 +1196,21 @@ os.ui.FeatureEditCtrl.prototype.saveGeometry_ = function(feature) {
     // make sure the coordinate values are numeric
     var lon = Number(this['pointGeometry']['lon']);
     var lat = Number(this['pointGeometry']['lat']);
+    var alt = Number(this['altitude']) ? this['altitude'] : null;
```
schmidtk: You can assume 0 for null/undefined altitude here as well, so `Number(this['altitude']) || 0`.

For the units, use values from `os.math.Units` instead of keys, changing the statement to `var altU = this['altUnits'] || os.math.Units.METERS;`.

Given those two changes, you can always set altitude on the coordinate, do the unit conversion, etc, instead of checking if both values are defined first.

schmidtk:
Created a feature branch and cleaned up in #285.

PR #277, opened 2018-08-08
Add Sound Action to Features

brelsftm:
Adds ability to notify user from a sound action.


wallw-bits:
Looks like you may have accidentally changed all the file permissions. Can you please fix that and either amend the commit or make a new PR?

PR #278, opened 2018-08-08
fix(wizard): Checked for null case in destroy function

### LINKED ISSUE #263: null config in wizard steps
Noticed an error in the dev tools console that the finalize() call in TimeStep (extends AbstractWizardStep) was getting null for the config input parameter then we were trying to pull things from the config. Not sure if this is representative of a larger issue where config should not be null or if we're just not handling null in the finalize calls in the wizard steps.

lilysikes:
I think that all that was needed for this was just to check if config is null in the destroy function.

Resolves #263

### DIFF
--- a/src/os/ui/wiz/wizard.js
+++ b/src/os/ui/wiz/wizard.js
@@ -120,7 +120,7 @@ os.ui.wiz.WizardCtrl = function($scope, $element, $timeout, $attrs) {
  */
 os.ui.wiz.WizardCtrl.prototype.destroy_ = function() {
   var step = this['steps'][this['activeIndex']];
-  if (!step.isDeactivated() && this.config) {
+  if (!step.isDeactivated()) {
     step.deactivate(this.config);
   }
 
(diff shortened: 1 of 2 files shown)


src/os/ui/wiz/wizard.js
```diff
@@ -120,7 +120,7 @@ os.ui.wiz.WizardCtrl = function($scope, $element, $timeout, $attrs) {
  */
 os.ui.wiz.WizardCtrl.prototype.destroy_ = function() {
   var step = this['steps'][this['activeIndex']];
-  if (!step.isDeactivated()) {
+  if (!step.isDeactivated() && this.config) {
```
wallw-bits: The real underlying problem here is that the wizard steps support generics, but the wizard itself is not defined with the necessary generics type info to pass on to the steps. That in itself is outside the scope of this issue. However, checking for null config here in `wizard.js` prevents `IWizardStep` instances defined with `null` types from being used to their full extent.

In other words, instead of doing this here, I'd rather see the check for null config inside each `step.deactivate()` since that will allow some types to support null config if they want to.

lilysikes: Ok that makes sense!

welchyd:
I read this and wonder if the class that extends this is where the issue is at since that class defines T. I could make a class that uses `@extends {os.ui.somewhere.Name<?some.type>}` and these changes would prevent `finalize` from being run if `config` was undefined, even though I was allowing for that possibility when creating the class.

PR #281, opened 2018-08-14
Add Sound Action to Features

brelsftm:
Adds ability to notify user from a sound action.


src/plugin/featureaction/featuresoundaction.js
```diff
@@ -0,0 +1,193 @@
...
+/**
+ * @inheritDoc
+ */
+plugin.im.action.feature.SoundAction.prototype.execute = function(items) {
+  for (var i = 0; i < items.length; i++) {
+    var item = items[i];
+    if (item) {
+      var featureConfig = /** @type {Array|Object|undefined} */ (item.get(
+          os.style.StyleType.FEATURE)) || {};
+      if (goog.isArray(featureConfig)) {
```
schmidtk: The feature config is for styling only. Don't add the sound to it.

src/plugin/featureaction/featuresoundaction.js
```diff
@@ -0,0 +1,193 @@
...
+        for (var j = 0; j < featureConfig.length; j++) {
+          os.style.mergeConfig(this.soundConfig, featureConfig[j]);
+        }
+      } else {
+        os.style.mergeConfig(this.soundConfig, featureConfig);
+      }
+
+      item.set(os.style.StyleType.FEATURE, featureConfig, true);
+      item.set(plugin.im.action.feature.SoundAction.FEATURE_ID, this.uid, true);
```
schmidtk: This appears to be copy/paste from the feature style action. It's not used anywhere, so you can remove it.

src/plugin/featureaction/featuresoundaction.js
```diff
@@ -0,0 +1,193 @@
...
+
+      var configSound = this.soundConfig['sound'];
```
schmidtk: Is this property read anywhere? If the intent is to play a sound when a feature matching this filter is loaded, then you don't need to save the sound to the feature.

src/plugin/featureaction/ui/featuresoundactionconfig.js
```diff
@@ -0,0 +1,106 @@
+goog.provide('plugin.im.action.feature.ui.SoundConfigCtrl');
+goog.provide('plugin.im.action.feature.ui.soundConfigDirective');
+goog.require('goog.color');
```
schmidtk: Several of these `goog.require` statements aren't needed. This UI doesn't do anything with colors, kml, vector styling, icons, etc. Remove any that aren't needed.

src/plugin/featureaction/ui/editfeatureaction.js
```diff
@@ -120,7 +163,8 @@ plugin.im.action.feature.ui.EditFeatureActionCtrl.prototype.buildStylePreview =
   var feature = new ol.Feature();
   var geometry = null;
 
-  this.styleCanvas = /** @type {HTMLCanvasElement} */ (document.createElement('canvas'));
+  this.styleCanvas = /** @type {HTMLCanvasElement} */ (document.createElement(
```
schmidtk: Did your IDE auto-format this file for 80 columns? There are a lot of indentation changes that we don't want (we allow up to 120 columns).

dwlueth:
you may want to update your commit messages to use the semantic message syntax ( https://seesparkbox.com/foundry/semantic_commit_messages)

(e.g., feat(sound): new files)

src/plugin/featureaction/featuresoundaction.js
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  os.audio.AudioManager.getInstance().play(this.soundConfig['sound']);
+
+  os.style.setFeaturesStyle(items);
+
+  var layer = os.feature.getLayer(items[0]);
+
+  if (layer) {
+    os.style.notifyStyleChange(layer, items);
```
schmidtk: You aren't changing the style. No need for this either.

src/plugin/featureaction/featuresoundaction.js
```diff
@@ -0,0 +1,193 @@
...
+plugin.im.action.feature.SoundAction.CONFIG_UI = 'featureactionsoundconfig';
+
+/**
+ * The default sound configuration.
+ * @type {!Object}
+ * @const
+ */
+plugin.im.action.feature.SoundAction.DEFAULT_CONFIG = {
+  'sound': 'Default'
+
```
dwlueth: blank line may cause a fail during build on high side.

schmidtk:
In addition to using the semantic commit message syntax, please also be a little more descriptive with the intent of the commit. Example:
```
feat(featureactions): Added a new action to play a sound when a feature is loaded.
```

dwlueth:
The code looks good. I approved prior to seeing Kevin's comments.  Obviously you'll want to address the commit messages and his comments before merging.  :)

src/plugin/featureaction/ui/editfeatureaction.js
brelsftm: I have it set for wrapping at 120

src/plugin/featureaction/featuresoundaction.js
brelsftm: So the idea of this ticket is to play the sound when creating a feature action with a filter (ex ALTITIUDE greater than 0 ) Unfortunately I dont have anyway to stream features in the unclass environment. I'm not sure exactly how that works and I wasnt sure if it needed to be saved to the feature or not. But I removed it since it doesnt have a purpose to save it to the feature from my side.

src/plugin/featureaction/featuresoundaction.js
```diff
@@ -0,0 +1,164 @@
...
+};
+
+/**
+ * @inheritDoc
+ */
+plugin.im.action.feature.SoundAction.prototype.execute = function(items) {
+  for (var i = 0; i < items.length; i++) {
```
schmidtk: I would recommend removing the for loop and calling `play` once. You don't want the sound playing multiple times if several matching features are loaded.

I would also suggest giving the user a way to set the "time between plays" parameter to `play`. Default it to something infrequent (we have used 30 seconds in similar UI's), and enforce a reasonable minimum (probably 1 or 5 seconds) to prevent the sound from playing too often.

views/plugin/featureaction/featuresoundactionconfig.html
```diff
@@ -0,0 +1,21 @@
+<div class="feature-action-sound-config">
```
schmidtk: HTML files should use 2 spaces to indent a new DOM level, 4 spaces to indent multi-line elements.

src/plugin/featureaction/featuresoundaction.js
```diff
@@ -0,0 +1,183 @@
...
+  this.label = plugin.im.action.feature.SoundAction.LABEL;
+  this.xmlType = plugin.im.action.feature.SoundAction.ID;
+  this.configUI = plugin.im.action.feature.SoundAction.CONFIG_UI;
+
+  /**
+   * Unique identifier for this action.
+   * @type {number}
+   * @protected
+   */
+  this.uid = goog.getUid(this);
```
schmidtk: This property is never used and can be removed.

src/plugin/featureaction/featuresoundaction.js
```diff
@@ -0,0 +1,183 @@
...
+   * @type {number}
+   */
+  this.delay = this.soundConfig['playDelay'] * 1000;
+
+  this.refreshTimer_ = os.debounce(this.onRefreshTimer_, this.delay, true);
+};
+goog.inherits(plugin.im.action.feature.SoundAction,
+    os.im.action.AbstractImportAction);
+
+os.implements(plugin.im.action.feature.SoundAction,
```
schmidtk: Sound actions will not be shown in the legend, so this class should not implement `ILegendRenderer`.

src/plugin/featureaction/featuresoundaction.js
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * The default sound configuration.
+ * @type {!Object}
+ * @const
+ */
+plugin.im.action.feature.SoundAction.DEFAULT_CONFIG = {
+  'sound': 'Default',
+  'playDelay': 30
+
```
schmidtk: Remove extra empty line.

src/plugin/featureaction/featuresoundaction.js
```diff
@@ -0,0 +1,183 @@
...
+  this.soundConfig = /** @type {!Object} */ (os.object.unsafeClone(
+      plugin.im.action.feature.SoundAction.DEFAULT_CONFIG));
+
+  /**
+   * User defined time between sound notifications in seconds.
+   * @type {number}
+   */
+  this.delay = this.soundConfig['playDelay'] * 1000;
+
+  this.refreshTimer_ = os.debounce(this.onRefreshTimer_, this.delay, true);
```
schmidtk: The audio manager's `play` function takes the delay as the second parameter (`opt_timeBetweenPlays`). You don't need to manage the delay yourself.

src/plugin/featureaction/ui/editfeatureaction.js
```diff
@@ -32,27 +32,24 @@ goog.require('os.ui.im.action.EventType');
 goog.require('os.ui.im.action.editFilterActionDirective');
```
schmidtk: Can you undo the indentation changes in this file? Some lines were split into multiple, when the original line doesn't exceed 120 chars.

src/plugin/featureaction/ui/editfeatureaction.js
```diff
@@ -111,6 +115,45 @@ goog.exportProperty(
     'showActionPreview',
     plugin.im.action.feature.ui.EditFeatureActionCtrl.prototype.showActionPreview);
 
+/**
+ * Create the canvas preview for a selected sound
+ * @param {plugin.im.action.feature.SoundAction} soundAction
+ */
+plugin.im.action.feature.ui.EditFeatureActionCtrl.prototype.buildSoundPreview = function(soundAction) {
+  this.soundCanvas = /** @type {HTMLCanvasElement} */ (document.createElement(
```
schmidtk: Is this writing the sound name to a canvas? This would be much more easily done with a DOM element like `span` or `div`. I would suggest using a font awesome icon to indicate that it's a sound, along with text indicating the sound name. Example:
```
<span><i class="fa fa-music"></i> (sound name)</span>
```

brelsftm: I was playing around with the icon and to me, it seemed like to cluttered the space and looked a little awkward. So I left it out , but if you think it would work better I can easily add it back in. I did add the span instead of using the canvas!

src/plugin/featureaction/ui/featuresoundactionconfig.js
```diff
@@ -0,0 +1,96 @@
...
+ * @param {!angular.Scope} $scope The Angular scope.
+ * @param {!angular.JQLite} $element The root DOM element.
+ * @extends {plugin.im.action.feature.ui.ActionConfigCtrl<plugin.im.action.feature.SoundAction>}
+ * @constructor
+ * @ngInject
+ */
+plugin.im.action.feature.ui.SoundConfigCtrl = function($scope, $element) {
+  plugin.im.action.feature.ui.SoundConfigCtrl.base(this, 'constructor', $scope,
+      $element);
+  $scope['sounds'] = os.audio.AudioManager.getInstance().getSounds();
```
schmidtk: - When creating new properties for the template that are not already on the scope, we generally prefer putting them on the controller (so, `this['sounds'] = ....`), then reference them in the template as `ctrl.sounds`. The `controllerAs` property in the template adds a reference to the controller on the scope with the provided name.
- The audio manager has a `None` option that is useful in some UI's that need to disable sound effects by selecting it, but it's less useful for this UI. If the user adds a sound action, we want a sound to be played. Remove that option from the array returned here.
- After making those two changes, add a `sound` property to the controller (as described with `sounds`) and update the controller/template to use that instead of a property on the scope. Assign it an initial value of `this['sounds'][0]` to make sure the first sound is auto-selected when the UI launches.

views/plugin/featureaction/featuresoundactionconfig.html
```diff
@@ -0,0 +1,19 @@
...
+  <select
```
schmidtk: This template is not following the Bootstrap 4 design decisions:
- Don't add classes (like `feature-action-sound-config`) that aren't used anywhere. Classes should be prefixed with `c-` if they have SCSS to style them, and `js-` if they're used in a JS selector.
- The select should have `class="form-control custom-select"`, and a label (`Sound`).
- The table is overkill and should be a `div.form-group` with a label and the spinner control. You can use `col-*` classes to align the labels/controls for the sound and delay.
- Form labels should not have a `:` after them, and should be right-justified (`text-right` class).
- Avoid using `style`. You can use `w-100` or `col` (depending on alignment with other controls) to get a full-width control.

We still need to write all of this up in the repo, but for now you can look through existing forms and Bootstrap 4 docs for additional guidance.

views/plugin/featureaction/featuresoundactionconfig.html
```diff
@@ -0,0 +1,19 @@
...
+      ng-model="sound"
+      ng-change=" ctrl.onSoundChange()"
+      style="width:100%"
+      ng-options="val for val in sounds">
+  </select>
+  <table>
+    <tr>
+      <td class="control-sound-play">Time Between Plays:</td>
```
schmidtk: Change this label to `Delay`, give it a tooltip to explain what it does, and put it on the same row as the control.

For inline label/control, you can wrap the two in a block like this:
```
<div class="d-flex flex-row form-group">
  <label class="form-control-label text-right">My Label</label>
  ... control ...
</div>

src/plugin/featureaction/ui/featuresoundactionconfig.js
brelsftm: These changes are current with the last commit

views/plugin/featureaction/featuresoundactionconfig.html
brelsftm: This change is now current with the last commit

views/plugin/featureaction/featuresoundactionconfig.html
brelsftm: These changes are current with the last commit , except for the tooltip implementation

PR #283, opened 2018-08-15
fix(file): fix non-latin1 characters in file names

### LINKED ISSUE #282: File import fails on files with non-standard characters
In the file import process, we encode the URL in Base64 using `window.btoa`: https://github.com/ngageoint/opensphere/blob/master/src/os/file/file.js#L379

This function doesn't work on characters outside the standard 16 bit encoding range. Cyrillic characters are a good test case.

wallw-bits:
Resolves #282

### DIFF
--- a/src/os/file/file.js
+++ b/src/os/file/file.js
@@ -376,7 +376,7 @@ os.file.getFileUrl = function(path) {
  * @return {string}
  */
 os.file.getLocalUrl = function(fileName) {
-  return os.file.FileScheme.LOCAL + '://' + btoa(fileName);
+  return os.file.FileScheme.LOCAL + '://' + btoa(unescape(encodeURIComponent(fileName)));
 };
 
 


wallw-bits:
This currently breaks duplicate file name detection with files added before this change. If that's a big deal then I can do something about it.

schmidtk:
I'm not worried about the duplicate detection. Small impact there.

Should this be using `decodeURI` or `decodeURIComponent` instead of `unescape`? MDN warns that it's deprecated:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape

wallw-bits:
I'm not even sure why it is "hashed" in the first place. Could we not just use `'local://' + filename` directly? Or is the original intended to avoid breaking something in IndexedDB keys? MDN says it takes `string`, which should be the full UTF-16 range supported by Javascript.

schmidtk:
I'm not sure why `btoa` was used. I wouldn't think there would be any limitations on strings used for object store keys, short of browser bugs. I'm fine with removing it.

PR #285, opened 2018-08-21
Add altitude to saved places.

### LINKED ISSUE #183: Add altitude to places

schmidtk:
Resolves #183


schmidtk:
Added an extra places bug fix. Label configs weren't always loaded correctly by feature edit. The feature config can be an array or a single object, and the latter wasn't being handled.

schmidtk:
Rebased with BS4 changes.

PR #300, opened 2018-09-13
Upgrade Cesium to 1.49.0

schmidtk:


wallw-bits:
Rectangle drawing control appears to be busted. Circle, Polygon, and adding a rectangle via coordinate entry worked fine. Other KML tests were fine too.

PR #304, opened 2018-09-18
fix(layerZoomSpinner): Fixed typing a value into a spinner's input

### LINKED ISSUE #302: Min/Max Zoom For Map Layers Won't Work If Typed In
When adjusting the zoom settings for a map layer, you can type in a value but it will not take effect. Using the arrows always works.

Either take away the type-in field or use it.

MarcelloNicoletti:
This fixes #302. You can now type a value into either spinner box to set the min/max zoom for the current layer(s). The change takes effect once the field looses focus. Values are clamped to the min and max for the field and non-numerical input is ignored.

Issues:
* "Current" buttons still set the layer zoom boundaries correctly but don't update the spinner's value anymore. This will need to be fixed but I'm not sure where I broke this binding.

### DIFF
--- a/src/os/ui/spinner.js
+++ b/src/os/ui/spinner.js
@@ -168,28 +168,37 @@ os.ui.SpinnerCtrl.prototype.onDisabledChange_ = function(opt_new, opt_old) {
  * @private
  */
 os.ui.SpinnerCtrl.prototype.onSpin_ = function(e, spinner) {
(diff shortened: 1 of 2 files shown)


schmidtk:
This looked good testing locally. Is the Current button issue expected to be fixed now? I couldn't reproduce it... button was working normally.

MarcelloNicoletti:
(Updated to current `master`)

PR #307, opened 2018-09-21
fix(zoom): Replace hard coded map max/min with JS constant for zoom spinners

MarcelloNicoletti:
* Replaced the hard coded min and max to use the constant already defined elsewhere
* Added extra `goog.require`s for extra needed requirements

### DIFF
--- a/views/plugin/basemap/basemaplayerui.html
+++ b/views/plugin/basemap/basemaplayerui.html
@@ -53,7 +53,7 @@
         <tr title="Sets the minimum zoom level at which the layer will display">
           <td class="text-nowrap">Min Zoom:</td>
           <td class="w-100">
-            <spinner min="2" max="maxZoom" step="1" value="minZoom" name="minZoom"></spinner>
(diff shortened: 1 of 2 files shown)


views/plugin/basemap/basemaplayerui.html
```diff
@@ -53,7 +53,7 @@
         <tr title="Sets the minimum zoom level at which the layer will display">
           <td class="text-nowrap">Min Zoom:</td>
           <td class="w-100">
-            <spinner min="2" max="maxZoom" step="1" value="minZoom" name="minZoom"></spinner>
+            <spinner min="::basemap.mapMinZoom" max="maxZoom" step="1" value="minZoom" name="minZoom"></spinner>
```
MarcelloNicoletti: In angular 1.3+ `::` creates a one time binding. the `$watch` is automatically deregistered after the first resolved value. Since map max/min doesn't change during run time there is no need to `$watch` for changes

PR #311, opened 2018-10-01
feat(terrain)Add selector for layer altitude mode

### LINKED ISSUE #303: Enable clamp-to-ground in Cesium
Our lack of support for clamp-to-ground causes vector features without explicitly-defined altitude to "appear underneath" (positioned underneath, but drawn on top of due to Cesium's render order) the terrain.

We previously experimented (almost two years ago now) with using Cesium's `CLAMP_TO_GROUND` altitude mode support. At the time it was too sluggish (especially when combined with raster terrain). However, Cesium has made improvements over time and we are now using their STK terrain mesh. Re-enable altitude mode support.

Things to keep in mind:
* This is all handled by the `heightReference` items in `plugin.cesium.sync.FeatureConverter`. There's some commented-out stuff in `getHeightReference()` that might be useful.
* I think we explicitly add an altitude of zero to coordinates lacking an altitude. Verify if this is the case. If it is, we'll need to distinguish from a parsed altitude of `0` (leave it at 0) and an added altitude of 0 (clamp to ground). This can likely be done via an `altitudeMode` field on the feature.
* The user needs an option on the feature layer to disable altitude, which should clamp all features to ground. The code exists for this (`os.source.Vector.prototype.setAltitudeEnabled` and also a handler in the vector layer UI) but currently there is no UI item or menu item hooked up to it. Bonus points if they get to decide the altitude mode for the layer (Altitude from mean sea level, Altitude relative to ground, Clamp to ground).

MarcelloNicoletti:
* Add support for choosing altitude mode for layer through ui options
* Add backend support for getting supported modes and handling changes
* Currently only works on point type features
* Closes #303


MarcelloNicoletti:
Some thoughts:
* The different functions I've added to IWebGLRenderer are designed to work between different backends.
* I haven't done the work in Cesium to enable draping polygon features on terrain when using the "Clamp to Ground" mode.
* When switching to and from 3D mode the "Altitude Mode" option appears in the layer options card. There is no animation though. The selector instantly (dis)appears which feels a little jarring since the card otherwise fluidly slides up and down when activated

MarcelloNicoletti:
Note: Commit title will have to change before pulling `altidude -> altitude`

schmidtk:
Items to address before this can be merged:
- Add `altitudeMode` to ngageoint/opensphere-state-schema. That change will need to be merged/released for tests to pass in this PR.
- Trigger a descriptor save when `altitudeMode` changes so the value is persisted on application refresh. `os.data.LayerSyncDescriptor` will save the descriptor when the layer fires a property change event. You can either move management of `altitudeMode` to the layer and fire the event directly in the setter, or fire a property change event on the source and have `os.layer.Vector#onSourceChange` fire an event from the layer.

src/os/webgl/webgl.js
```diff
@@ -48,3 +48,29 @@ os.webgl.hasPerformanceCaveat = function() {
   }
   return null;
 };
+
+/**
+ * An altitude mode id and name
+ * @typedef {Object} AltitudeModeValue
```
MarcelloNicoletti: Should I redo this line to be `@typedef {Object} os.webgl.AltitudeModeValue` or is it already scoped here? It's use below as the enum type is from line 58 below

schmidtk: Currently OpenSphere uses [this format](https://github.com/ngageoint/opensphere/blob/master/src/os/color.js#L10-L17) for `@typedef`, so please use that for consistency.

src/os/layer/vector.js
```diff
@@ -311,6 +311,12 @@ os.layer.Vector.prototype.onSourceChange = function(event) {
     case os.source.PropertyChange.COLUMN_ADDED:
       this.dispatchEvent(new os.events.PropertyChangeEvent(p));
       break;
+    case os.source.PropertyChange.ALTITUDE:
+      // forward as a layer event
+      e = new os.events.PropertyChangeEvent(os.layer.PropertyChange.ALTITUDE, event.getNewValue(),
+          event.getOldValue());
+      this.dispatchEvent(e);
+      break;
```
MarcelloNicoletti: I'm following the example set by `TIME_ENABLED`, `STYLE`, etc

src/os/ui/layer/vectorlayerui.js
```diff
@@ -977,3 +998,34 @@ os.ui.layer.VectorLayerUICtrl.prototype.onRotationColumnChange = function(event,
...
+
+
+/**
+ * Handle map property changes.
+ * @param {os.events.PropertyChangeEvent} event The change event.
+ * @private
+ */
+os.ui.layer.VectorLayerUICtrl.prototype.onMapView3DChange_ = function(event) {
+  if (event.getProperty() == os.MapChange.VIEW3D) {
+    this['showAltitudeModes'] = event.getNewValue() && this['altitudeModes'].length > 0;
```
MarcelloNicoletti: Should i change this to > 1. As it stands the drop down menu can appear with only one option

src/plugin/cesium/sync/featureconverter.js
```diff
@@ -1035,13 +1037,20 @@ plugin.cesium.sync.FeatureConverter.prototype.olPolygonGeometryToCesiumPolyline
...
+  switch (altitudeMode) {
+    case os.webgl.AltitudeMode.ABSOLUTE:
+      this.heightReference_ = Cesium.HeightReference.NONE;
+      break;
+    case os.webgl.AltitudeMode.RELATIVE_TO_GROUND:
+      this.heightReference_ = Cesium.HeightReference.RELATIVE_TO_GROUND;
+      break;
+    case os.webgl.AltitudeMode.CLAMP_TO_GROUND:
+    default:
+      this.heightReference_ = Cesium.HeightReference.CLAMP_TO_GROUND;
```
schmidtk: The layer default should be `ABSOLUTE`, since `CLAMP_TO_GROUND` will incorrectly display features that are actually above/below the ground.

Later we can address having our parsing stack set individual features to `CLAMP_TO_GROUND` if they don't have an altitude.

views/layer/vector.html
```diff
@@ -71,6 +71,16 @@
           </select>
         </div>
       </div>
+      <div class="form-row align-items-center pb-1" ng-if="vector.showAltitudeModes">
+        <div class="col-auto" title="Choose the layer's altitude mode.">
```
schmidtk: Move the `title` up a level so it displays to the user when hovering the control as well as the label.

I would also suggest making the title a bit more descriptive. Something like:
`Choose how the layer interprets altitude in 3D mode.`

src/os/ui/layer/vectorlayerui.js
schmidtk: Selects with a single option aren't great, but hiding it may be alarming to a user that's used to the control in Cesium (if they switch to Strata and Strata doesn't have full support yet). I think this is fine as is, and I would also expect Strata to have this support before we release it.

views/layer/vector.html
MarcelloNicoletti: Since the control doesn't appear in 2D mode should I still add "in 3D mode"?

PR #315, opened 2018-10-08
KML BalloonStyle tag support

brelsftm:
Added Support for parsing the KML BalloonStyle tag.


src/plugin/file/kml/kmlparser.js
```diff
@@ -841,6 +849,55 @@ plugin.file.kml.KMLParser.prototype.createTreeNode_ = function(el, opt_parent) {
...
+    }
+  }
+};
+
+
+/**
+ * Read the KML balloon style.
+ * @param {ol.Feature} feature The feature
+ * @private
+ */
```
welchyd: Add `@suppress {accessControls}` to fix one of the build failures

src/plugin/file/kml/kml.js
```diff
@@ -217,6 +217,42 @@ plugin.file.kml.readStyle = function(node, objectStack) {
 };
 
 
+/**
+ * Parse text field from BalloonStyle tag.
+ * @param {Node} node Node.
+ * @return {string|*}
+ * @private
+ */
+plugin.file.kml.readBalloonText_ = function(node) {
```
schmidtk: This function is identical to `ol.format.XSD.readString`. Please use that instead.

src/plugin/file/kml/kml.js
```diff
@@ -217,6 +217,42 @@ plugin.file.kml.readStyle = function(node, objectStack) {
...
+};
+
+
+/**
+ * Parse display mode from the BalloonStyle tag.
+ * @param {Node} node Node.
+ * @return {string|*}
+ * @private
+ */
+plugin.file.kml.readDisplayMode_ = function(node) {
```
schmidtk: Identical to the previous function. Again, please use `ol.format.XSD.readString`.

src/plugin/file/kml/kmlparser.js
brelsftm: Ah okay I was wondering what the best way to get past that build error was, thanks for that!!

src/plugin/file/kml/kmlparser.js
```diff
@@ -841,6 +849,55 @@ plugin.file.kml.KMLParser.prototype.createTreeNode_ = function(el, opt_parent) {
...
+    text = text.replace(regex, function(match) {
+      var key = match.slice(2, -1);
+      if (key in feature.values_) {
+        return feature.get(key);
+      } else {
+        return '';
+      }
+    });
+    var description = '<div style="background:rgba(' + bgColor + ');color:rgba(' + textColor + ')">' + text + '</div>';
+    feature.set('description', description);
```
schmidtk: Do not store this in the `description` property on the feature. That field is already used for the KML `<description>` tag on the Placemark, and we don't want to wipe that out.

src/plugin/file/kml/kmlparser.js
```diff
@@ -841,6 +849,57 @@ plugin.file.kml.KMLParser.prototype.createTreeNode_ = function(el, opt_parent) {
...
+  var text = style.text;
+  var bgColor = style.bgColor || '255, 255, 255, 1';
+  var textColor = style.textColor || '0, 0, 0, 1';
+
+  if (style && text) {
+    var pattern = /[$]\[(.*?)\]/g;
+    var regex = new RegExp(pattern);
+
+    text = text.replace(regex, function(match) {
+      var key = match.slice(2, -1);
```
schmidtk: The first capture group will be the [second param to the replacer function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_function_as_a_parameter). No need to do this slice.

src/plugin/file/kml/kmlparser.js
```diff
@@ -125,6 +125,13 @@ plugin.file.kml.KMLParser = function(options) {
    */
   this.styleMap_ = {};
 
+  /**
+   * The KML styles not supported by OpenLayers map.
+   * @type {!Object<string, !Object<string>>}
```
schmidtk: This type is invalid (inner `Object` has a key type but no value type).

If the value is an object with arbitrary keys, the type here should be `!Object<string, !Object>`. `<string, *>` is implied when a key/value signature isn't specified.

src/plugin/file/kml/kmlparser.js
```diff
@@ -1590,17 +1654,22 @@ plugin.file.kml.KMLParser.prototype.mapStyleToConfig_ = function(style) {
  * Finds the first instance of a style id on the style stack
  * @param {string} id The style id
  * @param {boolean=} opt_highlight Whether to check the highlight style map
+ * @param {boolean=} opt_otherStyle Whether to check externally parsed styles.
+
  * @return {Array<Object>} The style configs, or null if not found
  * @private
  */
-plugin.file.kml.KMLParser.prototype.findStyle_ = function(id, opt_highlight) {
+plugin.file.kml.KMLParser.prototype.findStyle_ = function(id, opt_highlight, opt_otherStyle) {
```
schmidtk: This function is a little awkward, given the parameter train to adjust which map is used. Instead, we should extract the bit that determines the id into a function and use the result to retrieve values from the appropriate map directly.

src/plugin/file/kml/kmlparser.js
wallw-bits: I told him to do that. What's the point of keeping it around if we only want to show the other?

wallw-bits:
Please ignore the suggestion to "keep it down to one commit". @welchyd, we prefer a series of small, logical commits with good messages. Long-running PRs with multiple iterations may certainly squash/rebase as necessary, but there is still no requirement for a single commit.

src/plugin/file/kml/kmlparser.js
schmidtk: While they may contain the same content, that's up to the KML and it's possible they'll each have unique content. We don't want to discard the description in favor of the balloon text so we need a way to show both.

Google Earth shows the balloon text on click and the description in the Get Info context menu option. I'll look into getting a balloon implementation we can use to be closer to that behavior.

welchyd:
Just trying to be helpful with one commit for this change - if not for the sake of @wallw-bits, it could make rebasing more straightforward with less steps ... and it looks like a rebase is needed.

schmidtk:
I merged this into #331 with appropriate updates to display annotations for KML balloons.

The changes were merged into [this commit](https://github.com/ngageoint/opensphere/pull/331/commits/87b0b242639a6303dff976b4f6cffde66d8091d5), and integrated with annotations in [this commit](https://github.com/ngageoint/opensphere/pull/331/commits/3439374755ffc8fb1d8208f04eb5ae3924f02aa7). If that's acceptable for this work, we can close this PR.

PR #320, opened 2018-10-11
Refactor goog.array calls

schmidtk:
This replaces:
- `goog.array.clear`
- `goog.array.contains`
- `goog.array.find`
- `goog.array.findIndex`
- `goog.array.forEach`
- `goog.array.remove`

### DIFF
--- a/conformance_config.textproto
+++ b/conformance_config.textproto
@@ -190,3 +190,15 @@ requirement: {
 #  value: 'goog.isString'
 #  error_message: 'goog.isString(value) is not permitted. Use typeof value === \'string\' instead.'
 #}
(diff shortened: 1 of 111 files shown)


schmidtk:
I added `os.array.clear` and `os.array.forEach` implementations that provide a 1:1 translation with the `goog` equivalents. They're also safe when passed `null` or `undefined`, so I deprecated `os.array.forEachSafe`. The implementations do use `IArrayLike` from the Closure Compiler externs in their JSDoc to maintain compatibility with both arrays and array-like objects, so that will need to be addressed separately if we ever stop using the compiler.

PR #323, opened 2018-10-16
Cascade windows with the same id

schmidtk:
This causes saved windows with the same `id` to be cascaded. The cascade behavior was designed to mimic macOS:

- Find top-most like window and add x/y offset to that position.
- If the new window hits the right edge of the screen, change x position of new window to 0.
- If it hits the bottom edge of the screen, change y position of the new window to 0 (under the top bar).

In OpenSphere, this results in windows being positioned on top of the top nav because the window container contains the navs. This feels a little odd, but the best way to resolve that would be changing the window container to the map container. That seemed like a potentially unwanted change, and perhaps not fit for this PR.

### DIFF
--- a/src/os/ui/query/ui/modifyarea.js
+++ b/src/os/ui/query/ui/modifyarea.js
@@ -394,8 +394,7 @@ os.ui.query.ui.launchModifyArea = function(config) {
   var windowId = 'modifyArea';
   if (os.ui.window.exists(windowId)) {
     // update the existing window
-    var win = document.querySelector('.window#' + windowId);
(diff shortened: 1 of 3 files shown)


schmidtk:
I updated this to cascade all windows based on `id` instead of just saved windows. This allows any like window to be opened multiple times and cascaded, if necessary.

I changed the Create/Edit place windows to all use the same `id`, so they'll all be cascaded against one another. Previously edit windows were given a unique `id` based on the feature, and would be centered and potentially overlap other similar windows.

Also found a bug where Modify Area wasn't updating the existing window if present, because the selector wasn't updated with Bootstrap 4.

PR #326, opened 2018-10-24
Export CSV with no geometry data.

brelsftm:
Fixed bug where you cannot export csv data that has no geometry data.

### DIFF
--- a/test/plugin/file/csv/csvexporter.test.js
+++ b/test/plugin/file/csv/csvexporter.test.js
@@ -27,11 +27,6 @@ describe('plugin.file.csv.CSVExporter', function() {
     expect(result).toBeNull();
   });
 
-  it('does not process features lacking a geometry', function() {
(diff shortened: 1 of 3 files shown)


wallw-bits:
OK. Those specific test failures are directly related and need to be addressed.

test/plugin/file/csv/csvexporter.test.js
```diff
@@ -27,11 +27,6 @@ describe('plugin.file.csv.CSVExporter', function() {
     expect(result).toBeNull();
   });
 
-  it('does not process features lacking a geometry', function() {
-    var result = ex.processItem(new ol.Feature());
-    expect(result).toBeNull();
-  });
-
```
wallw-bits: This is not a good test fix. The better fix is to change the test from "It should not process features lacking a geometry" to "It should process features lacking a geometry" so that your new code path is tested.

PR #331, opened 2018-10-31
Add annotations to Saved Places

schmidtk:
Latest build can be tested here:
https://schmidtk.github.io/annotations/

This adds map annotations to Saved Places. Currently supported:
- Create annotations and persist across sessions.
- Sync between 2D/3D.
- Annotations are included in screen capture (screenshots and recordings).
- KML BalloonStyle rendered as an annotation.
- Annotations exported to KML BalloonStyle for display in Google Earth.

General OpenSphere changes:
- `plugin.cesium.CesiumRenderer#getCoordinateFromPixel` now picks the globe instead of the ellipsoid. This provides a more accurate coordinate (with altitude) when terrain is enabled.
- `os.webgl.IWebGLRenderer#getPixelFromCoordinate` now takes an optional param to test if the coordinate is in view. If not, `null` is returned. This has been implemented in the Cesium renderer.
- `plugin.capture.MapOverlayRenderer` now renders `ol.Overlay` instances. The original class has been renamed to `plugin.capture.MapOverviewRenderer`, to align with OpenLayers terms.

TODO:
- [x] Merge #315 and display annotations for KML balloons.
- [x] Fix 3D sync when annotations are created outside the current view.
- [x] Export annotations to KML for display in Google Earth.


src/os/ui/menu/mapmenu.js
```diff
@@ -29,50 +39,51 @@ os.ui.menu.map.setup = function() {
     children: [{
       label: 'Map',
       type: os.ui.menu.MenuItemType.GROUP,
+      sort: os.ui.menu.map.GroupSort.GROUPS++,
```
wallw-bits: This works nicely here but forces plugins to use decimal values to insert something between them.

schmidtk: Fair point. I suppose it also makes the order for plugin menu items dependent on plugin load order... also bad. I'll switch back to the hard coded values and use fixed values for the group sort as well.

wallw-bits:
Just a few things from testing (I was looking at https://schmidtk.github.io/annotations/):

1. The pointer does not work well from the sides.
screen shot 2018-11-15 at 08 54 11
2. I'm not a fan of how the edit button floats outside the "window". It is fine if it disappears when the mouse is not over it, but I would rather have it show on the right side of the title bar so that it is easier to see (this goes for the Legend too, but that's separate work).
3. Using enough text to force a vertical scroll for a given window size caused this:
screen shot 2018-11-15 at 09 02 00
4. These things eat my mousewheel while zooming around. It's the inverse of a zoomable map on a scrollable page. Can we have them ignore the events unless they are focused?

schmidtk:
#1, 3, 5 have been fixed/deployed.

For #2, moving the edit button inside the annotation causes further problems. It can't go in the header because that can be disabled, and it's difficult to display well inside the body due to text truncate/wrapping constraints with the embedded `simplemde` directive that renders the description text. Skipping this one, at least for now.

#4 is going to be problematic as well. Passing mouse events through to the canvas requires setting `pointer-events: none` on the overlay, which also prevents attaching a click/mousedown handler to remove that and "focus" the overlay to allow scrolling its contents.

PR #332, opened 2018-11-01
AlertPopup now selectable and urls have hyperlinks

### LINKED ISSUE #329: Alerts Not Selectable in Opensphere
Users would like the capability to highlight text in the opensphere alerts (i.e., toaster popup messages) so they can copy it.

To force an alert in opensphere:

1. Open opensphere in dev mode and open the dev tools
2.  In the console type:  os.alert.AlertManager.senAlert('some text here')

### LINKED ISSUE #330: Change URL text in Alerts Popups to links
In the alert popup directive, make urls that are within the alert text links.

There should already be a regex in the code that determines if text has URL.

lilysikes:
resolves #329 
resolves #330

### DIFF
--- a/scss/os/_alertpopup.scss
+++ b/scss/os/_alertpopup.scss
@@ -1,5 +1,4 @@
 // We want alerts to popup above windows and modals
 .c-alert-popup {
-  user-select: text;
   z-index: $zindex-modal + 1;
 }
--- a/scss/os/_alertpopup.scss
+++ b/scss/os/_alertpopup.scss
@@ -1,4 +1,5 @@
 // We want alerts to popup above windows and modals
 .c-alert-popup {
+  user-select: text;
   z-index: $zindex-modal + 1;
 }
--- a/src/os/ui/alert/alertpopup.js
+++ b/src/os/ui/alert/alertpopup.js
@@ -241,7 +241,7 @@ os.ui.alert.AlertPopupCtrl.prototype.displayAlert_ = function(event) {
 
     // Check if message has a url and if it does adds a hyperlink
     var match = message.match(os.url.URL_REGEXP_LINKY);
-    if (match) {
+    if (match && match.index + match[0].length <= 500) {
       var html = [];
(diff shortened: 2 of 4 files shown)


src/os/ui/alert/alertpopup.js
```diff
@@ -238,6 +238,24 @@ os.ui.alert.AlertPopupCtrl.prototype.displayAlert_ = function(event) {
...
+      html.push('">');
+      html.push(os.ui.escapeHtml(message.substring(i, i + match[0].length)));
+      html.push('</a>');
+      message = message.substring(i + match[0].length);
+      html.push(message);
+      message = html.join('');
+    }
     var popup = {
       'id': id,
       'msg': goog.string.truncate(message, 500),
```
mattpaulsen: So, it looks like the message gets truncated once it is over 500 characters.  I wonder if it will cause issues if it gets truncated in the middle of your link?

If it is an issue, maybe add a check where if it will get truncated you don't add the link?

mattpaulsen:
I made a comment in the code wondering if we need to handle the fact that the message is truncated how that will interact with now adding a link

scss/os/_alertpopup.scss
```diff
@@ -1,4 +1,5 @@
 // We want alerts to popup above windows and modals
 .c-alert-popup {
+  user-select: text;
```
wallw-bits: This prompted me to ask @gregroyal what happened to the `.selectable` and `.unselectable` classes in the BS4 update. Some of our views are definitely referencing `.unselectable` even though it does not appear to exist in the compiled CSS.

Rather than just tacking it on here, it would make sense to have some very generic `.selectable` and `.unselectable` utility classes in `scss/os/_utils.scss`, and then use the `selectable` class in the alert popup view template.

wallw-bits:
@dwlueth The commit message is fine as far as formatting. Are you complaining about the tense requested by commitizen? Because I do not think we have ever enforced that.

PR #333, opened 2018-11-08
Add WMTS server support

wallw-bits:
Adds support for WMTS servers, such as [NASA GIBS](https://earthdata.nasa.gov/about/science-system-description/eosdis-components/global-imagery-browse-services-gibs). Time dimensions are supported.

This also fixes a couple of issues with projection switching. When choosing the best projection for a layer, it must choose one enumerated for support by the server. This means that the layer can choose something like `urn:ogc:def:crs:OGC:1.3:CRS84`. However, the projection switch should change the overall map projection to `EPSG:4326`, since things like default map layers are done by string codes in settings.

- [x]  WMTS servers from config
- [x]  WMTS servers from users
- [x]  Auto detect date/time format from dimension default if not set
-----------
### Testing
Config:
```
      "wmts-test": {
        "type": "wmts",
        "label": "NASA GIBS WMTS",
        "url": "https://gibs.earthdata.nasa.gov/wmts/epsg4326/best/wmts.cgi"
      }
```
For user added URLs, [use this](https://gibs.earthdata.nasa.gov/wmts/epsg4326/best/wmts.cgi?request=GetCapabilities).

GeoServer's WMTS implementation would also be a good test since it supports multiple projections in the same capabilities document.

### DIFF
--- a/src/plugin/wmts/wmtsimport.js
+++ b/src/plugin/wmts/wmtsimport.js
@@ -1,0 +1,66 @@
+goog.provide('plugin.wmts.ImportCtrl');
+goog.provide('plugin.wmts.importDirective');
+goog.require('os.defines');
+goog.require('os.ui.Module');
+goog.require('os.ui.SingleUrlProviderImportCtrl');
+goog.require('os.ui.window');
+goog.require('plugin.wmts.Server');
+
+
+/**
+ * The geoserver import directive
+ * @return {angular.Directive}
+ */
+plugin.wmts.importDirective = function() {
+  return {
+    restrict: 'E',
+    replace: true,
+    templateUrl: os.ROOT + 'views/forms/singleurl.html',
+    controller: plugin.wmts.ImportCtrl,
+    controllerAs: 'ctrl'
+  };
+};
+
+
+/**
+ * Add the directive to the module
+ */
+os.ui.Module.directive('wmtsserver', [plugin.wmts.importDirective]);
+
+
+
+/**
+ * Controller for the geoserver import dialog
+ * @param {!angular.Scope} $scope
+ * @param {!angular.JQLite} $element
+ * @extends {os.ui.SingleUrlProviderImportCtrl}
+ * @constructor
+ * @ngInject
+ */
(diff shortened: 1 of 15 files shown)


wallw-bits:
I'm waiting on some feedback from some third-parties on whether item 4 there will be necessary. I would rather avoid it since it is messy.

wallw-bits:
Item 4 was not necessary and I've removed it. Ready for review.

src/plugin/wmts/wmtsimport.js
```diff
@@ -0,0 +1,66 @@
...
+  dp.configure(this.scope['config']);
+  return dp;
+};
+
+
+/**
+ * @inheritDoc
+ */
+plugin.wmts.ImportCtrl.prototype.getUrl = function() {
+  return this.dp ? /** @type {plugin.ogc.GeoServer} */ (this.dp).getUrl() : '';
```
schmidtk:
```suggestion
  return this.dp ? /** @type {plugin.wmts.Server} */ (this.dp).getUrl() : '';
```

src/plugin/wmts/wmtsimport.js
```diff
@@ -0,0 +1,66 @@
...
+goog.provide('plugin.wmts.importDirective');
+goog.require('os.defines');
+goog.require('os.ui.Module');
+goog.require('os.ui.SingleUrlProviderImportCtrl');
+goog.require('os.ui.window');
+goog.require('plugin.wmts.Server');
+
+
+/**
+ * The geoserver import directive
```
schmidtk:
```suggestion
 * The WMTS server import directive
```

src/plugin/wmts/wmtsimport.js
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Add the directive to the module
+ */
+os.ui.Module.directive('wmtsserver', [plugin.wmts.importDirective]);
+
+
+
+/**
+ * Controller for the geoserver import dialog
```
schmidtk:
```suggestion
 * Controller for the WMTS server import dialog
```

PR #335, opened 2018-11-13
feat(tracks): WIP: include metadata with tracks

welchyd:


welchyd:
I've decided to take a different approach to this.
The goal here was to expose the metadata used for making a track

PR #338, opened 2018-11-20
fix(alertpopup): Removed tooltip from alertpopup and added url hyperl…

lilysikes:
…inks to alerts window.


wallw-bits:
Can you please have a look at the failing tests?

PR #357, opened 2018-12-13
feat(ecql): Add an option to remove altitude from area queries

wallw-bits:
Some servers that we have encountered only support the WKT `POLYGON()` type and error when given `POLYGON Z()`. Therefore, adding an option to remove the altitude portion of the coordinates allows the queries to be made as expected.

### DIFF
--- a/src/os/filter/impl/ecql/areaformatter.js
+++ b/src/os/filter/impl/ecql/areaformatter.js
@@ -36,6 +36,28 @@ os.filter.impl.ecql.AreaFormatter = function(opt_column) {
    * @protected
    */
   this.group = 'OR';
+
+  /**
+   * @type {boolean}
+   * @protected
+   */
+  this.supportsAltitude = true;
+};
+
+
+/**
+ * @return {boolean}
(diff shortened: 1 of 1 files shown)


src/os/filter/impl/ecql/areaformatter.js
```diff
@@ -49,6 +83,12 @@ os.filter.impl.ecql.AreaFormatter.prototype.format = function(feature) {
   if (geom) {
     geom = geom.clone().toLonLat();
     os.geo.normalizeGeometryCoordinates(geom);
+
+    if (!this.supportsAltitude) {
+      var poly = /** @type {ol.geom.MultiPolygon|ol.geom.Polygon} */ (geom);
+      poly.setCoordinates(os.filter.impl.ecql.AreaFormatter.stripAltitude(poly.getCoordinates()));
```
schmidtk: `os.filter.impl.ecql.AreaFormatter.stripAltitude` doesn't return anything, so this sets the coordinates to undefined.

PR #359, opened 2018-12-19
fix(mapContainer): Checks to make sure layer exists before attempting…

justin-bits:
… to remove a feature

Fixes an issue where cls would fail intermittently due to the order of events


justin-bits:
This was found while testing with Cypress on calling cls(), but probably also occurs when users perform a settings reset.  If this needs an issue to be created and linked, let me know.

cls error

PR #363, opened 2018-12-20
refactor(TimelineLock): Lock button was added to the timeline panel a…

### LINKED ISSUE #352: Animation lock button issues
With the addition of the animation lock button at the top of the timeline, users can inadvertently move the flags and the button can become hidden. 

-  Move the lock button to a more prominent position along with the other buttons below the timeline.
-  When my view window (blue) is smaller than my load window (yellow), it loops back around to the left edge of the loaded view. It should loop back to start at the lock location.

lilysikes:
…nd the view window loops back to the locked position.

resolves #352

### DIFF
--- a/src/os/ui/timelinepanel.js
+++ b/src/os/ui/timelinepanel.js
@@ -9,6 +9,7 @@ goog.require('os.ui.animationSettingsDirective');
 goog.require('os.ui.hist.HistogramEventType');
 goog.require('os.ui.timeSettingsDirective');
 goog.require('os.ui.timeline.AbstractTimelineCtrl');
+goog.require('os.ui.timeline.Brush');
 
 
 /**
@@ -51,6 +52,18 @@ os.ui.resizeMap = 'resizeMap';
  */
 os.ui.TimelinePanelCtrl = function($scope, $element, $timeout) {
   os.ui.TimelinePanelCtrl.base(this, 'constructor', $scope, $element, $timeout);
+
+  /**
+   * @type {boolean}
+   */
+  this['locked'] = os.time.TimelineController.getInstance().getLock();
+
+  /**
+   * @type {?angular.Scope}
+   */
+  this['scope'] = $scope;
+
+  this.tlc.listen(os.time.TimelineEventType.LOCK_TOGGLE, goog.partial(this.setLock, true), false, this);
 
   /**
    * @type {?os.data.histo.TimelineHistManager}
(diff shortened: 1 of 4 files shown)


schmidtk:
The animation behavior isn't what I would expect. The Next/Previous/Last buttons are still changing the left edge of the view (blue) window when it's locked. The only time we should change the locked position is if it's outside the animation range when the user presses Play/Next/Prev/First/Last. If that happens, move it to the start of the animation range.

Animation is also broken when explicit animation ranges are drawn. You can draw these with shift click+drag, then choose Animate from the menu. Multiple animation ranges can be drawn, and when the view hits the end of one it will jump to the first frame of the next range. Currently, when the view is locked and one or more animation ranges are present, the view gets reset in an odd way as shown here:
timelinelockedanimationrange

As before, the left/locked edge of the view should not change if it's within the animation range. The right edge should jump correctly between multiple ranges, and when it hits the end of all animation ranges it should reset to the locked position.

schmidtk:
One minor request to fully resolve #352 - can you remove the lock button displayed in the top/left of the view window? It should be replaced by the one in the timeline button bar. Thanks!

src/os/ui/timelinepanel.js
```diff
@@ -52,6 +53,18 @@ os.ui.resizeMap = 'resizeMap';
...
+   * @type {boolean}
+   */
+  this['locked'] = os.time.TimelineController.getInstance().getLock();
+
+  /**
+   * @type {?angular.Scope}
+   */
+  this['scope'] = $scope;
+
+  this.tlc.listen(os.time.TimelineEventType.LOCK_TOGGLE, goog.partial(this.setLock, true), false, this);
```
schmidtk: The `goog.partial` call can be removed from this listener. That injects a param into the callback, but you're not using it. It will also prevent you from cleaning up the listener unless you assign the returned function to a param, so you can pass that instance of the function to `unlisten`. This line should be:
`this.tlc.listen(os.time.TimelineEventType.LOCK_TOGGLE, this.setLock, false, this);`

That said, you also need a corresponding `unlisten` in the `destroy` function. Please add that.

src/os/ui/timelinepanel.js
```diff
@@ -233,3 +246,40 @@ os.ui.TimelinePanelCtrl.prototype.adjust = function() {
...
+ */
+os.ui.TimelinePanelCtrl.prototype.setLock = function() {
+  this['locked'] = this.tlc.getLock();
+  os.ui.apply(this.scope);
+};
+
+
+/**
+ * Panel lock button click.
+ * @param {boolean=} opt_getlock
```
schmidtk: `opt_getlock` is not used and can be removed from this function.

src/os/ui/timelinepanel.js
```diff
@@ -233,3 +246,40 @@ os.ui.TimelinePanelCtrl.prototype.adjust = function() {
...
+ */
+os.ui.TimelinePanelCtrl.prototype.onBrushLockButtonUp = function(opt_getlock) {
```
schmidtk: This function is only called by `lock`, so the body can be moved there instead of having two functions.

src/os/ui/timelinepanel.js
```diff
@@ -233,3 +239,21 @@ os.ui.TimelinePanelCtrl.prototype.adjust = function() {
   os.MapContainer.getInstance().updateSize();
   os.dispatcher.dispatchEvent(os.MapEvent.GL_REPAINT);
 };
+
+
+/**
+ * Panel lock button click.
+ */
+os.ui.TimelinePanelCtrl.prototype.lock = function() {
```
jsalankey: This function is called at `timelinepanel.html#26`, so it needs to have the `@export` tag added to the comment. This prevents the compiler from renaming the function and without it, it might not work in compiled mode.

PR #367, opened 2018-12-21
fix(wmts): prune unsupported projections

wallw-bits:
OpenLayers throws an error if we pass it a matrix set whose projection is not defined in the
application. Therefore, prune the matrix sets down to just the projections that are supported.

Also fixed an issue where we were just using the first `Format` listed rather than preferring `image/png` when available.

### DIFF
--- a/src/plugin/wmts/wmtsserver.js
+++ b/src/plugin/wmts/wmtsserver.js
@@ -207,7 +207,7 @@ plugin.wmts.Server.prototype.configure = function(config) {
   this.setDateFormat(/** @type {string} */ (config['dateFormat']));
   this.setTimeFormat(/** @type {string} */ (config['timeFormat']));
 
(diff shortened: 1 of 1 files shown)


src/plugin/wmts/wmtsserver.js
```diff
@@ -207,7 +207,7 @@ plugin.wmts.Server.prototype.configure = function(config) {
   this.setDateFormat(/** @type {string} */ (config['dateFormat']));
   this.setTimeFormat(/** @type {string} */ (config['timeFormat']));
 
-  this.setParams('params' in config ? new goog.Uri.QueryData(/** @type {string} */ (config['wmsParams'])) : null);
+  this.setParams('params' in config ? new goog.Uri.QueryData(/** @type {string} */ (config['params'])) : null);
```
wallw-bits: This fixes some copy pasta that wasn't allowing custom params.

PR #368, opened 2018-12-27
feat(featureActions): implements support for feature actions based on…

jsalankey:
… time filters

- This is using the TIME display field and the `recordTime` field on the feature that we use for temporal indexing, so it should follow whatever is chosen on import to be used as that field.
- I added a slightly ugly global window variable to reduce the number of calls to `Date.now()` when executing the filter functions. It was getting called unnecessarily hundreds of thousands of times with large amounts of features vs. once per run now.
- I exported the `getStart` and `getEnd` methods on `TimeInstant` and `TimeRange` so that I can access them from the functions created by feature actions. If we have a reason not to do this, let me know.
- This also adds a global timer to `FeatureActionManager` for executing any time-based feature actions every 15 seconds. This is to ensure that they update regularly on old features as they age.


src/plugin/featureaction/featureaction.js
```diff
@@ -191,12 +191,5 @@ plugin.im.action.feature.editEntry = function(entryType, opt_entry) {
  * @return {boolean} Whether to refresh the layer.
  */
 plugin.im.action.feature.shouldRefresh = function(entryType) {
```
schmidtk: If this function is no longer needed, we should remove it entirely.

src/os/ui/filter/op/time/betweentimeop.js
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * @inheritDoc
+ */
+os.ui.filter.op.time.Between.prototype.getEvalExpression = function(varName, literal) {
+  var range = this.getRangeFromLiteral(literal);
+  if (range && range.length == 2 && range[0] !== range[1]) {
+    var r0 = range[0];
+    var r1 = range[1];
+    return '(window["currentTimestamp"]-' + varName + '.getStart()<=' + r1 +
```
schmidtk: This needs to handle if `varName` is not defined (no time).

src/os/ui/filter/op/time/newerthanop.js
```diff
@@ -0,0 +1,46 @@
...
+
+
+/**
+ * @inheritDoc
+ */
+os.ui.filter.op.time.NewerThan.prototype.getEvalExpression = function(varName, literal) {
+  if (!goog.string.isEmptyOrWhitespace(goog.string.makeSafe(literal))) {
+    // the value stored in the variable for a time-based operator is an {@link os.time.ITime}
+    // to execute the filter, we need to extract the raw time value and compare it to now
+    return 'window["currentTimestamp"]-' + varName + '.getStart()<' + os.ui.filter.string.quoteString(literal);
```
schmidtk: This needs to handle if `varName` is not defined (no time).

src/os/ui/filter/op/time/olderthanop.js
```diff
@@ -0,0 +1,47 @@
...
+
+
+/**
+ * @inheritDoc
+ */
+os.ui.filter.op.time.OlderThan.prototype.getEvalExpression = function(varName, literal) {
+  if (!goog.string.isEmptyOrWhitespace(goog.string.makeSafe(literal))) {
+    // the value stored in the variable for a time-based operator is an {@link os.time.ITime}
+    // to execute the filter, we need to extract the raw time value and compare it to now
+    return 'window["currentTimestamp"]-' + varName + '.getEnd()>' + os.ui.filter.string.quoteString(literal);
```
schmidtk: This needs to handle if `varName` is not defined (no time).

src/plugin/featureaction/featureaction.js
jsalankey: Agreed. I wanted to run it by you before making the call to remove it. I wasn't sure if there was some good reason to keep the refresh in for some case.

scss/os/_overrides_jqueryui.scss
```diff
@@ -15,7 +15,7 @@
   }
 
   &.ui-widget-content {
-    border: $border-width solid $border-color;
+    border: 0;
```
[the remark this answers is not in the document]
jsalankey: The spinner elements had a border around them that made them even taller than the rest of the inputs in the filter rows (which are already too tall).

Ever since BS4, the inputs in the filter tree look terrible. We originally had them set to a shorter height, which looked bad, now they're too tall and clip out. I played with upping the row height, but didn't like that either because it chewed up a ton more vertical space. 

I will look at making this more specific or taking it out.

src/os/ui/filter/op/op.js
```diff
@@ -205,7 +212,8 @@ os.ui.filter.op.Op.prototype.getLiteral = function(el) {
  */
 os.ui.filter.op.Op.prototype.matches = function(el) {
   if (el && el.length) {
-    return el[0].localName == this.localName;
+    var hint = el[0].getAttribute('hint');
+    return el[0].localName == this.localName && hint == this.matchHint;
```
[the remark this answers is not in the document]
jsalankey: I think we always need to require both. `this.matchHint` will always default to null, so any op that doesn't have it will pass that part of the check. The problem this was fixing was that the "older than" time operator was matching on the "greater than" operator match because this function only looked at the `localName` and not the hint. 

The reverse case could happen if the order of the ops in the `OPERATORS` array was changed for some reason (i.e. a "greater than" operator could hit on the "older than" operator.

src/os/ui/filter/op/time/betweentimeop.js
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * @inheritDoc
+ */
+os.ui.filter.op.time.Between.prototype.getEvalExpression = function(varName, literal) {
+  var range = this.getRangeFromLiteral(literal);
+  if (range && range.length == 2 && range[0] !== range[1]) {
+    var r0 = range[0];
+    var r1 = range[1];
+    return varName + '!=null?(os.ui.filter.currentTimestamp-' + varName + '.getStart()<=' + r1 +
```
schmidtk: Minor nit, but `!=null&&` would be a bit more concise in these ops.

PR #372, opened 2019-01-02
fix(storage): cls test helper should return promise

### LINKED ISSUE #371: cls helper needs to return promise
The `cls` helper function for tests needs to return the promise so that tests can chain off of it.

This should help with justin-bits/opensphere-tests#15

wallw-bits:
This helps eliminate arbitrary sleeps in test code.

Fixes #371


wallw-bits:
See https://github.com/ngageoint/opensphere/issues/371#issuecomment-450989770

PR #377, opened 2019-01-04
fix(zip): call close on all zip readers

### LINKED ISSUE #373: ZIP workers not being cleaned up
Every call to `zip.createReader` (i.e. https://github.com/ngageoint/opensphere/blob/969e837747c0d1b5cf167073f81886c969e28e71/src/plugin/file/kml/kmlparser.js#L552) creates a new worker with the z-worker.js script. Repeatedly refreshing a zipped file like a KMZ is causing an endless amount of workers to be created and never destroyed. The reader instances have a `close` method on them that calls `worker.terminate()`. We need to refactor everywhere that we create a ZIP worker to keep the reader reference around until we no longer need it and to clean it up.

wallw-bits:
This fixes leaking zip workers each time they are used.

### Test
For each of these, you're going to open dev tools and watch the "Threads" count to ensure that zip workers are not leaked in addition to ensuring that the feature you're using is working properly.

 * Mime stack
   * Load zip files that are not supported layer types (not KMZ or zipped SHP)
 * KMZ
   * Load KMZ files
   * Load KMZ files with external stylesheets in the zip
   * Load KMZ files with external stylesheets that load as a KMZ
   * Load KMZ files with network links that are KMZ files
   * Import KMZ areas into the Areas tab
 * SHP
   * Load zipped SHP files

fixes #373

### DIFF
--- a/src/os/file/mime/zip.js
+++ b/src/os/file/mime/zip.js
@@ -48,6 +48,7 @@ os.file.mime.zip.detectZip = function(buffer, opt_file) {
         try {
           zip.createReader(reader, function(reader) {
             reader.getEntries(function(entries) {
+              reader.close();
               resolve(entries);
             });
           }, function(error) {
--- a/src/os/parse/asynczipparser.js
+++ b/src/os/parse/asynczipparser.js
@@ -1,0 +1,111 @@
+goog.provide('os.parse.AsyncZipParser');
+
+goog.require('os.parse.AsyncParser');
+
+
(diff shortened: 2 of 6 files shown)


schmidtk:
Would it be possible to make an `os.parse.AsyncZipParser` that abstracts out the zip code added here? It seems the same pattern is repeated for each of the parsers.

wallw-bits:
Yup. Done.

PR #380, opened 2019-01-04
fix(wfs): pass custom WFS params in GetFeature requests

wallw-bits:
The custom WFS params set on the server were not being passed through to the GetFeature request.

### DIFF
--- a/src/os/ui/ogc/iogcdescriptor.js
+++ b/src/os/ui/ogc/iogcdescriptor.js
@@ -158,6 +158,19 @@ os.ui.ogc.IOGCDescriptor.prototype.setWmsParams;
 
 
 /**
+ * Get the additional WFS parameters to include in requests
+ * @return {?goog.Uri.QueryData}
+ */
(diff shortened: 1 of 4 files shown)


src/os/ui/ogc/iogcdescriptor.js
```diff
@@ -157,6 +157,19 @@ os.ui.ogc.IOGCDescriptor.prototype.getWmsParams;
 os.ui.ogc.IOGCDescriptor.prototype.setWmsParams;
 
 
+/**
+ * Get the additional WFS parameters to include in requests
+ * @return {?goog.Uri.QueryData}
+ */
+os.ui.ogc.IOGCDescriptor.prototype.getWfsParams;
```
schmidtk: `os.ui.ogc.OGCDescriptor` should implement these as well. It's only used by tests so the compiler didn't catch it. We may also want to move it out of `src` at some point, though not necessary here.

wallw-bits: Done.

PR #389, opened 2019-01-25
fix(AnimationLock): Lock button on timeline moved to the timeline pan…

lilysikes:
…el and when lock button is clic

#352

### DIFF
--- a/src/os/ui/timelinepanel.js
+++ b/src/os/ui/timelinepanel.js
@@ -53,10 +53,18 @@ os.ui.TimelinePanelCtrl = function($scope, $element, $timeout) {
   os.ui.TimelinePanelCtrl.base(this, 'constructor', $scope, $element, $timeout);
 
   /**
+   * @type {boolean}
+   */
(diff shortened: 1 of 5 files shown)


src/os/ui/timelinepanel.js
```diff
@@ -52,12 +52,20 @@ os.ui.resizeMap = 'resizeMap';
 os.ui.TimelinePanelCtrl = function($scope, $element, $timeout) {
   os.ui.TimelinePanelCtrl.base(this, 'constructor', $scope, $element, $timeout);
 
+  /**
+   * @type {boolean}
+   */
+  this['locked'] = os.time.TimelineController.getInstance().getLock();
+
+  this.tlc.listen(os.time.TimelineEventType.LOCK_TOGGLE, goog.partial(this.setLock, true), false, this);
```
jsalankey: This `listen` call needs a corresponding `unlisten` in this directive's destroy function.

Also, `this.setLock` is being bound with an argument of `true`, but that function doesn't take any arguments.

PR #392, opened 2019-01-28
fix(slickgrid): Use case-insensitive sort in grids

welchyd:


src/os/ui/slick/slickgrid.js
```diff
@@ -1321,8 +1321,7 @@ os.ui.slick.SlickGridCtrl.prototype.multiColumnSort = function(cols, a, b) {
     }
 
     if (result == undefined) {
-      // inlined goog.array.defaultCompare to avoid the function call
-      result = (v1 > v2 ? 1 : v1 < v2 ? -1 : 0) * sign;
+      result = goog.string.numerateCompare(v1, v2) * sign;
```
schmidtk: The original default compare was inlined for performance reasons with large data sets, and this will regress that unnecessarily for numeric values that don't benefit from `numerateCompare`. Can you put the inline comparison after the `parseFloat` calls above?

I would also suggest testing this for sorting strings with large (50k+) data sets.

wallw-bits: The number type comparison should be `result = (v1 - v2) * sign` (same result, but simpler).

welchyd: Good call - using `toLowerCase` was much more performant! It works very quickly on 150k strings, and the sort order is similar to the numerateCompare.
I also added the result comparison optimization for numbers

Issue #391, opened 2019-01-28
Annotations should obey timeframe


swtyler:
The time on an annotation is not respected in the timeline. The annotation is present when the active timespan does not overlap. After some discussion, we decided it was best to always make it respect time rather than some sort of checkbox "keep" option.

PR #401, opened 2019-02-18
fix(Annotations): Annotations obey the time on the timeline.

lilysikes:
#391

### DIFF
--- a/src/plugin/file/kml/kmlsource.js
+++ b/src/plugin/file/kml/kmlsource.js
@@ -6,6 +6,7 @@ goog.require('goog.log.Logger');
 goog.require('goog.object');
 goog.require('ol.array');
 goog.require('os.events.PropertyChangeEvent');
+goog.require('os.layer.AnimationOverlay');
 goog.require('os.layer.Image');
 goog.require('os.object');
 goog.require('os.source.PropertyChange');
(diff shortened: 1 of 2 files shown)


src/os/source/vectorsource.js
```diff
@@ -1683,6 +1683,22 @@ os.source.Vector.prototype.getFilteredFeatures = function(opt_allTime) {
       this.timeModel.intersection(range, true);
     }
 
+    // Hide annotations if needed
+    if (this instanceof plugin.file.kml.KMLSource) {
```
schmidtk: Generally speaking, if you only want code to execute in a subclass, you should do so in a function override in that class. This technically creates a circular dependency between `os.source.Vector` and `plugin.file.kml.KMLSource`, though JavaScript lets you get away with this.

This also references plugin code (`plugin.*`) in core application code (`os.*`), which should be avoided. Plugins should be removable without impacting application code.

src/os/source/vectorsource.js
```diff
@@ -1683,6 +1683,22 @@ os.source.Vector.prototype.getFilteredFeatures = function(opt_allTime) {
...
+      for (var i = 0, n = this.getFeatures().length; i < n; i++) {
```
schmidtk: This has the desired effect, but will not be performant for a few reasons:
- This is calling `Array.includes` for each feature in the source. Worst case here is `O(n^2)`, which will not perform well for layers with thousands of features. While unlikely to have thousands of _annotated_ features, we can make this faster.
- The code is in `getFilteredFeatures`, which may be called many times per animation frame because it's used by any data consumer reacting to timeline changes.

I think the best solution here is to override the functions for managing the animation overlay (what we use to quickly animate features) in `plugin.file.kml.KMLSource`. The function overrides should be:

- `createAnimationOverlay`: Called when the timeline is opened. Call the parent function then hide all annotations.
- `disposeAnimationOverlay`: Called when the timeline is closed. Call the parent function then show all annotations (if the feature isn't hidden).
- `updateAnimationOverlay`: Called on each animation frame. Hide annotations for features in the animation overlay (these are from the previous frame), call the parent function, then show annotations for features in the animation overlay (now updated for the current frame).

This will still have an outstanding issue that show/hide via the KML tree checkbox can show the annotation when not in frame, but that should be resolved separately.

src/os/source/vectorsource.js
```diff
@@ -1683,6 +1683,22 @@ os.source.Vector.prototype.getFilteredFeatures = function(opt_allTime) {
...
+        var feature = this.getFeatures()[i];
+        var source = /** @type {plugin.file.kml.KMLSource} */ (this);
+        var annotation = source.getFeatureNode(feature).getAnnotation();
+        if (annotation) {
+          if (features.includes(feature)) {
```
schmidtk: `Array.includes` is not implemented by the minimum supported browsers for Open Sphere (currently Chrome 36 and Firefox 38). Please use `ol.array.includes` instead.

src/os/source/vectorsource.js
lilysikes: I think that I fixed all of these problems. Could you take a look for me?

lilysikes: I think that I fixed all of these comments. Could you take a look for me?

src/plugin/file/kml/kmlsource.js
```diff
@@ -740,3 +741,72 @@ plugin.file.kml.KMLSource.prototype.restore = function(config) {
...
+/**
+ * Creates a basic feature overlay used to animate features on the map.
+ * @protected
+ * @override
+ */
+plugin.file.kml.KMLSource.prototype.createAnimationOverlay = function() {
+  plugin.file.kml.KMLSource.superClass_.createAnimationOverlay.call(this);
+
+  // Hide all annotations
+  for (var i = 0, n = this.getFeatures().length; i < n; i++) {
```
schmidtk: The runtime of `ol.source.Vector#getFeatures` is non-trivial because it creates a new array of features from the underlying r-tree. As a small optimization, call it once and save the result for the two places it's used in this function.

src/plugin/file/kml/kmlsource.js
```diff
@@ -740,3 +741,72 @@ plugin.file.kml.KMLSource.prototype.restore = function(config) {
...
+    // hide annotations in overlay
+    for (var i = 0, n = this.animationOverlay.getFeatures().length; i < n; i++) {
+      var feature = this.animationOverlay.getFeatures()[i];
+      var annotation = this.getFeatureNode(feature).getAnnotation();
+      if (annotation) {
+        annotation.setVisible(false);
+      }
+    }
+
+    plugin.file.kml.KMLSource.superClass_.updateAnimationOverlay.call(this);
```
schmidtk: Please split this function into two separate `if`'s with the parent call in between, so the parent function is always called.

While it wouldn't currently impact execution, we should defensively code to handle changes to `os.source.Vector#updateAnimationOverlay` that do something when the animation overlay is not defined.

PR #413, opened 2019-02-22
fix(kml): fix leak of disposed KML nodes

wallw-bits:
`KMLSource` keeps a map of node IDs to nodes. In the event that the KML is refreshed (or a portion of the tree is refreshed via a network link), nodes that were left out of the resulting merge by the `KMLParser` were already disposed but were still being added back into the map here, causing a leak.

### DIFF
--- a/src/plugin/file/kml/kmlsource.js
+++ b/src/plugin/file/kml/kmlsource.js
@@ -436,18 +436,12 @@ plugin.file.kml.KMLSource.prototype.onImportComplete = function(opt_event) {
  * @inheritDoc
  */
 plugin.file.kml.KMLSource.prototype.clearQueue = function() {
-  var toRemove = [];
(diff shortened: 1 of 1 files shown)


src/plugin/file/kml/kmlsource.js
```diff
@@ -432,6 +432,27 @@ plugin.file.kml.KMLSource.prototype.onImportComplete = function(opt_event) {
...
+  for (var key in this.nodeMap_) {
+    var node = this.nodeMap_[key];
+    if (node && node.isDisposed()) {
+      toRemove.push(key);
+    }
+  }
+
+  var nmap = this.nodeMap_;
+  toRemove.forEach(function(key) {
+    nmap[key] = undefined;
```
schmidtk: Can this be done directly in the above loop to avoid the extra `toRemove` array?

PR #415, opened 2019-02-25
fix(LayerReset): When slider bars are reset the settings are checked …

lilysikes:
…to see if they have a different

### DIFF
--- a/src/os/data/layersyncdescriptor.js
+++ b/src/os/data/layersyncdescriptor.js
@@ -143,13 +143,27 @@ os.data.LayerSyncDescriptor.prototype.getOptions = function() {
   var options = this.getLayerOptions();
   if (options) {
     if (!goog.isArray(options)) {
+      options['defaults'] = this.extractConfigDefaults(options);
       options = [options];
     }
-
     return options.map(this.applyLayerConfig, this);
   }
 
   return options;
+};
+
+
+/**
+ * @param {!Object<string, *>} options
+ * @return {!Object<string, *>}
+ */
+os.data.LayerSyncDescriptor.prototype.extractConfigDefaults = function(options) {
+  return {
+    'opacity': options['opacity'],
+    'brightness': options['brightness'],
+    'contrast': options['contrast'],
+    'saturation': options['saturation']
+  };
 };
 
 
--- a/src/os/ui/layer/defaultlayerui.js
+++ b/src/os/ui/layer/defaultlayerui.js
@@ -433,6 +433,15 @@ os.ui.layer.DefaultLayerUICtrl.prototype.onRefreshChange = function() {
  */
 os.ui.layer.DefaultLayerUICtrl.prototype.reset = function(key) {
   var defaultValue = this.defaults[key];
+  var nodes = this.getLayerNodes();
+  if (nodes && nodes[0].getLayer() != null && nodes[0].getLayer().getLayerOptions() &&
+    nodes[0].getLayer().getLayerOptions()['defaults']) {
+    var value = nodes[0].getLayer().getLayerOptions()['defaults'][key];
+    if (value != null) {
+      defaultValue = /** @type {number} */ (value);
+    }
+  }
+
   if (defaultValue != null) {
     var callback = this.getProperties()[key];
     this.onSliderStop(callback, key, null, defaultValue);
--- a/src/plugin/places/kmlplacesimportui.js
+++ b/src/plugin/places/kmlplacesimportui.js
@@ -151,7 +151,7 @@ plugin.places.KMLPlacesCtrl.prototype.onImportComplete_ = function(event) {
   var importer = /** @type {os.im.Importer} */ (event.target);
   var nodes = importer.getData();
   for (var i = 0; i < nodes.length; i++) {
-    if (nodes[i].layerUI == null) {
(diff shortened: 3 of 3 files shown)


src/os/ui/layer/defaultlayerui.js
```diff
@@ -433,6 +433,16 @@ os.ui.layer.DefaultLayerUICtrl.prototype.onRefreshChange = function() {
  */
 os.ui.layer.DefaultLayerUICtrl.prototype.reset = function(key) {
   var defaultValue = this.defaults[key];
+  var nodeLabel = this.getLayerNodes()[0].label;
+  if (nodeLabel) {
+    nodeLabel = nodeLabel.replace(/ /g, '');
+    nodeLabel = nodeLabel.toLowerCase();
+    var value = /** @type {number} */ (os.settings.get(['providers', 'basemap', 'maps', nodeLabel, key]));
```
wallw-bits: Layers can come from many more providers than just the `BaseMapProvider`, whose settings you have hard coded here. A better way to do this would be to have the thing making the layer config (generally either the provider or descriptor) store the default value on the config object. The best place to do that is probably in `os.data.LayerSyncDescriptor.prototype.getOptions`, just before the user layer config is applied. Then this function can loop over each layer's `getLayerOptions()['defaultOpacity']` or whatever you end up calling it.

Note that there is no longer a single default value if multiple layers are selected. Each layer will (potentially) have its own default and that will need to be addressed as well.

justin-bits:
@lilysikes, the build is failing due to missing the latest changes with Cypress. Merging in those changes should fix it.

src/os/data/layersyncdescriptor.js
```diff
@@ -142,17 +142,31 @@ os.data.LayerSyncDescriptor.prototype.getLayerOptions = goog.abstractMethod;
 os.data.LayerSyncDescriptor.prototype.getOptions = function() {
   var options = this.getLayerOptions();
   if (options) {
+    options['defaults'] = this.extractConfigDefaults(options);
```
wallw-bits: This line needs to be done after the `if` which follows it and in a form that works on arrays. This function is given either a single `options` or an array of them. This line as is does not work for the array case.

src/os/ui/layer/defaultlayerui.js
```diff
@@ -433,6 +433,14 @@ os.ui.layer.DefaultLayerUICtrl.prototype.onRefreshChange = function() {
  */
 os.ui.layer.DefaultLayerUICtrl.prototype.reset = function(key) {
   var defaultValue = this.defaults[key];
+  var nodes = this.getLayerNodes();
+  for (var i = 0, n = nodes.length; i < n; i++) {
+    var value = nodes[i].getLayer().getLayerOptions()['defaults'][key];
+    if (value) {
```
schmidtk: Two things here:
- The default value could be 0, so this needs to be `value != null`.
- This for loop is overwriting `defaultValue` on each iteration, which results in applying the default for the last layer in the array to all layers. It should either reset each layer individually to its own default, or use the first node in the array (we use the first to populate the control when multiple layers are selected).

@wallw-bits what are your thoughts on the multi-select behavior? I'm thinking each layer should be reset to its own default value, then the control value updated to reflect the first layer.

src/os/ui/layer/defaultlayerui.js
```diff
@@ -433,6 +433,12 @@ os.ui.layer.DefaultLayerUICtrl.prototype.onRefreshChange = function() {
  */
 os.ui.layer.DefaultLayerUICtrl.prototype.reset = function(key) {
   var defaultValue = this.defaults[key];
+  var nodes = this.getLayerNodes();
+  var value = nodes[0].getLayer().getLayerOptions()['defaults'][key];
```
schmidtk: This should be a bit more defensive instead of assuming everything in the call chain will be defined. The node list could be empty, `getLayer` and `getLayerOptions` are both allowed to return `null`, and `defaults` is currently only in layer options for layers created by a `LayerSyncDescriptor`. While it's admittedly unlikely a node would have a reset button and fall under these cases, it's worth verifying to avoid an exception.

src/os/data/layersyncdescriptor.js
```diff
@@ -143,16 +143,30 @@ os.data.LayerSyncDescriptor.prototype.getOptions = function() {
...
-
     return options.map(this.applyLayerConfig, this);
   }
 
   return options;
 };
 
 
+/**
+ * @param {(Array<!Object<string, *>>|Object<string, *>)} options
```
schmidtk: Both the `@param` and `@return` for this function should be `!Object<string, *>`. It's assuming an object input, and always returns an object.

src/plugin/places/kmlplacesimportui.js
```diff
@@ -151,7 +151,7 @@ plugin.places.KMLPlacesCtrl.prototype.onImportComplete_ = function(event) {
   var importer = /** @type {os.im.Importer} */ (event.target);
   var nodes = importer.getData();
   for (var i = 0; i < nodes.length; i++) {
-    if (nodes[i].label == 'kmlroot' || nodes[i].label == 'Saved Places') {
+    if (nodes[i].layerUI == null) {
```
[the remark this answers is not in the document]
lilysikes: Woops that was a change for a different PR. I must have put it on the wrong branch. I will change that.

PR #420, opened 2019-02-28
Animate reset rotation/view in 3D mode.

schmidtk:
This also removes the reset tilt/roll key bindings, which were both hidden and buggy.

Things to test:
- R resets both rotation and tilt in 3D, and animates the transition. We don't allow rotation in 2D so this has no impact.
- V resets the view in 2D and 3D, animated in 3D.


justin-bits:
@schmidtk, Help>Controls needs to be updated.
image

schmidtk:
The controls have been restored with animation, but Reset Roll still needs work to avoid also changing camera tilt (#432).

Issue #416, opened 2019-02-26
Sync tree/timeline visibility states for KML overlays/annotations


schmidtk:
KML overlays (GroundOverlay, ScreenOverlay, and annotations) should all animate with the timeline when they have a time component, and animation visibility should sync with the tree state visibility.

To achieve this, we'll should do the following:

1. Parse time values from KML for both GroundOverlay and ScreenOverlay elements.
2. Animate GroundOverlay and ScreenOverlay with the timeline. This should build on the work done in #401 to animate annotations.
3. For all three overlay types, ensure the timeline visibility state and tree node visibility state work together so overlays are correctly displayed at all times.

PR #423, opened 2019-03-01
Refactor KML GroundOverlay and ScreenOverlay parsers

schmidtk:
This refactors overlay parsers to use the `ol.xml` API, and parses additional fields like time to support #416.

Both overlay types should behave as before, except this fixes a bug where a `ScreenOverlay` with `visibility = 0` would not be added to the KML tree.

This file can be used as an example for testing (rename to `.kmz`):

[Overlays.zip](https://github.com/ngageoint/opensphere/files/2920862/Overlays.zip)

### DIFF
--- a/externs/os.externs.js
+++ b/externs/os.externs.js
@@ -375,6 +375,21 @@ osx.window.ConfirmColumnOptions;
 
 
 /**
+ * @typedef {{
+ *   id: string,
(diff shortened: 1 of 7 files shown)


schmidtk:
About done with the remaining work in #416, so closing this to make a single PR.

PR #434, opened 2019-03-11
fix(cesium): avoid showing full world areas

### LINKED ISSUE #396: Whole world query throws Cesium DeveloperErrors
There is a check in the Cesium createPolylineGeometry.js worker script that doesn't like the fact that there is no angle between the first and last coordinates of the whole world geometry. This causes it to throw a DeveloperError, so it doesn't affect the compiled version of the code.

The 3D view of the whole world query looks terrible anyway (it's just a line at the antimeridian), and we can't think of a better way to draw it, so putting a check into the synchronizer to tell it not to draw a whole world area is probably the only solution.

wallw-bits:
fixes #396

### DIFF
--- a/src/plugin/cesium/sync/featureconverter.js
+++ b/src/plugin/cesium/sync/featureconverter.js
@@ -1654,7 +1654,7 @@ plugin.cesium.sync.FeatureConverter.prototype.olGeometryToCesium = function(feat
     geomType = 'ellipse';
   }
 
-  if (os.query.isWorldQuery(geometry)) {
+  if (geometry.getType() === ol.geom.GeometryType.POLYGON && os.query.isWorldQuery(geometry)) {
     // do not show these
     return;
   }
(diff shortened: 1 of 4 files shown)


src/plugin/cesium/sync/featureconverter.js
```diff
@@ -1653,6 +1654,11 @@ plugin.cesium.sync.FeatureConverter.prototype.olGeometryToCesium = function(feat
     geomType = 'ellipse';
   }
 
+  if (os.query.isWorldQuery(geometry)) {
```
schmidtk: `os.query.isWorldQuery` doesn't work in 3857 because the tolerance of the `goog.math.nearlyEquals` call is too small given the projection units. We should change that function to test that the geometry's area is within a certain (very small) percent of the world area.

It might also be a good idea to cache the world area by projection, so we only have to call transformExtent/getArea on the world extent once per projection.

PR #438, opened 2019-03-11
fix(CSVExport): Prevent hidden columns from being included in export

MarcelloNicoletti:
Geometry and time fields were previously always included in exports. This commit adds an option to
keep them out of the export if they're hidden.

BREAKING CHANGE: Even though this defaults to the old behavior I haven't had time to ensure this doesn't break other uses of it

WIP:

- [x] I'd like some review on my naming. `alwaysIncludeGeometryAndTime` might be too verbose
    * Replaced with `alwaysIncludeLatLonTime` since not all geometry is included with this option
- [x] I'd like a little bit of feedback on my documentation
    * Adding notes that this is not a stand alone UI
- [x] Is there anywhere else that could benefit from this option?
- [x] I still need to write tests in `test/plugin/file/csv/csvexporter.test.js` that utilize `includeGeometry`

### DIFF
--- a/src/os/source/source.js
+++ b/src/os/source/source.js
@@ -227,9 +227,10 @@ os.source.getHoldRecordTime = function(item) {
  * Get the fields to export for features in the source.
  * @param {os.source.Vector} source The source
  * @param {boolean=} opt_internal If internal fields should be included
+ * @param {boolean=} opt_includeTime If the time field should be included
  * @return {Array<string>} The fields to export
  */
-os.source.getExportFields = function(source, opt_internal) {
+os.source.getExportFields = function(source, opt_internal, opt_includeTime) {
   var fields = null;
 
   if (source) {
@@ -251,7 +252,8 @@ os.source.getExportFields = function(source, opt_internal) {
         if (column['visible'] &&
             !goog.string.isEmptyOrWhitespace(goog.string.makeSafe(field)) &&
             !ol.array.includes(fields, field) &&
-            (opt_internal || !os.feature.isInternalField(field))) {
+            (opt_internal || !os.feature.isInternalField(field) ||
+            (opt_includeTime && field == os.data.RecordField.TIME))) {
           fields.push(field);
         }
       }
--- a/src/os/ex/abstractexporter.js
+++ b/src/os/ex/abstractexporter.js
@@ -90,6 +90,14 @@ os.ex.AbstractExporter.prototype.supportsMultiple = function() {
 /**
  * @inheritDoc
  */
+os.ex.AbstractExporter.prototype.supportsTime = function() {
+  return false;
+};
+
+
+/**
+ * @inheritDoc
+ */
 os.ex.AbstractExporter.prototype.reset = function() {
   this.fields = null;
   this.items = null;
--- a/src/os/ex/iexportmethod.js
+++ b/src/os/ex/iexportmethod.js
@@ -77,6 +77,13 @@ os.ex.IExportMethod.prototype.supportsMultiple;
 
 
 /**
+ * If the export method supports exporting time from the data source.
+ * @return {boolean}
+ */
+os.ex.IExportMethod.prototype.supportsTime;
+
+
+/**
  * Begins the export process.
  */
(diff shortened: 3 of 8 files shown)


swtyler:
- I would change the verbiage from includeGeometry. Technically it's not including all geometry (e.g. if I hide the SEMI_MAJOR field, it won't be included). It's more or less `includeLatLonTime` which is overly verbose but more correct. 
- The only change I recommend - make it clear in your documentation that these are the options provided to exportdialog.js.
  - I personally don't like the names for the directive or file (plugin.file.csv.ui.CSVExportCtrl and csvexportui.js). They seem to imply that it's the entire UI for the export but all it's doing is adding a checkbox to an existing export UI... (e.g. call it CSVExportOptionsCtrl). That has nothing to do with your changes since they mirror the KML export.
- In terms of what else might benefit, GeoJSON and KML properly ignore hidden columns. On export, SHP files bring back in all of the same things as CSV (LAT_DMS, MGRS, etc.). We could probably add this option there as well (it already has one for including Ellipses).

MarcelloNicoletti:
I've changed the variable to `alwaysIncludeLatLonTime` and I've added a note that the `csvexportui` directive is meant for use within `exportdialog.js`

src/os/ui/file/csv/abstractcsvexporter.js
```diff
@@ -97,6 +97,34 @@ os.ui.file.csv.AbstractCSVExporter.prototype.process = function() {
 os.ui.file.csv.AbstractCSVExporter.prototype.processItem = goog.abstractMethod;
 
 
+/**
+ * Get whether to always include WKT geometry in the export
+ * @return {boolean} If true, WKT geometry will always be included in the export
+ */
+os.ui.file.csv.AbstractCSVExporter.prototype.getAlwaysIncludeWkt = goog.abstractMethod;
```
schmidtk: These functions are specific to geospatial/temporal CSV export, so I would leave them solely on `plugin.file.csv.CSVExporter`. Your CSV export UI can assume its exporter is a `plugin.file.csv.CSVExporter`. 

If other CSV exporters have a need for these functions in the future, we could extract the API to an interface and use `os.implements` to detect support from each exporter. That's more work than necessary at the moment.

MarcelloNicoletti:
I still haven't figured out the time column order. Mainly I don't want to over complicate the logic due to time being either an instant or a duration. Adding multiple conditions to the fields loop isn't ideal

MarcelloNicoletti:
I feel that this is ready for another review.

src/os/source/source.js
```diff
@@ -251,7 +252,9 @@ os.source.getExportFields = function(source, opt_internal) {
         if (column['visible'] &&
             !goog.string.isEmptyOrWhitespace(goog.string.makeSafe(field)) &&
             !ol.array.includes(fields, field) &&
-            (opt_internal || !os.feature.isInternalField(field))) {
+            (opt_internal || !os.feature.isInternalField(field) ||
+            (opt_includeTime && field == os.data.RecordField.TIME))) {
+              // Todo: `os.data.RecordField.TIME == 'recordTime'` and is written as such in the column
```
schmidtk: What outstanding work is there for this todo?

schmidtk:
Noticed the export ellipses option (from a previous PR) is not using the custom checkbox, which looks weird next to the WKT one. Fixed in #508.

src/os/source/source.js
MarcelloNicoletti: Oh, I forgot to clean this up after taking in your suggested code.

MarcelloNicoletti:
@schmidtk Ok, I've removed that extra todo. Since you've already made #508 I don't think there's anything else to do here right?

schmidtk:
> @schmidtk Ok, I've removed that extra todo. Since you've already made #508 I don't think there's anything else to do here right?

Yep this looks good!

MarcelloNicoletti:
Ok, branch is up to date with master and Travis should pass. It looks like I can't request a reviewer with my current permissions so you'll have to self request again

Issue #360, opened 2018-12-19
Multiple Issues Importing Saved Places


swtyler:
~~First Error~~ OBE
------------
1. Save a place with a style set to None
2. Edit the place - check the console for the error

This appears to have been fixed with the introduction of annotations.

Second Error
------------
1. Export places as a KML(directly from the Places tab)
2. Import places (Places tab), you cannot edit them or remove them

Third Error
------------
1. Export places as KML (directly from the Layers tab)
2. Import places (Places tab), this will more or less crash the app

~~Fourth Error~~ OBE
------------
1. Create a saved place and add a description of some sort
2. Turn on show annotation
3. Export the saved place
4. Reset settings
5. Import the saved place - you will have 2 of the same annotation (drag one, they will be on top of each other) and one can never be removed

~~Fifth Error~~ OBE
------------
1. Create a saved place and set a time instant or range
2. Re-import the saved place - the time is dropped from the feature info and missing when you edit. It is there when you drag drop the places kmz file into the tool, so it's something specific to the UI or the way places are being rebuilt.

~~Sixth Error (likely related the previous two)~~ OBE
------------
1. Export places and import them back via both means, the Places tab import and drag/drop the kmz Opensphere.
2. Zoom a fair amount out on the map
3. Double click one of the features, this should pull up the "multiple feature info" window because they are laid over the top of each other
4. Inconsistently the window breaks with the following error:

```
aTimeObj.getStart is not a function
at Object.os.feature.sortByTime (feature.js:976)
.....
at os.ui.feature.MultiFeatureInfoCtrl (multifeatureinfo.js:101)
```

swtyler:
Closed via https://github.com/ngageoint/opensphere/pull/453

PR #453, opened 2019-03-25
fix(ImportPlaces): Saved Places imports kml files successfully.

lilysikes:
Addresses issues in #360.

### DIFF
--- a/src/plugin/places/kmlplacesimportui.js
+++ b/src/plugin/places/kmlplacesimportui.js
@@ -156,14 +156,26 @@ plugin.places.KMLPlacesCtrl.prototype.onImportComplete_ = function(event) {
       i--;
     }
   }
-  // plugin.places.menu.saveKMLToPlaces_(nodes);
   var rootNode = plugin.places.PlacesManager.getInstance().getPlacesRoot();
   if (rootNode) {
     var cmds = [];
     for (var i = 0; i < nodes.length; i++) {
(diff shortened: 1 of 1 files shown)


src/plugin/places/kmlplacesimportui.js
```diff
@@ -156,14 +156,23 @@ plugin.places.KMLPlacesCtrl.prototype.onImportComplete_ = function(event) {
...
   var rootNode = plugin.places.PlacesManager.getInstance().getPlacesRoot();
   if (rootNode) {
     var cmds = [];
     for (var i = 0; i < nodes.length; i++) {
-      var cmd = new plugin.file.kml.cmd.KMLNodeAdd(nodes[i], rootNode);
-      cmd.title = 'Save ' + nodes[i].getLabel() + ' to Places';
-      cmds.push(cmd);
+      var feature = nodes[i].getFeature();
+      if (feature) {
+        if (nodes[i].annotation_) {
```
swtyler: You will want to call `getAnnotation()` like you did here: https://github.com/ngageoint/opensphere/pull/401/files so that this works in compiled mode.

schmidtk: You shouldn't be interacting with the annotation outside of `KMLNode`, which is why I removed `getAnnotation`. Places should be imported exactly as they were exported, so there is no need to hide annotations manually here.

If you must hide it, the correct place to avoid conflicts with tree/animation visibility is the `show` property in the annotation options (`feature.get(os.annotation.OPTIONS_FIELD)`).

PR #459, opened 2019-03-26
docs(test): Corrects commands used to start Cypress and adds addition…

justin-bits:
…al information


justin-bits:
I neglected to update the readme.md file.  Updated.

README.md
```diff
@@ -107,15 +107,10 @@ Software source code previously released under an open source license and then m
 End to end tests for Opensphere are written using [Cypress](https://github.com/cypress-io/cypress).
 
 ### Setup
-By default the ```baseURL``` is set to the [local compiled build](http://localhost:8282/workspace/opensphere/dist/opensphere), but this can be changed in cypress.json to an alternate  target.
+By default the ```baseURL``` is set to the [local compiled build](http://localhost:8282/workspace/opensphere/dist/opensphere), but this can be changed in [cypress.json](https://github.com/ngageoint/opensphere/blob/master/cypress.json) to an alternate  target.
```
wallw-bits: I wouldn't link the local compiled build, or if you do, you should do it with the URL visible. That is not the local compiled build for all setups. Most of us do not use that at all (we use nginx or Apache HTTPD).

justin-bits: Updated.

PR #463, opened 2019-04-02
fix(cesium): replace use of Cesium's camera.computeViewRectangle

wallw-bits:
Replaces use of Cesium's `Camera.prototype.computeViewRectangle()` because it is inaccurate in several ways. It only uses the corners of the screen as sample points (4 compared to 64 now), and it also gives up and assumes that the rectangle is `Rectangle.MAX_VALUE` if at least 3 of those corners are in space (`Rectangle.MAX_VALUE` is not ever a correct view rectangle because the camera can only see roughly half the globe at once).

### DIFF
--- a/externs/Cesium.externs.js
+++ b/externs/Cesium.externs.js
@@ -592,7 +592,7 @@ Cesium.Camera.prototype.direction;
 
 
 /**
- * @type {Cesium.PerspectiveFrustrum}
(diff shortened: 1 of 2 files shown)


bradh:
Upstreamable?

wallw-bits:
Possibly. I got sidetracked and I'll look into it this week.

PR #469, opened 2019-04-05
WIP: fix antimeridian queries via second r-tree

### LINKED ISSUE #468: Spatial index query over antimeridian
`os.source.Vector` indexes spatially using an R-Tree (via the parent `ol.source.Vector`). That R-Tree uses extents normalized to the projection extent (longitudes in the range `[-180, 180]` for `EPSG:4326`). Extents that cross the antimeridian such as `[-190, -170]` or `[170, 190]` are thus cut off halfway through their range.

Potential solutions:
1. (Faster, more memory): Keep a second R-Tree with features indexed with their extents normalized to `[0, 360]` (the inverse from `[-180, 180]`). Any extent which crosses the antimeridian should use this R-Tree rather than the original.
1. (Slower, less memory): Split any extent crossing the antimeridian into two extents. Query the R-Tree for both extents and dedupe the resulting feature arrays.

Testing:
`os.source.Vector.getFeaturesInExtent`:
1. Load features over the antimeridian
1. Draw area over the antimeridian
1. Choose "Select"
1. Select the layer in the Layers window and turn on labels

`os.extent.getFunctionalExtent`
1. Load data only near the antimeridian
1. Right-click the layer and choose "Zoom"

wallw-bits:
Add a second r-tree indexed with extents in the range [0, 360] and use that r-tree when the extent query crosses the antimeridian.

resolves #468

### DIFF
--- a/src/os/extent.js
+++ b/src/os/extent.js
@@ -1,4 +1,5 @@
 goog.provide('os.extent');
+goog.require('os.geo2');
 
 
 /**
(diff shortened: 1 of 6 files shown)


wallw-bits:
This is marked as WIP because it should not be merged until it is compared with the [other solution PR](#470).

wallw-bits:
Closing because we are going with #470 instead.

PR #470, opened 2019-04-05
Fix antimeridian queries via multiple world queries

### LINKED ISSUE #468: Spatial index query over antimeridian

wallw-bits:
Mixes in functionality to the OL RBush wrapper that queries for multiple 'worlds' of an extent and
dedupes those results.

Edit:
This now also addresses hit detection over the antimeridian from OpenLayers. A good test file for both of these items is
[KML Test.txt](https://github.com/ngageoint/opensphere/files/3074784/KML.Test.txt)

resolves #468

### DIFF
--- a/src/os/ui/ol/interaction/dragboxinteraction.js
+++ b/src/os/ui/ol/interaction/dragboxinteraction.js
@@ -1,6 +1,6 @@
 goog.provide('os.ui.ol.interaction.DragBox');
 goog.require('ol.MapBrowserEvent');
-goog.require('os.geo');
+goog.require('os.geo2');
 goog.require('os.olm.render.Box');
 goog.require('os.ui.ol.interaction.AbstractDrag');
 
@@ -54,9 +54,7 @@ os.ui.ol.interaction.DragBox.prototype.getGeometry = function() {
   var geom = this.box2D.getOriginalGeometry();
 
   if (geom) {
-    geom.toLonLat();
-    os.geo.normalizeGeometryCoordinates(geom);
-    geom.osTransform();
+    os.geo2.normalizeGeometryCoordinates(geom);
   }
 
   return geom;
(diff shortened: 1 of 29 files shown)


wallw-bits:
This is marked WIP because it shouldn't be merged until being compared with the [other solution PR](#469).

src/os/ui/ol/interaction/dragboxinteraction.js
```diff
@@ -93,14 +91,13 @@ os.ui.ol.interaction.DragBox.prototype.update = function(mapBrowserEvent) {
       this.extent = [];
     }
 
-    var start = ol.proj.toLonLat(this.startCoord, this.getMap().getView().getProjection());
-    var end = ol.proj.toLonLat(this.endCoord, this.getMap().getView().getProjection());
+    if (this.startCoord && this.endCoord) {
+      this.extent[0] = Math.min(this.startCoord[0], this.endCoord[0]);
```
schmidtk: `update2D` expects the extent to be in lon/lat, so drag boxes are only working in 4326.

PR #472, opened 2019-04-09
fix(geo): Enforce DMS/DDM requirements in regexes.

### LINKED ISSUE #471: Leading zeros on lat/long
Opensphere does not plot DMS coordinates correctly if there is not a leading zero. For example, if you search for the coordinate 380516N 25360W, it gives a different location than 380516N 025360W gives.

schmidtk:
DMS Changes
- Longitudes with 3 digits must have a 0 or 1 as the first digit.
- Seconds are now required.

DDM Changes
- Longitudes with 3 digits must have a 0 or 1 as the first digit.
- Decimal minutes are now optional. The group isn't captured to avoid changes to parser configs.

Fixes #471.

### DIFF
--- a/src/os/geo/geo.js
+++ b/src/os/geo/geo.js
@@ -190,19 +190,33 @@ os.geo.END_ = '[^NSEW\\d]*$';
 
 
 /**
+ * Matches the longitude portion of a DMS coordinate.
+ *  - Preceded by optional direction/white space/sign.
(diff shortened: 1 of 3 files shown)


schmidtk:
Updated the new coordinate test to reflect [this comment](https://github.com/ngageoint/opensphere/issues/471#issuecomment-481371329).

Issue #398, opened 2019-02-13
Remove dependency on third party server


justin-bits:
Per @wallw-bits from https://github.com/ngageoint/opensphere/pull/390:

> The GeoServer and Arc tests need to be refactored to spin up a container and hit that rather than depending on third-party services for tests (this should be easy for GeoServer, possibly less easy for Arc -- but there are better options for live servers there than a local gov one). Additionally, the state file test is also depending on a third-party server when it doesn't have to

justin-bits:
We might just want to stub out the server responses instead: https://docs.cypress.io/guides/guides/network-requests.html

wallw-bits:
That would be best in my opinion. We commonly do that for unit tests already.

justin-bits:
https://github.com/ngageoint/opensphere/pull/481/

bradh:
Is anything more needed on this post #481?

justin-bits:
Nope

PR #481, opened 2019-04-15
test(cypress): Removes dependency on third party server for feature requests

justin-bits:
- Uses stubs for server responses (excluding base map tiles)
- Renames load-state-file to more accurately reflect that it’s testing an arcgis server (additional state file smoke tests to come in a separate PR)
- Adds subfolders to contain files for each test; renamed support files, updates tests accordingly
- Whitelisted XHR requests containing ‘workspace’ and ‘arcgisonline’ (this just removes some output from test runner)

Addresses https://github.com/ngageoint/opensphere/issues/398


justin-bits:
Tile images need to be stubbed. Hold on merge.

schmidtk:
Image stubbing should be added by https://github.com/cypress-io/cypress/issues/687, so we should track that issue to finish up this work.

PR #482, opened 2019-04-16
feat(map): Scale down icons/labels as the map/globe is zoomed out.

### LINKED ISSUE #374: Scale icon size at low camera zoom levels
Google Earth scales icons (after a certain threshold) as the camera zoom is reduced.  Recommend adding this capability OpenSphere.

Note in this example how the icons are much smaller in Google Earth than OpenSphere.  Depending on the number of features and icon selection, the icons can end up completely covering the globe in OpenSphere as the camera is zoomed out.  With Google Earth, the icons become smaller, then eventually disappear.  They never end up covering the globe.
google earth
opensphere

schmidtk:
[Demo Here](https://schmidtk.github.io/opensphere-zoom-scale/)

This attempts to mimic the behavior of Google Earth, where icons and labels are scaled/hidden when zooming far out. The general goal with this approach was to get the behavior as close as possible, without impacting performance.

* OpenLayers: Mixin the image/label canvas renderers to apply the scale value without modifying OpenLayers style objects. This avoids changing individual styles on features, which we know to be costly. I did not scale label opacity (like Cesium/GE) either, because that would require manipulating color strings on each render. Instead they're hidden (scale = 0) at roughly the same level.
* Cesium: Use Cesium's `NearFarScalar` to scale billboards/labels, and to fade out labels.

There is one potentially odd side effect to our Cesium highlight behavior - highlighted items will appear at normal size. This happens because we adjust `eyeOffset` to put highlighted items immediately in front of the camera, so they're always visible. Because they're moved closer to the camera, Cesium will not scale them. I'm not sure if that should be considered a bug or a feature, but it is inconsistent with 2D behavior and would probably require a hack to make them consistent.

Resolves #374.

### DIFF
--- a/externs/Cesium.externs.js
+++ b/externs/Cesium.externs.js
@@ -445,6 +445,7 @@ Cesium.BillboardCollection = function(opt_options) {};
  *   pixelOffsetScaleByDistance : (Cesium.NearFarScalar|undefined),
  *   scale: (number|undefined),
  *   scaleByDistance: (Cesium.NearFarScalar|undefined),
+ *   translucencyByDistance: (Cesium.NearFarScalar|undefined),
  *   position: !Cesium.Cartesian3,
  *   geomRevision: (number|undefined)
  * }}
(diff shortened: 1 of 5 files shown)


justin-bits:
Tested, looks good.

jsalankey:
This is causing our standard point features to look very small at low zoom levels, to the point where I think it's going to bug people. It's a pretty major UX change that affects all vector layers. 

Is there some tuning work that we can do here to increase the minimum zoom size? Or some way to make it not affect point styles? I think the point of doing this was really more for non-point icons.

PR #485, opened 2019-04-17
WIP: Load OL tiles through the request stack.

### LINKED ISSUE #484: Load tiles using the request stack
- Allows requests to be cancelled on layer removal, date change, timeline update, zoom change
- Allows requests to be stubbed in Cypress tests

schmidtk:
This loads all tiles through `os.net.Request`. Tiles should be tested for all sources (WMS, WMTS, ArcGIS, etc) to ensure they're requested through our stack (then via a `blob://` URL), and are still colorable as before.

Resolves #484.

TODO:
- [x] Confirm compatibility with https://github.com/wallw-bits/opensphere-plugin-geopackage. [PR here](https://github.com/wallw-bits/opensphere-plugin-geopackage/pull/10).
- [x] Ensure tiles are disposed on layer removal to avoid leaking object URL's.
- [x] Cancel pending tile requests on layer removal.
- [ ] Determine if `image` tile loading performance can be matched with XHR loading.


schmidtk:
Marked WIP because OL doesn't expire/dispose its tile cache on layer removal, which seems generally bad. It's certainly going to be worse with the use of `URL.createObjectURL` in this ticket.

wallw-bits:
Can we remove the proxy wrapper check for tile layers with this addition? Also, how do we deal with layers which have `proxy: false` set?

schmidtk:
Loading images over XHR loses browser optimizations and reduces tile load performance more than seems acceptable for the benefits of this change. Closing this.

PR #487, opened 2019-04-17
Refresh

lilysikes:
#613 
For this PR I was working on the following problem: 
"We were asked to enable feature actions for merged/joined layers, which we can do now that they no longer require their data sources to implement refresh functionality. Change featureactionmenu.js to not bother checking whether the layer implements IImportSource and make sure that feature actions on static data layers don't persist on refresh (because the layers themselves don't persist on refresh)."

### DIFF
--- a/src/plugin/featureaction/featureactionmenu.js
+++ b/src/plugin/featureaction/featureactionmenu.js
@@ -54,7 +54,7 @@ plugin.im.action.feature.visibleIfSupported = function(context) {
     var layers = os.ui.menu.layer.getLayersFromContext(context).filter(os.MapContainer.isVectorLayer);
     if (layers && layers.length == 1 && layers[0].getOSType() != os.layer.LayerType.REF) {
       var source = /** @type {ol.layer.Layer} */ (layers[0]).getSource();
-      this.visible = os.implements(source, os.source.IImportSource.ID);
+      this.visible = source != null;
     }
   }
 };
--- a/src/os/layer/vector.js
+++ b/src/os/layer/vector.js
@@ -847,20 +847,22 @@ os.layer.Vector.prototype.isFilterable = function() {
  */
 os.layer.Vector.prototype.getFilterKey = function() {
   var options = this.getLayerOptions();
-  var id = /** @type {string} */ (options['id']);
+  if (options) {
+    var id = /** @type {string} */ (options['id']);
 
-  // try to get it from the descriptor
-  var d = os.dataManager.getDescriptor(id);
-  if (os.implements(d, os.filter.IFilterable.ID)) {
-    return /** @type {!os.filter.IFilterable} */ (d).getFilterKey();
-  }
+    // try to get it from the descriptor
+    var d = os.dataManager.getDescriptor(id);
+    if (os.implements(d, os.filter.IFilterable.ID)) {
+      return /** @type {!os.filter.IFilterable} */ (d).getFilterKey();
+    }
 
-  // try to derive it from the layer options
-  var url = /** @type {string} */ (options['url']);
-  var params = /** @type {string} */ (options['params']);
-  var typeName = params ? /** @type {string} */ (params.get('typename')) : null;
-  if (url && typeName) {
-    return url + '!!' + typeName;
+    // try to derive it from the layer options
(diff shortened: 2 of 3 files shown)


src/plugin/featureaction/featureactionmenu.js
```diff
@@ -19,7 +19,6 @@ plugin.im.action.feature.layerSetup = function() {
       eventType: plugin.im.action.feature.EventType.LAUNCH,
       tooltip: 'Perform actions on imported data matching a filter',
       icons: ['<i class="fa fa-fw ' + os.im.action.ICON + '"></i>'],
-      beforeRender: plugin.im.action.feature.visibleIfSupported,
```
schmidtk: This will show the Feature Actions option for all layers, which is not desirable. We still want to filter out non-vector layers, and reference layers (drawing/search/etc).

src/plugin/featureaction/featureactionmenu.js
```diff
@@ -41,25 +40,6 @@ plugin.im.action.feature.layerDispose = function() {
...
- */
-plugin.im.action.feature.visibleIfSupported = function(context) {
-  this.visible = false;
-
-  var am = os.im.action.ImportActionManager.getInstance();
-  if (am && am.hasActions() && context && context.length == 1) {
-    var layers = os.ui.menu.layer.getLayersFromContext(context).filter(os.MapContainer.isVectorLayer);
-    if (layers && layers.length == 1 && layers[0].getOSType() != os.layer.LayerType.REF) {
-      var source = /** @type {ol.layer.Layer} */ (layers[0]).getSource();
-      this.visible = os.implements(source, os.source.IImportSource.ID);
```
schmidtk: Keep this function, but change this line to `this.visible = source != null;`.

schmidtk:
If a PR isn't linked to an issue, please explain why the changes are being made in the PR description. This will help us review the changes properly. Thanks!

src/plugin/featureaction/featureactionmanager.js
```diff
@@ -114,6 +114,14 @@ plugin.im.action.feature.Manager.prototype.initialize = function() {
   for (var i = 0, n = sources.length; i < n; i++) {
     this.addSource_(sources[i]);
   }
+
+  var importActionMan = os.im.action.ImportActionManager.getInstance();
+  var actionEntries = importActionMan.getActionEntries();
+  for (var i = 0, n = actionEntries.length; i < n; i++) {
+    if (dm.getDescriptor(actionEntries[i].getType()) == null) {
```
schmidtk: This block is deleting all feature actions when I refresh the app, even those made on a layer with a descriptor.

lilysikes: This is not what happens for me. It only deletes feature actions on merged or joined layers.

schmidtk: I'm still seeing the behavior. To reproduce:

- Add the [GDP Geoserver](https://gdp-geoserver.dev.dev.east.paas.geointservices.io/geoserver/ows).
- Turn on the VIIRS layer.
- Import [VIIRS Detection FAs.xml.txt](https://github.com/ngageoint/opensphere/files/3216519/VIIRS.Detection.FAs.xml.txt) (drop the `.txt` extension, GitHub won't upload `.xml`).
- Right click the layer > Most Recent, then load some data. You should see purple stars from the FA.
- Reload the app. The FA should be gone.

schmidtk: This happens because the OGC descriptor has aliases (alternate ID's) that get populated when it finishes loading, and ties it to the WMS/WFS layers. When this check runs, the server is still in a loading state so those aliases have not been populated and the descriptor cannot be found. It's also possible the provider is disabled and will not load, so we can't simply wait for all providers to finish loading.

Instead of trying to clean up the entries when restoring the state, we can prevent it from being saved in the first place. The parent class (`os.filter.FilterEntry`) has a temporary flag set with `setTemporary` that will prevent `os.im.action.ImportActionManager#save` from saving the feature action.

I would suggest setting the flag in `os.ui.filter.ui.EditFiltersCtrl#finish`, which will prevent both feature actions and filters from being persisted to storage if the entry does not have a descriptor.

PR #488, opened 2019-04-17
fix(KMLExport): Unused options are removed from the export places.

### LINKED ISSUE #486: Export Places - Remove Unused Options
Export Places is intended to work in a way so that places are reimported with the exact same style (in Opensphere, Google Earth doesn't support our point/square/etc styles). Remove the following extraneous KML options because this is a very specific, tailored export:

- Use Feature Colors
- Export Ellipses
- Default Icon

lilysikes:
Resolves #486

### DIFF
--- a/src/plugin/file/kml/ui/kmltreeexportui.js
+++ b/src/plugin/file/kml/ui/kmltreeexportui.js
@@ -89,11 +89,6 @@ plugin.file.kml.ui.KMLTreeExportCtrl = function($scope, $element) {
   this['title'] = root && root.getLabel() || (os.config.getAppName() + ' KML Tree').trim();
 
   /**
-   * @type {boolean}
-   */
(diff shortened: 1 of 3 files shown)


src/plugin/file/kml/ui/kmlexportui.js
```diff
@@ -17,7 +17,8 @@ plugin.file.kml.ui.kmlExportDirective = function() {
     restrict: 'E',
     scope: {
       'exporter': '=',
-      'simple': '=?'
+      'simple': '=?',
+      'savedplaces': '='
```
schmidtk: Avoid referring to Saved Places anywhere outside the places plugin (`src/plugin/places`). I believe the `simple` attribute is already what you want, but it should also disable the `Use Feature Colors` option.

Issue #381, opened 2019-01-07
Places have interesting altitudes


csgibson:
Possibly related to #183 

Point places that are created by the user through context menu have a random value as altitude.

schmidtk:
`getCoordinateFromPixel` was modified in the Cesium renderer to pick the globe instead of the ellipsoid:
https://github.com/ngageoint/opensphere/blob/22764a6722ba2b320d00f45919595be1982bf4e2/src/plugin/cesium/cesiumrenderer.js#L238-L243

When terrain is enabled, this creates the saved place on the terrain where the user right-clicked, instead of a coordinate "through" the terrain on the surface of the ellipsoid. This was especially noticeable when viewing mountains at an angle. Creating a saved place in a mountainous region could put the coordinate very far from the clicked target.

This does not work well with the default terrain provider (`Cesium.EllipsoidTerrainProvider`), and the altitude you're seeing is likely the difference between the Earth's max radius and the radius of the ellipsoid where the user clicked.

We should fix this by zeroing out altitude in `getCoordinateFromPixel` when terrain is disabled.

swtyler:
Places now have a far less interesting value of 0 when created. I believe this can be closed via https://github.com/ngageoint/opensphere/commit/4ca67c4286e0b668906e7a0dad71012fed12e0fd unless we want to do something more with terrain.

schmidtk:
Further work will require support from Cesium for a non-hacky solution, so I'm fine with closing this.

PR #493, opened 2019-04-22
feat(cesium): support altitude modes via ground primitives

wallw-bits:
This addresses lines/polygons either drawn by the user or from data sources. In addition, `getCoordinateFromPixel` was improved with Cesium's recommendation for handling the terrain vs. ellipsoid cases.

Tests:
General Features
* Test drawing box/circle/polygon in 2D, 3D (no terrain), and 3D (terrain), and all in multiple projections.
* Test loading lines/polys from a file and changing the altitude mode
* Test the case from #381 

Places
* Verify that adding a new place allows you to choose an Altitude Mode and that the preview feature is correctly updated on each change
* Verify that editing the place populates the Altitude Mode dropdown correctly
* Verify that reloading the application displays the geometry correctly
* Verify that exporting the place results in the appropriate `<altitudeMode>` tag being added to the KML
* Verify that importing that KML results in the correct altitude mode
* Add a line and polygon with altitude to places and repeat the above steps

Feature Requests:
- [x] Can places be assigned an altitude if terrain is on in 3D when created via Layers widget (pasting coordinates or clicking on the map)?
- [x] Can non-point places support terrain like polygons? Currently they are flattened.

Discussions:
- [x] Cesium upgrade and new items like `Scene.highDynamicRange`
- [x] Interpolation granularity


wallw-bits:
Note that this does not address Measure Lines. The calculation for those is following the surface of the ellipsoid, and therefore that is where they will be drawn.

We certainly have the capability to do that calculation over terrain should anyone request it.

src/plugin/cesium/cesiumrenderer.js
```diff
@@ -231,13 +231,17 @@ plugin.cesium.CesiumRenderer.prototype.getCoordinateFromPixel = function(pixel)
     var cartesian = Cesium.Cartesian2.fromArray(pixel);
     var scene = this.olCesium_.getCesiumScene();
     if (scene && scene.camera && scene.globe) {
-      var pickRay = scene.camera.getPickRay(cartesian);
-      cartesian = scene.globe.pick(pickRay, scene);
+      // The Cesium team recommends different methods here based on whether terrain is enabled
+      // see https://github.com/AnalyticalGraphicsInc/cesium/issues/4368#issuecomment-406639086
+      if (scene.terrainProvider) {
```
schmidtk: `Cesium.EllipsoidTerrainProvider` is used when terrain is disabled, so this condition will always be true.

src/os/mapcontainer.js
```diff
@@ -880,6 +880,7 @@ os.MapContainer.prototype.init = function() {
   this.drawingLayer_.setSticky(true);
   this.drawingLayer_.setSynchronizerType(os.layer.SynchronizerType.DRAW);
   this.drawingLayer_.renderLegend = goog.nullFunction;
+  this.drawingLayer_.set('altitudeMode', 'clampToGround');
```
schmidtk: Use constants here?

src/plugin/cesium/sync/featureconverter.js
```diff
@@ -1146,34 +1165,19 @@ plugin.cesium.sync.FeatureConverter.prototype.setAltitudeMode = function(altitud
  * @return {!Cesium.HeightReference}
  */
 plugin.cesium.sync.FeatureConverter.prototype.getHeightReference = function(layer, feature, geometry) {
-  // disable height reference because the implementation is fairly slow right now
-  // TODO: Should we remove this since with the function above we are seting it for the whole layer?
-  return this.heightReference_;
+  var altitudeMode = geometry.get('altitudeMode') || feature.get('altitudeMode') || layer.get('altitudeMode');
```
schmidtk: Use `os.data.RecordField.ALTITUDE_MODE`?

package.json
```diff
@@ -264,7 +264,7 @@
     "blob-polyfill": "^2.0.20171115",
     "bootstrap": "4.3.1",
     "bootswatch": "4.3.1",
-    "cesium": "1.49.0",
+    "cesium": "1.56.1",
```
schmidtk: The Cesium upgrade has some additions like `Scene.highDynamicRange` that change how the globe is rendered fairly noticeably. I'm wondering if we should figure out what all changed and adjust to look more like it did previously. See differences here:

Screen Shot 2019-04-23 at 8 43 49 AM

justin-bits:
- Places have a negative altitude when created from the context menu with terrain off in 3D
negative altitude

- Can places be assigned an altitude if terrain is on in 3D when created via Layers widget (pasting coordinates or clicking on the map)?
- Console error thrown on attempting two draw box, 3D.  Map locks up (won’t pan).
console error on drawing box 3d

- Line for polygon disappears depending on orientation
polygon line disappears

schmidtk:
@justin-bits 
> Console error thrown on attempting two draw box, 3D. Map locks up (won’t pan).

Did you do a `yarn upgrade`? Cesium was updated in this PR, and it looks like you're on the previous version. That would cause the error you mentioned.

justin-bits:
Altitude update (1) after Will's fix: 

Altitude is non zero with terrain off after mode change. On 3d terrain off>2d>3d save place it shows altitude on saving the place (slightly positive to slightly negative). Not 100% consistent. 4.65 meters comes up frequently.

image

justin-bits:
Doing `yarn upgrade`, will update.

justin-bits:
1. Still valid (initially negative altitude, now 4.65)
2. Outstanding question/request.
3. Resolved by `yarn upgrade`.
4. Still valid.

wallw-bits:
1. I'm concerned that Cesium might have a bug there. Both the forums and the docs indicate that the value returned by `Camera.prototype.pickEllipsoid()` should have a height of zero.
1. Sure.
1. w00t 
1. This is due to our interpolation granularity defaulting to 100,000m while Cesium's internal granularity (which we have turned off) defaults to 9,999m. If you change the value in Settings > Interpolation from 100km to 10km that fixes it. However, that's 10x the number of vertices per area.

justin-bits:
6. Console errors thrown on buffering a line (3d, terrain on/off).
```
Uncaught TypeError: Cannot read property 'x' of undefined
    at v.setBatchedAttribute (Cesium.js:471)
    at Y (Cesium.js:471)
    at H.update (Cesium.js:472)
    at M.update (Cesium.js:474)
    at a.update (Cesium.js:483)
    at a.update (Cesium.js:483)
    at a.update (Cesium.js:483)
    at at (Cesium.js:522)
    at it (Cesium.js:522)
    at $e (Cesium.js:522)
```

wallw-bits:
That is the same thing that Kevin mentioned. It has to do with polygon fills and I'm looking into it.

wallw-bits:
For the interpolation granularity (affecting clipping through the earth), we could simply enable Cesium's interpolation in addition to ours. In the event that ours is at the default 100km or greater, the polygon on our side will keep the same number of vertices and only the Cesium primitive will have more at their default of 9.999km. This does appear to be a little chunky (slow) in my tests though (largely because drawn items are rendered with `async: false` and are updated quickly).

wallw-bits:
Just don't forget to put Settings > Interpolation > Granularity back to 100km.

justin-bits:
- Console error on buffering resolved
- Save places altitude working correctly
- Still getting clipping for lines/polygons, though it appears to be significantly better

wallw-bits:
Setting these items appears to fix the globe rendering issues.
```
scene.highDynamicRange = false;
scene.globe.showGroundAtmosphere = false;
```

schmidtk:
Line Interpolation

Geodesic (left) interpolation still clips the globe noticeably and I don't think our users will like that. Can we tune that a bit more so it's comparable to master?

I would also be in favor of disabling `highDynamicRange`, because it makes the scene generally appear dim compared to master and the 2D view. Does enabling it provide any benefits?

wallw-bits:
Here's the Cesium forum post where this is being discussed:

https://groups.google.com/forum/#!topic/cesium-dev/UvLaAr8VnRA

wallw-bits:
And here's the associated issue: AnalyticalGraphicsInc/cesium#7855

wallw-bits:
That issue has an open PR that fixes the problem. I'm going to proceed to the feature requests.

wallw-bits:
The Cesium PR was merged and will make their July 1 release.

wallw-bits:
Also, just for historical purposes: the line was not "clipping" through the earth. The line miter normal was not pointing in the correct direction, resulting in it not being rendered at all. That value is now guaranteed to be tangent to the ellipsoid surface.

wallw-bits:
I think this is ready to go other than deciding whether or not `scene.highDynamicRange` should be on or off.

wallw-bits:
I think all of those issues are fixed. I'll note that Cesium does not actually support `relativeToGround` for lines or polygons. It is picking either `GroundPrimitive` (`clampToGround`) or `Primitive` for polygons, and `GroundPolylinePrimitive/Geometry` or `Primitive/PolylineGeometry` for lines. There was no option similar to `heightReference` for `BillBoard`s. `absolute` and `relativeToGround` are displayed the same in those cases.

src/os/ui/featureedit.js
```diff
@@ -1227,28 +1298,59 @@ os.ui.FeatureEditCtrl.prototype.saveGeometry_ = function(feature) {
...
-        feature.set(os.Fields.BEARING, undefined);
-        feature.set(os.style.StyleField.SHOW_ROTATION, false);
-        feature.set(os.style.StyleField.ROTATION_COLUMN, '');
+        feature.set(os.Fields.BEARING, undefined, true);
+        feature.set(os.style.StyleField.SHOW_ROTATION, false, true);
+        feature.set(os.style.StyleField.ROTATION_COLUMN, '', true);
       }
     }
-  } else if (this.originalGeometry) {
+  } else if (this.originalGeometry && feature.getGeometry() === this.originalGeometry) {
```
schmidtk: This change is preventing non-point geometries from displaying on the map when creating a new place. The preview feature geometry is never set, so the second condition is always false.

justin-bits:
@wallw-bits 

1. Error on creating places (with defaults) under the projections 3031 and 3413.  These places shift around the globe on changing projections.  On removal via the layers dialog, they also remain on the map.

2. Box/circle/polygons shift around on the map when changing projections.  Even when returning to the projection in which they were created, they are in dramatically different positions at different sizes.

3. Units on Altitude field do not respect user settings, it defaults to meters.  Also, precision is unreasonably high.

4. On opening the Add Place dialog, the place is not visible.  It only becomes visible on clicking OK.

wallw-bits:
1. That is unrelated to this ticket
2. I'm not able to reproduce that
3. That is unrelated to this ticket
4. I'll fix that

PR #496, opened 2019-04-24
feat(columnAssociation): os454 column association

### LINKED ISSUE #454: Column Associations form validation and descriptor clean up issues
- OK button disabled even though the form is filled out correctly.  Server: https://gdp-geoserver.dev.dev.east.paas.geointservices.io/geoserver/ows
ok disabled

- Form incorrectly indicating layers are duplicated.  Server: https://ags.auroragov.org/aurora/rest/services/OpenData/MapServer
duplicate

- Layers from inactive or removed servers still present.
old descriptors

robertbolan-bits:
Fixes #454 
Fix a 3 issue in the mapping associations. 
First issue fix: change from using item.layer to item['layer'].
Second issue fix: Created a new interface for arc descriptor and implemented on the arclayerdescriptor. 
Third issue fix: Replaced a Try/Catch with an if statement also checked to make sure each descriptor implements featureType. Also when removing a provider you also removed all descriptors associated with that providers.


src/os/ui/servers.js
```diff
@@ -369,7 +369,13 @@ os.ui.ServersCtrl.prototype.remove = function(provider, opt_prompt) {
...
 
   os.settings.delete(['userProviders', provider.getId()]);
-
+  var descList = os.dataManager.getDescriptors();
+  for (var i = 0, ii = descList.length; i < ii; i++) {
+    var desc = descList[i];
+    if (desc.getDataProvider() == provider) {
+      os.dataManager.removeDescriptor(desc);
+    }
+  }
```
wallw-bits: This block is incorrect. The value provided by `descriptor.getProvider()` is only intended for display. Rather, you should remove descriptors via ID like so:

```
var descList = os.dataManager.getDescriptors(provider.getId() + os.data.BaseProvider.ID_DELIMITER);
for (var i = 0, ii = descList.length; i < ii; i++) {
  os.dataManager.removeDescriptor(descList[i]);
}
```

src/os/ui/column/mapping/columnmappingform.js
```diff
@@ -215,7 +213,7 @@ os.ui.column.mapping.ColumnMappingFormCtrl.prototype.validateLayers_ = function(
   }
 
   var incompleteLayer = ol.array.find(columns, function(item) {
-    return (!item.layer || item.layer.length == 0 || !item.column || item.column.length == 0);
+    return (!item['layer'] || item['layer'].length == 0 || !item['column'] || item['column'].length == 0);
```
wallw-bits: Why did you do this? `item` is a `os.column.ColumnModel`, which is a typedef defined in `os.column.IColumnMapping` and should not require bracket notation.

robertbolan-bits: This solution was to fix the first screenshot. When it wasn't in bracket notation the  incompleteLayer would always be set to false after building it. I assume closure didnt like something.

wallw-bits: Hmm. If `os.column.ColumnModel` is being displayed in an Angular template, that might explain it. However, we should track that down. If the names need to be preserved, we can move the typedefs to externs rather than switching to bracket notation and that will prevent their minification.

welchyd: Other parts of the file use bracket notation for the column model too
If we change the typedef to an extern, these will need to be updated
https://github.com/ngageoint/opensphere/blob/master/src/os/ui/column/mapping/columnmappingform.js#L147
https://github.com/ngageoint/opensphere/blob/master/src/os/ui/column/mapping/columnmappingform.js#L203
https://github.com/ngageoint/opensphere/blob/master/src/os/ui/column/mapping/columnmappingform.js#L211
https://github.com/ngageoint/opensphere/blob/master/src/os/ui/column/mapping/columnmappingform.js#L259

robertbolan-bits: @wallw-bits I updated this to so the typedef is now an extern. Is there anything else I should look at or can this change be closed?

wallw-bits: If the type definition is in the externs then you should no longer need the bracket notation. Once that is fixed here and in the places Doug mentioned, then I'll have a final look at it.

welchyd: Looks good

wallw-bits:
This looks good.

schmidtk:
Column associations on Arc layers aren't being picked up by the Merge operation because the ID's used aren't the same. For example, if I create an association with the Fire Stations layer on the AuroraGov server I see the following ID's:

ColumnModel layer: `https://ags.auroragov.org/aurora/rest/services/OpenData/MapServer!!Fire Stations`
Filter key: `https://ags.auroragov.org/aurora/rest/services/OpenData/MapServer/5/query!!Fire Stations`

The issue seems to be in using `getUrlKey` for the former, and `getFilterKey` for the latter. I'm not familiar with the background on the two calls. @wallw-bits?

src/plugin/arc/layer/arclayerdescriptor.js
```diff
@@ -554,3 +561,38 @@ plugin.arc.layer.ArcLayerDescriptor.prototype.getNodeUI = function() {
 plugin.arc.layer.ArcLayerDescriptor.prototype.getFilterableTypes = function() {
   return [this.getId() + os.ui.data.BaseProvider.ID_DELIMITER + 'features'];
 };
+
+/**
+ * @inheritDoc
+ */
+plugin.arc.layer.ArcLayerDescriptor.prototype.getUrlKey = function() {
+  return this.url_ + '!!' + this.getTitle();
```
wallw-bits: Kevin is correct. This should `return this.getFilterKey();`

This is silly and we should consolidate those, but I'll write that up separately.

schmidtk:
FYI - while testing I noticed some missing columns in the column mapping form. It was unrelated to these changes, so I submitted a fix in #546.

src/os/ui/arc/iarcdescriptor.js
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * @type {string}
+ * @const
+ */
+os.ui.arc.IARCDescriptor.ID = 'os.ui.arc.IARCDescriptor';
+
+/**
+ * @return {?string}
+ */
+os.ui.arc.IARCDescriptor.prototype.getUrlKey;
```
bradh: Can we please not add this to the API? See https://github.com/ngageoint/opensphere/issues/545

src/plugin/arc/layer/arclayerdescriptor.js
```diff
@@ -554,3 +561,38 @@ plugin.arc.layer.ArcLayerDescriptor.prototype.getNodeUI = function() {
 plugin.arc.layer.ArcLayerDescriptor.prototype.getFilterableTypes = function() {
   return [this.getId() + os.ui.data.BaseProvider.ID_DELIMITER + 'features'];
 };
+
+/**
+ * @inheritDoc
+ */
+plugin.arc.layer.ArcLayerDescriptor.prototype.getUrlKey = function() {
```
bradh: And not add this to the implementation. See https://github.com/ngageoint/opensphere/issues/545

test/os/ui/column/mapping/columnmappingform.test.js
```diff
@@ -0,0 +1,229 @@
...
+
+describe('os.ui.column.mapping.ColumnMappingFormCtrl', function() {
+  var $scope, cmForm, element;
+  var mapping = new os.column.ColumnMapping();
+
+  var ogclayerMock = new plugin.ogc.OGCLayerDescriptor;
+  var providerMock = new plugin.ogc.GeoServer;
+  var ogclayerMock2 = new plugin.ogc.OGCLayerDescriptor;
+  var providerMock2 = new plugin.ogc.GeoServer;
+  // var dataManager = new os.dataManager;
```
bradh: Possibly remove commented out code.

test/os/ui/column/mapping/columnmappingform.test.js
```diff
@@ -0,0 +1,229 @@
...
+  beforeEach(function() {
+    spyOn(providerMock, 'getEnabled').andReturn(true);
+    spyOn(ogclayerMock, 'getDataProvider').andReturn(providerMock);
+    spyOn(os.dataManager, 'getDescriptors').andReturn(descriptorList);
+    spyOn(providerMock2, 'getEnabled').andReturn(false);
+    spyOn(ogclayerMock2, 'getDataProvider').andReturn(providerMock2);
+    inject(function($compile, $rootScope) {
+      $scope = $rootScope;
+      // windowScope = $rootScope.$new();
+      // ctrlScope = windowScope.$new();
```
bradh: Remove commented out code

test/os/ui/column/mapping/columnmappingform.test.js
```diff
@@ -0,0 +1,229 @@
...
+  it('should return the all descriptros in the descriptor list', function() {
+    descriptorList.push(ogclayerMock);
+    descriptorList.push(ogclayerMock);
+    var formCtrl = new os.ui.column.mapping.ColumnMappingFormCtrl($scope, element, timeout);
+
+    var actuallist = formCtrl.getLayersFunction();
+    expect(actuallist.length).toBe(2);
+  });
+
+  it('should not add a decriptor type not feature type', function() {
```
bradh: typo: descriptor

src/os/ui/arc/iarcdescriptor.js
wallw-bits: I was planning on merging that after this, but it looks like Kevin already merged it, so you're correct. This should not be added.

src/plugin/arc/layer/arclayerdescriptor.js
wallw-bits: Ditto to interface comment.

src/os/ui/arc/iarcdescriptor.js
schmidtk: Sorry I should have checked first. I wanted to avoid adding this, then removing it again later.

robertbolan-bits: I have removed it in the most recent push. I think it got added to the merge commit

src/os/ui/column/mapping/columnmappingform.js
```diff
@@ -127,14 +127,12 @@ os.ui.column.mapping.ColumnMappingFormCtrl.prototype.init_ = function() {
 
   for (var i = 0, ii = descList.length; i < ii; i++) {
     var desc = descList[i];
-    try {
-      desc = /** @type {os.ui.ogc.IOGCDescriptor} */ (desc);
-      if (desc.isWfsEnabled() === true) {
-        descMap[desc.getFilterKey()] = desc;
+
+    if (desc.getDataProvider().getEnabled()) {
+      if (os.implements(desc, os.ui.ogc.IFeatureTypeDescriptor.ID)) {
```
wallw-bits: I thought we were going to be using `os.filter.IFilterable` for this?

src/os/ui/column/mapping/mappingexpression.js
```diff
@@ -130,7 +130,7 @@ os.ui.column.mapping.MappingExpressionCtrl.prototype.onColumnChange_ = function(
 
 /**
  * Sets the layer and gets the columns for it.
- * @param {os.ui.ogc.IOGCDescriptor} layer
+ * @param {os.ui.ogc.IFeatureTypeDescriptor} layer
```
wallw-bits: Same here. `IFilterable`?

src/os/ui/column/mapping/columnmappingform.js
```diff
@@ -127,14 +127,14 @@ os.ui.column.mapping.ColumnMappingFormCtrl.prototype.init_ = function() {
...
-    try {
-      desc = /** @type {os.ui.ogc.IOGCDescriptor} */ (desc);
-      if (desc.isWfsEnabled() === true) {
-        descMap[desc.getFilterKey()] = desc;
-        this.cachedDescriptorList_.push(desc);
+
+    if (desc.getDataProvider()) {
+      if (desc.getDataProvider().getEnabled()) {
+        if (os.implements(desc, os.ui.ogc.IFeatureTypeDescriptor.ID)) {
+          descMap[desc.getFilterKey()] = desc;
```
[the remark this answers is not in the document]
schmidtk: I don't think the compiler will complain if it determines every possible type for `desc` implements `getFilterKey`.

PR #500, opened 2019-04-25
build(cypress): Always exits code 0, even when tests fail

justin-bits:


wallw-bits:
We'll do this as long as it is temporary and reverted before the `travis-artifacts` branch is merged.

Issue #495, opened 2019-04-24
Hide Annotation Button


welchyd:
In annotations - add an icon next the the edit button to allow a user to quickly hide an annotation. This should effectively uncheck the annotation node in the layers window when clicked.

<img width="1181" alt="Screen Shot 2019-04-24 at 12 16 43 PM" src="https://user-images.githubusercontent.com/35077650/56683671-44119100-668b-11e9-8786-ea9c3fcb663b.png">

THIN-13365
1 story point

lilysikes:
Fixed via #502

mgettman:
looks good

Issue #353, opened 2018-12-12
Fix/Remove Annotations on Tracks


welchyd:
Tracks have the option to show annotations, but nothing shows up when the option is turned on.
![trackannotations](https://user-images.githubusercontent.com/35077650/49885530-b11b3d00-fdf4-11e8-936e-cb17dfd78967.png)

robertbolan-bits:
I will look into this one

PR #502, opened 2019-04-29
feat(HideAnnotation): Hide button added to annotations to make them q…

lilysikes:
…uickly removeable.

#495

### DIFF
--- a/src/plugin/file/kml/ui/kmlnode.js
+++ b/src/plugin/file/kml/ui/kmlnode.js
@@ -256,6 +256,9 @@ plugin.file.kml.ui.KMLNode.prototype.loadAnnotation = function() {
         this.feature_.get(os.annotation.OPTIONS_FIELD));
     if (annotationOptions && annotationOptions.show) {
       this.annotation_ = new os.annotation.FeatureAnnotation(this.feature_);
+
+      // set initial visibility based on the tree/animation state
+      this.setAnnotationVisibility_(this.getState() === os.structs.TriState.ON && this.animationState_);
     }
   }
 };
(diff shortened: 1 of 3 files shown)


welchyd:
Annotations don't currently show on geometries (also happens on tracks - #353), so for now the checkbox represents the shown state of the annotation - we need to decide on if the annotation should be tied to the geometry or a separate entity.
I'm in favor of combining them - have the annotation be a label or the floating text and in that case I could see having an icon on the note to show and even toggle the visibility of the annotation itself.

schmidtk:
KML nodes that support showing an annotation should have a show/hide annotation button in their Layers window UI to toggle the annotation on/off.

That button should control the `show` flag in `os.annotation.OPTIONS_FIELD` on the feature. We _should not_ be exposing the annotation via a `getAnnotation` function to call `setVisible` because the actual annotation visibility is managed by multiple things (node visibility, annotation show flag, timeline) and that should all be controlled within the node itself.

When the `show` flag is changed on the feature's annotation options, the node should update the annotation by either disposing the existing annotation or creating a new one. We should update `plugin.file.kml.ui.KMLNode#loadAnnotation` to handle both of these cases (it currently only handles creating the annotation if one doesn't exist and the flag is true).

src/plugin/file/kml/ui/kmlnode.js
```diff
@@ -893,3 +893,10 @@ plugin.file.kml.ui.KMLNode.prototype.getLayerUI = function(item) {
 
   return null;
 };
+
+/**
+ * @return {os.annotation.FeatureAnnotation}
+ */
+plugin.file.kml.ui.KMLNode.prototype.getAnnotation = function() {
```
schmidtk: Please remove this function. The annotation should not be exposed outside the KML node. This is explained further in my comment in the issue.

PR #515, opened 2019-05-02
WIP: Test/heatmap

justin-bits:
- Need this to build to grab artifacts, build expected to fail
- S3KEY and S3SECRET must be set as environment variables, DONE
- https://github.com/justin-bits/opensphere/blob/test/heatmap/cypress/support/s3-artifact-upload.sh must be configured in Travis to run on build failures, DONE
- This is a different approach than before, therefore doesn't use the built-in artifact uploader and doesn't need to be a direct push to https://github.com/ngageoint/opensphere/tree/travis-artifacts
- Artifacts will output here: https://opensphere-travis-artifacts.s3.amazonaws.com/
- S3 bucket is setup to expire artifacts after one day
- All artifacts grouped by build number


justin-bits:
This won't work: https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions

```
Pull requests sent from forked repositories do not have access to encrypted variables or data even if these are defined in the fork source project.
```

Closed.

PR #527, opened 2019-05-08
fix(timeline): Change timeline view window alert to be info level.

### LINKED ISSUE #466: Change timeline error to info
When you have the timeline open with the lock on and then move the view window, an error pops up letting you know this unlocks the timeline.
Having an error message leaves me feeling like I did something wrong when it should be fine.
Let's change the message to be information rather than an error
Screen Shot 2019-04-04 at 9 18 08 AM

bradh:
Resolves #466.


bradh:
alert info

PR #543, opened 2019-05-09
test(cypress): Adds image comparison, heatmap test, artifact upload, …

justin-bits:
…single spec execution


justin-bits:
features_and_heatmap_removed diff

Current implementation has rendering differences between local builds and CI.  I'm not sure this can be fixed, but if there are ideas let me know.

schmidtk:
These changes are also in #549, so closing in favor of that PR.

PR #547, opened 2019-05-10
refactor: replace goog.abstractMethod with @abstract

### LINKED ISSUE #507: Replace goog.abstractMethod with @abstract
Replace all use of `goog.abstractMethod` with the `@abstract` JSDoc tag. Closure [added support for this](https://github.com/google/closure-compiler/wiki/@abstract-classes-and-methods) in 2016, and it is considered the preferred method of noting abstract classes/methods.

wallw-bits:
BREAKING CHANGE: This allows the compiler to do proper checks on
missing implementation of abstract classes.

Resolves #507

### DIFF
--- a/src/os/webgl/abstractwebglrenderer.js
+++ b/src/os/webgl/abstractwebglrenderer.js
@@ -183,7 +183,7 @@ os.webgl.AbstractWebGLRenderer.prototype.forEachFeatureAtPixel = function(pixel,
  * @abstract
  * @inheritDoc
  */
-os.webgl.AbstractWebGLRenderer.prototype.toggleMovement = function() {};
+os.webgl.AbstractWebGLRenderer.prototype.toggleMovement = function(value) {};
 
 
 /**
--- a/src/os/webgl/abstractwebglrenderer.js
+++ b/src/os/webgl/abstractwebglrenderer.js
@@ -9,6 +9,7 @@ goog.require('os.webgl.IWebGLRenderer');
 
 /**
  * Abstract WebGL renderer implementation.
+ * @abstract
  * @implements {os.webgl.IWebGLRenderer}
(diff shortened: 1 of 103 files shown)


src/os/webgl/abstractwebglrenderer.js
```diff
@@ -158,27 +159,31 @@ os.webgl.AbstractWebGLRenderer.prototype.resetSync = function() {
...
-os.webgl.AbstractWebGLRenderer.prototype.forEachFeatureAtPixel = goog.abstractMethod;
+os.webgl.AbstractWebGLRenderer.prototype.forEachFeatureAtPixel = function(pixel, callback) {};
 
 
 /**
+ * @abstract
  * @inheritDoc
  */
-os.webgl.AbstractWebGLRenderer.prototype.toggleMovement = goog.abstractMethod;
+os.webgl.AbstractWebGLRenderer.prototype.toggleMovement = function() {};
```
schmidtk: This function is missing the `value` parameter from the interface.

src/plugin/cesium/cesiumrenderer.js
```diff
@@ -377,7 +377,7 @@ plugin.cesium.CesiumRenderer.prototype.toggleMovement = function(value) {
   if (this.olCesium_) {
     var scene = this.olCesium_.getCesiumScene();
     if (scene && scene.screenSpaceCameraController) {
-      scene.screenSpaceCameraController.enableInputs = value;
+      scene.screenSpaceCameraController.enableInputs = !!value;
```
schmidtk: `value` is typed as `boolean`, so this seems to be a superfluous addition. It looks like the missing param in my previous comment is causing the compiler to think the type is `boolean|undefined`.

PR #550, opened 2019-05-10
Create separate functions to get original columns or a copy.

### LINKED ISSUE #548: Vector sources should return a copy of their columns
`os.source.Vector#getColumns` currently returns the original array of columns for the source. This can be confusing to consumers of the columns, and has led to a few cases where the original columns array was unintentionally modified.

Create separate functions, `getColumns` (returns a copy) and `getColumnsArray` (returns the original. Update all calls to `getColumns` to use the appropriate function, based on if the original array needs to be modified.

schmidtk:
This prefers `getColumnsArray` for performance when the source columns are not being modified. Otherwise use `getColumnsArray` when intentionally modifying columns, and `getColumns` when modifying for local use.

Fixes #548.

### DIFF
--- a/src/os/source/isource.js
+++ b/src/os/source/isource.js
@@ -88,7 +88,7 @@ os.source.ISource.prototype.getColumns;
 
 
 /**
- * Get the origial array of columns for the data source.
+ * Get the original array of columns for the data source.
  * @return {!Array<!os.data.ColumnDefinition>}
  */
 os.source.ISource.prototype.getColumnsArray;
--- a/src/os/source/isource.js
+++ b/src/os/source/isource.js
@@ -81,10 +81,17 @@ os.source.ISource.prototype.getColor;
 
 
 /**
- * Get the array of data grid columns for the data source.
+ * Get a copy of the columns for the data source.
  * @return {!Array<!os.data.ColumnDefinition>}
  */
 os.source.ISource.prototype.getColumns;
+
+
+/**
+ * Get the origial array of columns for the data source.
+ * @return {!Array<!os.data.ColumnDefinition>}
+ */
+os.source.ISource.prototype.getColumnsArray;
 
 
 /**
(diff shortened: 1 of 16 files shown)


bradh:
I think I see the logic, but the commit message is confusing me a bit.

To check if I understand it:
 * If I'm creating a read-only view (e.g. a heatmap or graph) I should use `getColumnsArray()`?
 * If I'm trying to edit the original data (e.g. a spreadsheet-style editor), I should use `getColumnsArray()`?
 * If I need to modify the copy, but not the original (e.g. an export that needs to project coordinates to WGS-84 before writing them), I should use `getColumns()`?

src/os/source/isource.js
```diff
@@ -81,12 +81,19 @@ os.source.ISource.prototype.getColor;
...
 /**
- * Get the array of data grid columns for the data source.
+ * Get a copy of the columns for the data source.
  * @return {!Array<!os.data.ColumnDefinition>}
  */
 os.source.ISource.prototype.getColumns;
 
 
+/**
+ * Get the origial array of columns for the data source.
```
bradh: typo: original

schmidtk: Thanks!

schmidtk:
@bradh These calls get the column definitions (name, field, data type, etc), not actual data from features. `getColumns` previously returned the original array of columns, and we had a few instances where developers did not realize this and sorted the columns for display in a UI. The stored column order should generally not be modified unless driven by a user action, because UI's (a tabular list, for example) may rely on that order.

Going forward, `getColumns` is preferred unless:
1. The column order stored on the source needs to be modified (this can also be achieved with `setColumns`).
2. The caller is not modifying the array, and wants to avoid the additional overhead of copying the array.

I preferred `getColumnsArray` in the PR to avoid changing existing behavior, except for cases where the list was being sorted for display.

PR #551, opened 2019-05-11
chore(api): Deprecate IOGCDescriptor::getUrlKey(), use getFilterKey()

### LINKED ISSUE #545: Remove getUrlKey
`getUrlKey` (defined on both `IOGCDescriptor` and `IArcDescriptor`) is the same as `getFilterKey` (defined on `IFilterable`).

Ensure the implementation is on `getFilterKey` and then deprecate the other function.

bradh:
IOGCDescriptor extends IFilterable, so its better to just implement in terms
of the parent API.

Resolves #545.

### DIFF
--- a/src/os/ui/ogc/iogcdescriptor.js
+++ b/src/os/ui/ogc/iogcdescriptor.js
@@ -234,6 +234,7 @@ os.ui.ogc.IOGCDescriptor.prototype.getLayerName;
 
 /**
  * @return {?string}
+ * @deprecated Use IFilterable.getFilterKey()
  */
 os.ui.ogc.IOGCDescriptor.prototype.getUrlKey;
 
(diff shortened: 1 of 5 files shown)


src/os/ui/ogc/iogcdescriptor.js
```diff
@@ -234,6 +234,7 @@ os.ui.ogc.IOGCDescriptor.prototype.getLayerName;
 
 /**
  * @return {?string}
+ * @deprecated Use IFilterable::getFilterKey()
```
wallw-bits: We don't have anything else that references methods in this manner. For consistency, I'd prefer something like `IFilterable.getFilterKey()` instead.

bradh: Updated.

PR #556, opened 2019-05-14
Add active area/filter counts to Layers window tabs

### LINKED ISSUE #555: Display active area/filter counts in Layers window tabs
As a user, it would be helpful to have quick reference to what is impacting server queries. Areas/filters are a key part of that, so it would be nice to know how many are active at a glance. Add badges to the Areas/Filters tabs that show a count of how many are currently active. On hover, display either a tooltip or popover with an explanation of the count and possibly additional information.

schmidtk:
Resolves #555.

### DIFF
--- a/cypress/support/selectors.js
+++ b/cypress/support/selectors.js
@@ -2886,7 +2886,7 @@ exports.statusBar = {
   LEGEND_BUTTON: '[title=\'View Legend\']',
   SERVERS_BUTTON: '[title=Servers]',
   ALERTS_BUTTON: '[title=Alerts]',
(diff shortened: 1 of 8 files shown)


schmidtk:
Added simple badges that show the number of active areas (inclusion/exclusion/both), and active filters. Here's a screenshot of the result:
Screen Shot 2019-05-14 at 3 47 47 PM

I explored putting more details in a popover, but it felt unnecessary and also inconsistent. We currently only use popovers for inline help.

bradh:
Looks like you have a transient failure on the arcgis cypress test (see that locally sometimes). May need a close/reopen to trigger a rebuild.

Issue #533, opened 2019-05-08
Places Folder Option


welchyd:
When Adding/Editing a Place, the user should be able to pick a folder to save it to, similar to being able to pick a folder when favoriting a layer
THIN-13363
2 story points

smithtb:
Closed via #558

PR #558, opened 2019-05-15
feat(places): Add option to choose folder in places add/edit menu

smithtb:
re #533

### DIFF
--- a/src/os/structs/structs.js
+++ b/src/os/structs/structs.js
@@ -1,0 +1,21 @@
+goog.provide('os.structs');
+
+
+/**
+ * Flattens a tree into a specified array - optionally filtering.
+ * @param {os.structs.ITreeNode} node
+ * @param {Array<os.structs.ITreeNode>} results
+ * @param {function(os.structs.ITreeNode): boolean=} opt_filter
+ */
+os.structs.flattenTree = function(node, results, opt_filter) {
+  opt_filter = opt_filter || goog.functions.TRUE;
+  if (opt_filter(node)) {
+    results.push(node);
+  }
+  var children = node.getChildren();
+  if (children) {
+    for (var i = 0, n = children.length; i < n; i++) {
+      os.structs.flattenTree(children[i], results, opt_filter);
+    }
+  }
+};
--- a/src/plugin/file/kml/ui/kmlui.js
+++ b/src/plugin/file/kml/ui/kmlui.js
@@ -2,10 +2,12 @@ goog.provide('plugin.file.kml.ui');
 
 goog.require('goog.asserts');
 goog.require('goog.events.Event');
+goog.require('os.command.SequenceCommand');
 goog.require('os.object');
 goog.require('os.style');
 goog.require('plugin.file.kml.KMLField');
 goog.require('plugin.file.kml.cmd.KMLNodeAdd');
+goog.require('plugin.file.kml.cmd.KMLNodeRemove');
 goog.require('plugin.file.kml.kmlNodeLayerUIDirective');
 goog.require('plugin.file.kml.ui.KMLNode');
 
@@ -163,8 +165,14 @@ plugin.file.kml.ui.updatePlacemark = function(options) {
   // add the placemark to a parent if provided
   if (options['parent']) {
     var parent = plugin.file.kml.ui.verifyNodeInTree_(/** @type {!plugin.file.kml.ui.KMLNode} */ (options['parent']));
-    var cmd = new plugin.file.kml.cmd.KMLNodeAdd(placemark, parent);
-    os.commandStack.addCommand(cmd);
+    var currentParent = placemark.getParent();
+    var cmds = [new plugin.file.kml.cmd.KMLNodeAdd(placemark, parent)];
+    if (currentParent) {
+      cmds.push(new plugin.file.kml.cmd.KMLNodeRemove(placemark));
+    }
+    var sequence = new os.command.SequenceCommand();
+    sequence.setCommands(cmds);
+    os.commandStack.addCommand(sequence);
   }
 
   return placemark;
--- a/src/plugin/file/kml/ui/placemarkedit.js
+++ b/src/plugin/file/kml/ui/placemarkedit.js
@@ -6,11 +6,13 @@ goog.require('os.annotation');
 goog.require('os.annotation.FeatureAnnotation');
 goog.require('os.annotation.annotationOptionsDirective');
 goog.require('os.data.ColumnDefinition');
+goog.require('os.structs');
 goog.require('os.ui.FeatureEditCtrl');
 goog.require('os.ui.Module');
 goog.require('os.ui.list');
 goog.require('plugin.file.kml.KMLField');
 goog.require('plugin.file.kml.ui');
+goog.require('plugin.places.PlacesManager');
 
 
 /**
@@ -45,6 +47,8 @@ plugin.file.kml.ui.PlacemarkEditCtrl = function($scope, $element, $timeout) {
 
   // if the name wasn't set in the base class, set it to New Place
   this['name'] = this['name'] || 'New Place';
+
+  this['customFoldersEnabled'] = false;
 
   // by default, show column choices for the default KML source. remove internal columns because they're not generally
   // useful to a user.
@@ -95,6 +99,21 @@ plugin.file.kml.ui.PlacemarkEditCtrl = function($scope, $element, $timeout) {
       // if creating a new annotation, expand the Annotation Options section by default
       this.defaultExpandedOptionsId = 'featureAnnotation' + this['uid'];
     }
+
+    this['customFoldersEnabled'] = true;
+
+    var folders = [];
+    var rootFolder = plugin.places.PlacesManager.getInstance().getPlacesRoot();
+    os.structs.flattenTree(rootFolder, folders, function(node) {
+      return /** @type {plugin.file.kml.ui.KMLNode} */ (node).isFolder();
+    });
+    this['folderOptions'] = folders;
+
+    var parentFolder = this.options['parent'];
+    if (!parentFolder && this.options['node']) {
+      parentFolder = this.options['node'].getParent();
+    }
+    this['folder'] = parentFolder || rootFolder;
   }
 };
 goog.inherits(plugin.file.kml.ui.PlacemarkEditCtrl, os.ui.FeatureEditCtrl);
@@ -220,6 +239,17 @@ plugin.file.kml.ui.PlacemarkEditCtrl.prototype.saveToFeature = function(feature)
 plugin.file.kml.ui.PlacemarkEditCtrl.prototype.updatePreview = function() {
   plugin.file.kml.ui.PlacemarkEditCtrl.base(this, 'updatePreview');
   this.updateAnnotation();
+};
+
+
+/**
+ * Updates the option for the placemark's parent folder
+ * @export
+ */
+plugin.file.kml.ui.PlacemarkEditCtrl.prototype.updateFolder = function() {
+  if (this['folder']) {
+    this.options['parent'] = this['folder'];
+  }
 };
 
 
--- a/views/windows/featureedit.html
+++ b/views/windows/featureedit.html
@@ -10,18 +10,29 @@
         </div>
       </div>
 
-      <div class="d-flex flex-row form-group">
+      <div class="d-flex flex-row form-group mt-2">
         <label class="col-2 col-form-label text-right">Description</label>
-        <div class="col mt-2">
+        <div class="col">
           <simplemde text="ctrl.description" edit="true" is-required="false" maxlength="4000"></simplemde>
         </div>
       </div>
+
+      <div ng-if="ctrl.customFoldersEnabled" class="d-flex flex-row form-group">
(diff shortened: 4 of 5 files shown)


smithtb:
Working on fixing tests now.

views/windows/featureedit.html
```diff
@@ -16,6 +16,17 @@
           <simplemde text="ctrl.description" edit="true" is-required="false" maxlength="4000"></simplemde>
         </div>
       </div>
+
+      <div ng-if="ctrl.customFoldersEnabled" class="d-flex flex-row form-group">
+        <label class="col-2 col-form-label text-right">Folder</label>
```
wallw-bits: This markup doesn't match the other rows in the form, and is misaligned.

Screen Shot 2019-05-21 at 11 40 35 AM

views/windows/featureedit.html
```diff
@@ -16,6 +16,17 @@
...
+        <div class="col mt-2">
+          <select class="custom-select" ng-model="ctrl.folder"
+            ng-options="folder.label for folder in ctrl.folderOptions"
+            ng-change="ctrl.updateFolder()">
```
wallw-bits: This style of dropdown eliminates the tree structure of the folders and puts them in a flat list. Can we present the tree instead?

src/plugin/places/placesmanager.js
```diff
@@ -162,6 +165,44 @@ plugin.places.PlacesManager.prototype.getPlacesRoot = function() {
...
+      this.folders_.push(node);
+    }
+    if (node.getChildren()) {
+      for (var i = 0; i < node.getChildren().length; i++) {
+        var child = node.getChildren()[i];
+        this.addFolders_(child);
+      }
+    }
+  }
+};
```
wallw-bits: Ignoring the comment about presenting a tree, this is just generically flattening a tree. A generic function like this would be better placed on a static helper package for tree nodes. Also, there are several unnecessary class member items used here. This function could be more simply implemented as `function(ITreeNode, Array<ITreeNode>)` or `function(ITreeNode):Array<ITreeNode>` rather than storing the result of the function as a side effect on a member variable.

However, since we want to present a tree, this is probably no longer necessary.

src/plugin/file/kml/ui/kmlui.js
```diff
@@ -163,8 +163,15 @@ plugin.file.kml.ui.updatePlacemark = function(options) {
   // add the placemark to a parent if provided
   if (options['parent']) {
     var parent = plugin.file.kml.ui.verifyNodeInTree_(/** @type {!plugin.file.kml.ui.KMLNode} */ (options['parent']));
-    var cmd = new plugin.file.kml.cmd.KMLNodeAdd(placemark, parent);
-    os.commandStack.addCommand(cmd);
+    var currentParent = placemark.getParent();
+    if (currentParent) {
+      // update parent
+      currentParent.removeChild(placemark);
+      placemark.setParent(parent);
```
wallw-bits: Why does adding get a command on the stack but updating does not?

src/plugin/places/placesmanager.js
```diff
@@ -162,6 +165,44 @@ plugin.places.PlacesManager.prototype.getPlacesRoot = function() {
...
+ */
+plugin.places.PlacesManager.prototype.addFolders_ = function(opt_node) {
+  var node;
+  if (!opt_node) {
+    node = this.placesRoot_;
+  } else {
+    node = opt_node;
+  }
+  if (node) {
+    if (node.isFolder() && !this.folders_.includes(node)) {
```
wallw-bits: I don't believe it is possible to parent a folder under multiple parent folders, and thus the `includes` check should never be necessary.

src/plugin/file/kml/ui/kmlui.js
smithtb: I modified this to use the stack as-is, so removing and adding it. If we want to implement an update instead of just an add and remove that's fine, but might be better done outside of this ticket as it would affect more places in the code. Let me know what you think about the proposed changes.

src/plugin/places/placesmanager.js
smithtb: Fixed

views/windows/featureedit.html
smithtb: In other places we also present a flattened tree for folder-only structures (such as managing favorites), so I would prefer to keep this consistent with those choices. If we decide to change how we present those options in the future we may want to change them all.

views/windows/featureedit.html
smithtb: Fixed

src/plugin/places/placesmanager.js
smithtb: Since I left the flattened tree for now, I did make this change. If there is a better place to put that utility method or you have some feedback on its implementation/use, please let me know.

src/os/structs/nodeutils.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param {os.structs.ITreeNode} node
+ * @param {Array<os.structs.ITreeNode>} results
+ * @param {function(os.structs.ITreeNode): boolean} opt_filter
+ */
+os.structs.nodeutils.flattenTree = function(node, results, opt_filter) {
+  if (opt_filter ? opt_filter(node) : true) {
+    results.push(node);
+    var children = node.getChildren();
+    if (children) {
+      for (var i = 0; i < node.getChildren().length; i++) {
```
wallw-bits: This should be `for (var i = 0, n = children.length; i < n; i++)`. We always try to cache the length if it is not changing.

src/os/structs/nodeutils.js
```diff
@@ -0,0 +1,21 @@
...
+        var child = node.getChildren()[i];
```
wallw-bits: `var child = children[i]`

src/os/structs/nodeutils.js
```diff
@@ -0,0 +1,21 @@
+goog.provide('os.structs.nodeutils');
+
+
+/**
+ * Flattens a tree into a specified array - optionally filtering.
+ * @param {os.structs.ITreeNode} node
+ * @param {Array<os.structs.ITreeNode>} results
+ * @param {function(os.structs.ITreeNode): boolean} opt_filter
```
wallw-bits: The param name indicates this is optional but the type does not. Try `{function(os.structs.ITreeNode):boolean=}` to make it optional.

views/windows/featureedit.html
wallw-bits: I'll just go on record as stating that I disagree with that approach because two folders can have the same name but under different parent folders, and it would be difficult to then distinguish them from one another in a flat list.

src/os/structs/nodeutils.js
```diff
@@ -0,0 +1,21 @@
+goog.provide('os.structs.nodeutils');
```
wallw-bits: We dropped the "*utils" convention a while back. Utilities for arrays are on `os.array`, object are on `os.object`, etc. I think you can do `os.structs` or `os.structs.treenode` and that would fit better in that convention.

src/plugin/file/kml/ui/placemarkedit.js
```diff
@@ -95,6 +99,21 @@ plugin.file.kml.ui.PlacemarkEditCtrl = function($scope, $element, $timeout) {
...
+    os.structs.nodeutils.flattenTree(rootFolder, folders, function(node) {
+      return /** @type {plugin.file.kml.ui.KMLNode} */ (node).isFolder();
+    });
+    this['folderOptions'] = folders;
+
+    var parentFolder = this.options['parent'];
+    if (!parentFolder && this.options['node']) {
+      parentFolder = this.options['node'].getParent();
+    }
+    this['folder'] = parentFolder ? parentFolder : rootFolder;
```
wallw-bits: For items which are not booleans or numbers, this is more succinctly written as `this['folder'] = parentFolder || rootFolder.`

src/os/structs/nodeutils.js
```diff
@@ -0,0 +1,21 @@
...
+
+
+/**
+ * Flattens a tree into a specified array - optionally filtering.
+ * @param {os.structs.ITreeNode} node
+ * @param {Array<os.structs.ITreeNode>} results
+ * @param {function(os.structs.ITreeNode): boolean} opt_filter
+ */
+os.structs.nodeutils.flattenTree = function(node, results, opt_filter) {
+  if (opt_filter ? opt_filter(node) : true) {
```
wallw-bits: One alternative here is to use a default function:
```
opt_filter = opt_filter || goog.functions.TRUE;
if (opt_filter(node)) {
  ...
```

src/os/structs/nodeutils.js
smithtb: Fixed

src/os/structs/nodeutils.js
smithtb: Fixed

src/os/structs/nodeutils.js
smithtb: Fixed

src/os/structs/nodeutils.js
smithtb: Fixed

src/plugin/file/kml/ui/placemarkedit.js
smithtb: Fixed

src/os/structs/nodeutils.js
smithtb: I went with your suggestion - good tip.
Thanks for catching all these things!

src/os/structs/structs.js
```diff
@@ -0,0 +1,21 @@
...
+ * Flattens a tree into a specified array - optionally filtering.
+ * @param {os.structs.ITreeNode} node
+ * @param {Array<os.structs.ITreeNode>} results
+ * @param {function(os.structs.ITreeNode): boolean=} opt_filter
+ */
+os.structs.flattenTree = function(node, results, opt_filter) {
+  opt_filter = opt_filter || goog.functions.TRUE;
+  if (opt_filter(node)) {
+    results.push(node);
+    var children = node.getChildren();
```
wallw-bits: This only recurses if the result of the filter is `true`. Was that your intention? I don't think that is correct.

smithtb: Thanks! Fixed that

views/windows/featureedit.html
schmidtk: Completely agree with Will here. Saved Places is not a good candidate for displaying a flattened list, both because we often see folders with the same name and because the tree can get quite large and a list view would be easy to navigate/understand.

jsalankey: The standard behavior for this kind of thing is most comparable to the Bookmark UI present in browsers. Chrome flattens the list and repeatedly shows identical folder names which, if confusing, is the user's fault or the code's fault. I know we just implemented something that will create duplicate folder names @schmidtk, but it would be better for it to try to number them.

Tree structures are not browsable, we don't have a tree search that maintains tree structure, and I think we should do everything we can to not show trees anywhere ever. The context of the tree being totally devoid of leaf nodes would also look very strange in my opinion. They suck if the tree becomes too deep because it's hard to even show all of the folder depth in the available horizontal space. We've been trying for years to get rid of the Add Data window for exactly these sorts of reasons.

I think a select2 is a better solution to the problem of a large tree structure and quickly finding the folder you want.

schmidtk: Chrome presents a list of commonly used folders, along with a "Choose another folder" option that shows you the tree structure. This accommodates users with a lot of bookmarks, who may not find what they want in the limited initial list. Currently if you have a lot of places, you get something like this:
Folders
I know we have folks with _way_ more places than that, making the list even more difficult to use.

A select2 would at least give them a way to search, but it would still present the duplicate name problem. That's not entirely in their control, because we allow saving an entire KML tree from an imported file.

smithtb:
Live demo: http://places-folder-options-opensphere.surge.sh/

Test:
1. Add folders to the places tab (some nested)
2. Create a place from the map (right click menu > Save to Places)
    - verify that a folder choice is presented and defaulted to saved places
    - verify that the other folders are listed as options in a flattened tree
3. Choose a folder for the place and save it
    - verify that the place shows under the correct folder
4. Edit the place you just created
    - verify that the folder selected is correct
5. Choose a different folder and save
    - verify that the place moves to the correct folder
6. Create a place from a folder other than saved places (click the place icon to the right of the folder name in the places manager)
    - verify that the folder you created the place from is chosen
7. Save the place
    - verify that it saved to the correct folder
8. Select a folder in the places manager by clicking it
9. Create a place from the top of the places manager (Add Place button)
    - verify that the folder selected is correct (whatever folder you clicked in step 8)
10. Try out other combinations of creating/editing with places and folders as needed

mgettman:
Test looks good!

views/windows/featureedit.html
jsalankey: Another option might be adding spacers/tabs in to the front of the displayed text based on the depth of the folder. This could simulate the nesting structure without having any weirdness with the tree search or with collapsing/expanding nodes. It would also avoid having to build a whole custom UI (though would still have the problem with very deep nesting being terrible looking or unusable).

PR #560, opened 2019-05-17
feat(csvImport): Add option to pick icon as part of CSV import

welchyd:
When importing a file, the options step allows selection of an initial color - this adds the choice of selecting an icon as well

Screenshot from 2019-05-21 11-23-16

Live demo: http://csv-icon-branch-opensphere-ngageoint.surge.sh/
To test it out:
1. Import a CSV file and in the last step pick a shape such as an icon
2. Verify that the created layer starts off with it's features showing that shape
3. Import another CSV file and verify that not selecting an icon defaults the features to be a point
4. Repeat steps 1-3 with several other file types - shp, geojson, gml, gpx, kml, and georss should all be supported


src/plugin/file/csv/csvdescriptor.js
```diff
@@ -181,6 +181,8 @@ plugin.file.csv.CSVDescriptor.createFromConfig = function(config) {
  */
 plugin.file.csv.CSVDescriptor.updateFromConfig = function(descriptor, config) {
   descriptor.setColor(config['color']);
+  descriptor.setIcon(config['icon']);
+  descriptor.setShapeName(config['shapeName']);
```
wallw-bits: I know this is static, but since the options step is available on nearly all import types, common items like this should be consolidated in another method and called from each descriptor type.

wallw-bits:
Please note that the wizard step in question is present on nearly every layer import type other than KML. You'll need to also test GeoJSON, GML, GPX, and Shapefile.

src/plugin/file/csv/csvdescriptor.js
welchyd: Interesting ... there's a lot of repeated code across those import types. Seems static functions can still be inherited, I'll try it out.

schmidtk: Yes, SHP supports all geometry types including Point/MultiPoint.

schmidtk:
Did you consider replacing the Color control with Style, to include both the color picker and shape dropdown? That would be more consistent with the layer style UI, would remove the need for an additional checkbox to toggle icon selection on/off, and would provide the user with more styling options. It could stick to the basic options (Default/Point/Square/Triangle/Icon) for now, but I could see users wanting to enable ellipses on import if the appropriate columns were available on the data.

welchyd:
I was considering it, but that's more than what I need of just the icon option.
The shapeui brings in icon rotation, lob, ellipse, and the ellipse center shape.
Also, the user can go back to different steps and change the values that determine the shape options.
The changes I'm making to support the icon will help make it easier to add though options in the future though

src/plugin/file/csv/csvdescriptor.js
wallw-bits: > Interesting ... there's a lot of repeated code across those import types. Seems that static functions can still be inherited, I'll try it out.

To be clear, static functions are not inherited. You're just consolidating repeated logic to a function and calling it from those places instead.

wallw-bits:
I'm with @schmidtk on this. I don't think we should be duplicating functionality from the Layer UI in the wizard step. It either needs to be refactored to be reusable (or the similar UI in Add/Edit Place needs to be reusable), or it needs to be removed altogether. We really shouldn't have 3 different UIs for setting all of that stuff.

welchyd:
I agree - and went with the recommendation from @schmidtk.
He let me know yesterday he digs where it's at now.

mgettman:
Looks good!

PR #565, opened 2019-05-23
fix(state): Support save / restore of timeline slices.

### LINKED ISSUE #467: Time slices not saved to state
If you create a time slice on the time line and then save the state, when you load the state the time slice is not remembered.
Other timeline elements are saved just fine - hold range, load range, view range, animate  range. In a similar fashion, save the time slice ranges too

bradh:
Resolves #467.

This will fail until https://github.com/ngageoint/opensphere-state-schema/pull/10 is merged and opensphere-state-schema is released as 2.3.0. I tested locally by modifying the package.json to 2.2.0 and then copying in the required files - the workspace resolver doesn't seem to do the right thing here.

### DIFF
--- a/src/os/state/v4/timestate.js
+++ b/src/os/state/v4/timestate.js
@@ -37,7 +37,12 @@ os.state.v4.TimeTag = {
   HOLDS: 'heldIntervals',
   HOLD_ITEM: 'held',
   SEQ_INTERVAL: 'interval',
-  TIME: 'time'
+  TIME: 'time',
+  SLICES: 'slices',
+  SLICE: 'slice',
+  SLICE_INTERVAL: 'sliceInterval',
+  INTERVAL_START: 'intervalStart',
+  INTERVAL_END: 'intervalEnd'
 };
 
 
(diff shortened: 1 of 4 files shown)


schmidtk:
@bradh Did you run `yarn upgrade` with `opensphere-state-schema` in your `workspace` folder? That should create a symlink to the project in the hoisted `node_modules` so the resolver will use that, and tests will pass. You shouldn't need to modify `package.json` or copy files around.

src/os/state/v4/timestate.js
```diff
@@ -520,3 +530,57 @@ os.state.v4.TimeState.prototype.parsePeriod = function(period) {
...
 };
+
+/**
+ * Returns slice intervals element for timeranges.
+ * @param {Array<goog.math.Range>} sliceRanges
+ * @return {!Element}
+ * @private
+ */
+os.state.v4.TimeState.prototype.sliceRangesToXml_ = function(sliceRanges) {
+  let slices = os.xml.createElement(os.state.v4.TimeTag.SLICES);
```
schmidtk: Please use `var` instead of `let`. The compiler will replace these with `var`, but we haven't officially started allowing ES6 that can be transpiled by Closure. This is in part to ensure we can still load the debug build in older supported browsers, but also for the sake of consistency across the code base.

bradh: OK. Changed to using `var`.

PR #578, opened 2019-05-30
fix: Many windows

### LINKED ISSUE #506: Many edit place dialogs for same place
When I go to edit a place, if I click on the edit button several times, it spawns multiple windows.
There should be one unique dialog that is brought to the front if it is already open.
Also, if I edit a place and mess with the style accordion options in the layers window there can be conflicting changes. (This last part is captured in #573 )

editEditEdit

0.5 story points

adamstja:
click of edit btn created multiple windows. now only one window.

fix #506

### DIFF
--- a/src/plugin/file/kml/ui/kmlui.js
+++ b/src/plugin/file/kml/ui/kmlui.js
@@ -105,7 +105,7 @@ plugin.file.kml.ui.updateFolder = function(options) {
  */
 plugin.file.kml.ui.createOrEditPlace = function(options) {
   var windowId = 'placemarkEdit';
-  windowId +=  options['feature'] ? ol.getUid(options['feature']) : goog.string.getRandomString();
+  windowId += options['feature'] ? ol.getUid(options['feature']) : goog.string.getRandomString();
   var scopeOptions = {
     'options': options
   };
--- a/src/plugin/file/kml/ui/kmlui.js
+++ b/src/plugin/file/kml/ui/kmlui.js
@@ -105,29 +105,33 @@ plugin.file.kml.ui.updateFolder = function(options) {
  */
 plugin.file.kml.ui.createOrEditPlace = function(options) {
   var windowId = 'placemarkEdit';
+  windowId +=  options['feature'] ? ol.getUid(options['feature']) : goog.string.getRandomString();
   var scopeOptions = {
     'options': options
   };
-
   var label = (options['feature'] ? 'Edit' : 'Add') + ' Place';
-  var windowOptions = {
-    'id': windowId,
-    'label': label,
-    'icon': 'fa fa-map-marker',
-    'x': 'center',
-    'y': 'center',
-    'width': 700,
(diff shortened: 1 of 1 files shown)


src/plugin/file/kml/ui/kmlnodeui.js
```diff
@@ -13,7 +13,7 @@ goog.require('plugin.file.kml.cmd.KMLNodeRemove');
 plugin.file.kml.ui.kmlNodeUIDirective = function() {
   return {
     restrict: 'E',
-    replace: true,
+    replace: false,
```
schmidtk: Why was this needed? We generally try to use `replace: true` because it prevents breaking the chain of flex items (the directive element will not have `d-flex`) which can cause layout problems.

src/plugin/file/kml/ui/kmlui.js
```diff
@@ -110,6 +110,9 @@ plugin.file.kml.ui.createOrEditPlace = function(options) {
   };
 
   var label = (options['feature'] ? 'Edit' : 'Add') + ' Place';
+  if (os.ui.window.exists(windowId)) {
+    os.ui.window.bringToFront(windowId);
+  } else {
   var windowOptions = {
```
schmidtk: This section isn't tabbed in, causing lint errors. Do you have eslint configured in your editor?

src/plugin/file/kml/ui/kmlui.js
```diff
@@ -110,6 +110,9 @@ plugin.file.kml.ui.createOrEditPlace = function(options) {
   };
 
   var label = (options['feature'] ? 'Edit' : 'Add') + ' Place';
+  if (os.ui.window.exists(windowId)) {
```
schmidtk: This currently causes strange behavior if I start creating a new place, then click edit on an existing place (the create window is brought to the front, when I'm expecting to get an edit window). Similarly if I click edit on one place, then click edit on another or try creating a new place.

This should use a unique id for creating a new place, and editing each existing place. I would suggest using the existing id for create, and for the edit case append `ol.getUid(options['feature'])` to the id.

src/plugin/places/placesmenu.js
```diff
@@ -397,7 +397,7 @@ plugin.places.menu.visibleIfIsPlace = function(context) {
 plugin.places.menu.onSpatialEdit_ = function(event) {
   var features = os.ui.menu.spatial.getFeaturesFromContext(/** @type {Object} */ (event.getContext()));
   if (features.length === 1) {
-    var feature = features[0];
+    var feature = features[1];
```
schmidtk: This was correct as is. If the array length is 1, there will only be an item at index 0.

src/plugin/file/kml/ui/kmlui.js
adamstja: Sounds good. Random string was added to allow multiple new places to be created at the same time.

PR #542, opened 2019-05-09
feat(annotations) :add color options annotation

robertbolan-bits:
re #528 
Added the ability to select colors for annotations Headers and Body. The color of the tail should match the color selected for the body unless you are not showing the body. If you are not showing the body then the color will match the header.

PR #583, opened 2019-05-31
test(cypress): Increases comparison wait to 2500, updates to images

justin-bits:
https://github.com/ngageoint/opensphere/pull/542 was failing due to the new image comparisons.  Likely other builds too.  This is the fix.


justin-bits:
Added: The target server for a legend icon was missing.  This was rendering differently between Chrome/Electron, so I just removed references to the legend and updated the images.

Issue #385, opened 2019-01-16
Triple clicking on menus leaves the menu closed


justin-bits:
On rapid triple click, a menu should be open (open/close/open), instead the menu doesn't open again on the third click.  Found during testing with Cypress.  This results in two flaky tests https://github.com/justin-bits/opensphere-tests/issues/25.

This is reproducible by manually triple clicking on Add Data dropdown, Save Options, or State Options from the toolbar.

To clarify, the test isn't testing a triple click. There are three tests back to back that use the dropdown menus.  They complete fast enough, that it results in a triple click on the menus.

justin-bits:
I'm also seeing this with Using Next/Prev buttons back to back.  In this case, it's not a double or triple click, it's clicking a button to change the date in one direction, then rapidly clicking a button to change the date back to where it was: https://github.com/justin-bits/opensphere-tests/blob/6d6329770c6c0ebb7b138dc3cade7ff2a42a77e7/cypress/integration/smoke-tests/toolbar-center.spec.js#L22

![image](https://user-images.githubusercontent.com/44378016/51277940-539b8280-1996-11e9-8798-cd88c6246c42.png)

wallw-bits:
@justin-bits, @csgibson is correct. That is unrelated.

PR #584, opened 2019-05-31
fix(mapping): autodetect all ellipse mappings

wallw-bits:
Ellipses require SemiMajor, SemiMinor, and Orientation mappings to work properly.


justin-bits:
This build is failing due to the issue described in #385.

justin-bits:
After this is merged https://github.com/ngageoint/opensphere/pull/590, grab those changes and it should resolve the issue.

Issue #579, opened 2019-05-31
Polygon queries crossing antimeridian broken on GeoServer


wallw-bits:
Apparently GeoServer allows Envelope queries that cross the antimeridian (e.g. `[160, 0, 200, 50]`), but does not allow polygons/multipolygons with longitudes outside of `[-180, 180]`.

We should fix this by splitting the polygon on the antimeridian and normalizing/clamping all coordinates within the origin projection "world" before writing the query. This is probably going to need some of the same JSTS splitting functions being added for #540.

PR #585, opened 2019-05-31
feat: add polar polygon support

### LINKED ISSUE #540: Polar polygon support
Our support for polar polygons leaves a lot to be desired. Drawing or loading a polygon near (but not necessarily containing) the pole can result in unexpected behavior.

3D (note artifacts):
Screen Shot 2019-05-09 at 9 28 09 AM

The artifacts come from how the polygon is represented in 2D, which is visualized like so:
Screen Shot 2019-05-09 at 9 30 10 AM

Even more regular polygons which do not produce artifacts
Screen Shot 2019-05-09 at 9 26 59 AM

still have issues with their 2D representation:
Screen Shot 2019-05-09 at 9 27 32 AM

These polygons should have both the correct 2D representation (to be properly queried against the spatial index and against remote servers) and the correct 3D display.

wallw-bits:
Adds support for polar polygons by splitting them by the meridian plane for display/query in 2D. Additionally, queries for OGC servers such as GeoServer are now split over the antimeridian in the event that the remote server does not allow coordinates outside of the longitudes [-180, 180].

resolves #540, #579

### DIFF
--- a/config/settings.json
+++ b/config/settings.json
@@ -52,6 +52,32 @@
       "methods": ["GET", "POST", "PUT", "DELETE"],
       "encode": false
     },
+    "projections": [
+      {
+        "code": "EPSG:3031",
+        "extent": [
+          -12369000,
+          -12369000,
+          12369000,
+          12369000
+        ],
+        "isGlobal": false,
+        "proj4": "+proj=stere +lat_0=-90 +lat_ts=-71 +lon_0=0 +k=1 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs",
+        "title": "Antarctic Polar Stereographic"
+      },
+      {
+        "code": "EPSG:3413",
+        "extent": [
+          -12369000,
+          -12369000,
+          12369000,
+          12369000
+        ],
+        "isGlobal": false,
+        "proj4": "+proj=stere +lat_0=90 +lat_ts=70 +lon_0=-45 +k=1 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs",
+        "title": "NSIDC Sea Ice Polar Stereographic North"
+      }
+    ],
     "providers": {
       "basemap": {
         "type": "basemap",
--- a/src/os/geo/jsts.js
+++ b/src/os/geo/jsts.js
@@ -523,6 +523,198 @@ os.geo.jsts.toPolygon = function(geometry) {
 
 
 /**
+ * @param {jsts.geom.Geometry} geometry
+ * @return {Array<jsts.geom.Polygon>}
+ * @private
+ */
+os.geo.jsts.polygonize_ = function(geometry) {
+  var lines = jsts.LineStringExtracter.getLines(geometry);
+  var polygonizer = new jsts.operation.polygonize.Polygonizer();
+  polygonizer.add(/** @type {jsts.Collection<jsts.geom.Geometry>} */ (lines));
+  var polys = polygonizer.getPolygons();
+  return jsts.geom.GeometryFactory.toPolygonArray(polys);
+};
+
+
+/**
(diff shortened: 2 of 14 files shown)


config/settings.json
```diff
@@ -52,6 +52,32 @@
...
+          -12369000,
+          -12369000,
+          12369000,
+          12369000
+        ],
+        "isGlobal": false,
+        "proj4": "+proj=stere +lat_0=90 +lat_ts=70 +lon_0=-45 +k=1 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs",
+        "title": "NSIDC Sea Ice Polar Stereographic North"
+      }
+    ],
```
wallw-bits: Not positive we want to include these. I hardcode the `proj4` defs later on so that the polar split function can work even if they are not provided here (and they won't show up in the UI).

wallw-bits:
Ah, that test failure reminds me: JSTS, when splitting/polygonizing, does not linearly interpolate the altitude between the two points. Do we want to do that?

src/os/geo/jsts.js
```diff
@@ -526,6 +526,198 @@ os.geo.jsts.toPolygon = function(geometry) {
...
+  }
+
+  return false;
+};
+
+
+proj4.defs('EPSG:3031',
+    '+proj=stere +lat_0=-90 +lat_ts=-71 +lon_0=0 +k=1 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs');
+proj4.defs('EPSG:3413',
+    '+proj=stere +lat_0=90 +lat_ts=70 +lon_0=-45 +k=1 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs');
```
jsalankey: Placement of these seems random. Do we want to keep these here, or are we loading them from settings?

src/os/geo/jsts.js
```diff
@@ -526,6 +526,198 @@ os.geo.jsts.toPolygon = function(geometry) {
...
+  } else if (os.geo.jsts.needsSplit(/** @type {ol.geom.Polygon} */ (polygon))) {
+    var north = ol.extent.getCenter(polygon.getExtent())[1] >= 0;
+    polygon.toLonLat();
+    var pLonLat = ol.proj.get(os.proj.EPSG4326);
+    var pPolar = ol.proj.get(north ? 'EPSG:3413' : 'EPSG:3031');
+    var pFinal = os.map.PROJECTION;
+
+    polygon.transform(pLonLat, pPolar);
+    var meridian = new ol.geom.LineString([[0, 0], [0, north ? 90 : -90], [180, 0]]);
+    meridian.transform(pLonLat, pPolar);
```
jsalankey: Rather than creating and re-transforming the two meridians each time, we could just create them the first time then store them off to a static variable for re-use.

test/os/geo/jsts.test.js
```diff
@@ -0,0 +1,109 @@
...
+        } else if (coord[0] === 0) {
+          expect(coord[2]).toBe(3);
+        } else if (coord[0] === 2) {
+          expect(coord[2]).toBe(4);
+        }
+      }
+    });
+  });
+
+  it('should split north polar polygons properly', function() {
```
jsalankey: It would probably be good to add tests covering the cases of the input being a `MultiPolygon`. Probably an unrealistic case, but maybe even check a `MultiPolygon` where one wraps the north pole and the other the south?

src/os/geo/jsts.js
wallw-bits: They must exist in `proj4` regardless of whether they are configured in settings for use by the user.

wallw-bits:
This algorithm does not work well for large polygons due to the stereographic projections. I'm considering adapting Cesium's approach of splitting the polygon by the meridian plane in 3D Cartesian space.

PR #586, opened 2019-06-01
chore(deps): Upgrade replace dependency.

bradh:


wallw-bits:
@bradh I think this is a defunct dependency and it can be removed.

bradh:
Will close in favour of a future PR to remove the dep.

PR #590, opened 2019-06-03
test(cypress): Adds waits to avoid rapidly changing dates to prevent …

justin-bits:
…test failure

This build was failing due to this issue.  In the future a more broad solution could be considered.


wallw-bits:
This is a bandaid fix. We need to write up an issue to fix the underlying issue in a more central fashion so that waits/sleeps aren't littered all over the tests.

PR #594, opened 2019-06-03
fix: Clear button window closure

adamstja:
Clear button on log no longer closes window, only clears log in FF60.7.0esr and Chrome

### DIFF
--- a/src/os/debug/fancierwindow.js
+++ b/src/os/debug/fancierwindow.js
@@ -43,10 +43,6 @@ os.debug.FancierWindow.STYLE_RULES_ = goog.string.Const.from(
 os.debug.FancierWindow.prototype.writeInitialDocument = function() {
   os.debug.FancierWindow.superClass_.writeInitialDocument.call(this);
 
-  // close button should set the debug window to disabled
-  if (this.win) {
-    goog.events.listenOnce(this.win, goog.events.EventType.BEFOREUNLOAD, this.onBeforeUnload_, false, this);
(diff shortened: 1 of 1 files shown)


adamstja:
Test Steps:
Using Fire Fox version 60.7.0esr
1. Top right hand corner click the "support" button
2. Drop down appears
3. Click "View Log"
4. Top right corner select "clear"
5. After selecting "clear" the data should be cleared, and the window should remain open.

src/os/debug/fancierwindow.js
```diff
@@ -43,10 +43,6 @@ os.debug.FancierWindow.STYLE_RULES_ = goog.string.Const.from(
 os.debug.FancierWindow.prototype.writeInitialDocument = function() {
   os.debug.FancierWindow.superClass_.writeInitialDocument.call(this);
 
-  // close button should set the debug window to disabled
-  if (this.win) {
-    goog.events.listenOnce(this.win, goog.events.EventType.BEFOREUNLOAD, this.onBeforeUnload_, false, this);
```
schmidtk: I think the intent of this handler was to close the log window when the application was closed/refreshed. It should be listening to `window` (the application window) instead of `this.win` (the log window). Please make that change, and also remove the `unlisten` call below because `listenOnce` will remove the handler when the callback fires.

adamstja: Will do. Thanks for the feed back.

PR #600, opened 2019-06-06
Fix-594(Window Closure)

adamstja:
Clear button on log no longer closes window, only clears log in FF60.7.0esr and Chrome.
Test Steps:
Using Fire Fox version 60.7.0esr

1.Top right hand corner click the "support" button
2.Drop down appears
3.Click "View Log"
4.Top right corner select "clear"
5.After selecting "clear" the data should be cleared, and the window should remain open.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "init": "npm run init:base && npm run init:debug && npm run init:version && npm run init:vfolder",
     "lint:css": "sass-lint -v -s scss -i 'scss/nolint/**/*.scss' 'scss/**/*.scss'",
     "lint:js": "eslint --cache 'src/**/*.js' 'cypress/**/*.js'",
-    "lint": "npm run lint:css",
+    "lint": "npm run lint:js && npm run lint:css",
     "compile:dossier": "npm run init:apidoc && os-docs-gen-config dossier.conf.json .build/dossier.conf.json && os-docs -c .build/dossier.conf.json",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -92,7 +92,7 @@
     "init": "npm run init:base && npm run init:debug && npm run init:version && npm run init:vfolder",
     "lint:css": "sass-lint -v -s scss -i 'scss/nolint/**/*.scss' 'scss/**/*.scss'",
     "lint:js": "eslint --cache 'src/**/*.js' 'cypress/**/*.js'",
-    "lint": "npm run lint:js && npm run lint:css",
+    "lint": "npm run lint:css",
```
wallw-bits: Why did you disable `eslint`?

src/os/debug/fancierwindow.js
```diff
@@ -42,9 +42,8 @@ os.debug.FancierWindow.STYLE_RULES_ = goog.string.Const.from(
  */
 os.debug.FancierWindow.prototype.writeInitialDocument = function() {
   os.debug.FancierWindow.superClass_.writeInitialDocument.call(this);
-
   // close button should set the debug window to disabled
-  if (this.win) {
+  if (this.window) {
```
schmidtk: This only fixes the issue because `this.window` is always undefined and you'll never register a `beforeunload` listener. A correct solution should ensure the log window closes when the application is closed/refreshed, which is what this listener was intended to do.

PR #601, opened 2019-06-06
feat: webgl fly to should fly to bounding sphere rather than extent

wallw-bits:
FlyTo in 3D was still using the 2D `FlyToExtent`, which does not take altitude into account and could potentially fly the camera too far in to see the feature.

Test:
* Make some places with and without altitude
* Right click each one and select "Go To" in 2D and 3D
  * In 3D, tilt the camera before and verify that it keeps the general orientation after flying to the feature(s)
* Place the features in a folder and select "Go To" on the folder in both 2D and 3D

### DIFF
--- a/externs/Cesium.externs.js
+++ b/externs/Cesium.externs.js
@@ -1097,9 +1097,9 @@ Cesium.Camera.prototype.flyTo = function(options) {};
 
 /**
  * @param {!Cesium.BoundingSphere} boundingSphere The bounding sphere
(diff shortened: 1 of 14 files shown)


bradh:
Failing for pixel differences in a Cypress test (`smoke-tests/load-state-file-arcgis.spec.js`). May or may not be related.

wallw-bits:
> Failing for pixel differences in a Cypress test (`smoke-tests/load-state-file-arcgis.spec.js`). May or may not be related.

We are currently operating under the assumption that any failure in those tests is unrelated to what we are working on.

Issue #537, opened 2019-05-08
Annotations Manager


welchyd:
Setup the places tab to function as a place to manage annotations
(update as needed)

welchyd:
Closed via #607

PR #607, opened 2019-06-12
feat(annotations): adds annotations folder to places, numerous manage…

jsalankey:
For #537 

- Adds a permanent folder to places where all new annotations are added. 
- Abstracts core annotation code out from the model/UI classes for future use. 
- Adds inline editing of annotation name/description on double click.
- Fixes annotations for non-point geometries not working at all.
- Adds ability to create annotations on non-point geometries.
- Adds nearest point anchor calculation for non-point geometries.
- Cleans up menus for creating annotations.

This was originally a bigger PR that refactored AreaManager to store its areas as Places, but that had far too significant of UX impacts and was pulled out. There are further UX improvements that can be made to the separation of areas and places.


schmidtk:
* When creating a new annotation, I get this error and my annotation appears on the map but is not added to Saved Places:
```
Error running command [ Add KML Node "New Place" ] - Node source is not available.
```
* Setting the shape to None when creating a new annotation was intentional, so the map could be annotated without drawing something additional in that location. Given the tail is anchored to the position, it doesn't seem necessary to draw an extra point that may obscure what's being annotated.
* The Annotations folder feels awkward to me. I want to organize annotations myself, and while I can still put them anywhere the extra folder that can't be removed feels like unnecessary clutter.

jsalankey:
- Fixed.
- Fixed. Right click annotations now default to None, while non-point ones use Default.
- I wanted to make it clear that Places was where Annotations are managed. I think there will be some followon work related to Annotation-specific behavior that I want to add to it (import/export annotations, show/hide all, etc.) I don't think it's ideal, but we decided against making annotations entirely their own thing (which was the right decision), but I still want to make them more clearly a special type of place, and to have somewhere to manage any special functionality we give them.
- Fixed. This was related to the initialization state of the overlay and the limitation that the `nearestPoints` JSTS calculating doesn't work when the annotation directive is not floating over a coordinate on the map. It will now default back to the first coordinate if it can't calculate the position correctly. If there's a smart solution to `nearestPoints` not working off the globe (i.e. not reimplementing it entirely), then we could explore that.

schmidtk:
* When inline editing the title/description, if I resize the annotation then save my text changes, the new annotation size is kept. I would prefer if it always reverted to the pre-edit annotation size.
* The inline description edit is not responsive to resize. The buttons should be fixed in the bottom/right, and the editor should fill the rest of the available space.
* When creating/editing annotations on non-point geometries, the annotation moves on save if positioned along an edge (between vertices). Draw a box, Create Annotation, move the annotation to anchor along an edge, then save. This also happens if I position an existing annotation along an edge, then edit the annotation.

jsalankey:
- Fixed
- I'm going to defer on this one. Sizing on simpleMDE is always a little tricky, and Greg's in the process of replacing it with ToastUI anyway, which may have its own sizing quirks to figure out. We'll have to handle that merge eventually anyway.
- Looking at this.

PR #608, opened 2019-06-12
build(cypress): Disable Cypress in Travis.

schmidtk:
This also disables API doc generation due to problems with the current version of js-dossier.


schmidtk:
Leaving Cypress in the Travis build. Will re-evaluate if other issues crop up in the future.

PR #610, opened 2019-06-12
test(travis): Fix for missing package that is causing Cypress to fail…

justin-bits:
… to load


justin-bits:
https://github.com/cypress-io/cypress/issues/4069

Issue #593, opened 2019-06-03
Documentation needs update to cover Cypress artifact upload to Amazon S3


justin-bits:
This PR https://github.com/ngageoint/opensphere/pull/543 added Cypress artifact upload to S3, but the documentation wasn't updated accordingly.

- Explain that artifact upload only occurs for non-PR builds
- How to get uploads working for personal builds in Travis

wallw-bits:
Resolved by #614

PR #614, opened 2019-06-12
Adds instructions for accessing Cypress artifacts

justin-bits:
Also changes bucket name to use an environment variable instead of being hard coded.

https://github.com/ngageoint/opensphere/issues/593


docs/guides/cypress_artifacts.rst
```diff
@@ -0,0 +1,42 @@
+Cypress Artifacts
+#################
+On Cypress test failure, test artifacts like screenshots, image comparisons, and videos are automatically generated. These artifacts are easily accessible from the local file system for test runs via the GUI and CLI.  To obtain artifacts from test runs under CI, custom solutions may be required as the build artifacts are often destoryed after a build is completed.  Instructions for Travis CI is below; other CI environments can use a similar implementation.
```
bradh: destroyed (rather than destoryed).

justin-bits: Good catch. Fixed.

PR #615, opened 2019-06-13
feat(lineDash): Add line style options to support a dashed border

### LINKED ISSUE #532: Line Styling Options
In the Add/Edit Place dialog, if a shape is a line or polygon, add controls for styling the stroke (line color, thickness, opacity (slider and numeric value), dot/dash)
THIN-13362
5 story points

welchyd:
In the Add/Edit Place dialog, if a shape is a line or polygon, add controls for styling the line Dash of the stroke 

fix #532

Live Demo: 3d-dash.surge.sh
Test steps:
1. Draw a polygon (square, circle, or polygon) on the map
2. In the style accordion try out various border options
3. Verify that the polygon border changes in the preview as different options are selected
4. Save the polygon and verify that it remembers the dash style when you edit it
5. Draw a line on the map and verify the border options
6. Verify that editing the line reflects the border dash style
7. Verify that the line style looks similar on the the map between 2D and 3D mode
8. In the layers menu expand the Saved Places layer and click on a place
9. Verify that the style accordion correctly shows the border dash style
10. Verify that you can change the dash style
11. Verify that the style works with the command stack (ctrl + z and ctrl + y)
12. Verify that exporting Saved Places as a place preserves the border dash style
13. Export saved places as a kml, verify that it imports with the boarder dash style
14. Select the kml layer in the layers window and in the style accordion check 'replace feature style'
15. Verify that all shapes reflect the selected border dash style
16. Export Saved Places as a csv, verity that it does not import with the border dash style
17. Select the csv layer in the layers window and from the style accordion verify that all shapes reflect the selected border dash style
18. In the layers window, right click on the csv file and from the menu select feature actions
19. Create a feature action that sets the style to have a dashed border
20. Verify that the dashed border changes as expected
21. Verify that exporting/importing the feature action with dashed border style works
22. Refresh Open Sphere and verify that the dash styles are persisted


bradh:
Looks like the failure is for one of the state tests:

```
Firefox 63.0.0 (Linux 0.0.0) OMAR.v4.ArcLayerState Arc state validates against the state.xsd FAILED
	Expected [ 'file_0.xml:1: element lineDash: Schemas validity error : Element '{http://www.bit-sys.com/mist/state/v4}lineDash': This element is not expected.' ] to be null.
	@test/plugin/arc/state/v4/arclayerstate.test.js:429:36
```

Probably related.

welchyd:
This change covers a lot of things - I'm guessing there's bound to be a bit of feedback.
That test has some deeper roots to state files and the schema that I'll be working on the meantime.

src/os/style/strokereader.js
```diff
@@ -31,8 +31,13 @@ os.style.StrokeReader.prototype.getOrCreateStyle = function(config) {
 
   hash = 31 * hash + goog.string.hashCode(color) >>> 0;
 
+  var dash = /** @type {!Array<number>} */ (config['lineDash']) || os.style.DEFAULT_LINE_STYLE;
+
+  hash += goog.string.hashCode(dash.join());
```
schmidtk: Please use the same method of updating the hash as other config components:
```
hash = 31 * hash + goog.string.hashCode(dash.join()) >>> 0;
```
This method avoids style collisions by isolating each component added to the hash.

WRT later comments, only change the hash if `lineDash` is defined. The default should be `undefined`.

src/os/style/style.js
```diff
@@ -57,6 +57,60 @@ os.style.DEFAULT_FEATURE_SIZE = 3;
...
+os.style.DEFAULT_LINE_DASH_STYLE_FIELDS = [
+  os.style.StyleField.STROKE
+];
+
+
+/**
+ * Default line style for geometries. Specifically for lines/polygons.
+ * @type {!Array<number>}
+ */
+os.style.DEFAULT_LINE_STYLE = [];
```
schmidtk: The default `lineDash` should be `undefined`, so this constant shouldn't be needed at all.

If a line is not dashed, all configs/styles/etc should be exactly as they were before this PR and there shouldn't be any additional processing required, other than the property check.

src/os/style/style.js
```diff
@@ -302,7 +356,8 @@ os.style.DEFAULT_VECTOR_CONFIG = {
   // this will only be applied to line and polygon types
   'stroke': {
     'width': os.style.DEFAULT_STROKE_WIDTH,
-    'color': os.style.DEFAULT_LAYER_COLOR
+    'color': os.style.DEFAULT_LAYER_COLOR,
+    'lineDash': os.style.DEFAULT_LINE_STYLE
```
schmidtk: This property should not be included at all if the line is not dashed.

src/os/ui/featureedit.js
```diff
@@ -696,6 +702,21 @@ os.ui.FeatureEditCtrl.prototype.isEllipse = function() {
...
+/**
+ * If a line or polygon is selected.
+ * @return {boolean}
+ * @export
+ */
+os.ui.FeatureEditCtrl.prototype.isPolygonOrLine = function() {
+  var geometry = this.previewFeature.getGeometry();
+
+  return (
+    geometry instanceof ol.geom.Polygon || geometry instanceof ol.geom.MultiPolygon ||
```
schmidtk: Use `geometry.getType()` (considering that the geometry could be null/undefined) and `ol.geom.GeometryType` to test the type. `instanceof` is much slower and unreliable if the geometry was created in another window context.

src/os/ui/layer/labelcontrols.js
```diff
@@ -62,6 +63,12 @@ os.ui.Module.directive('labelcontrols', [os.ui.layer.labelControlsDirective]);
 os.ui.layer.LabelControlsCtrl = function($scope, $element, $timeout) {
   os.ui.layer.LabelControlsCtrl.base(this, 'constructor');
 
+  this['pointGeometry'] = {
```
schmidtk: Is this intended for a different PR? Something for setting the label anchor?

src/os/style/style.js
```diff
@@ -57,6 +57,60 @@ os.style.DEFAULT_FEATURE_SIZE = 3;
...
+ * @type {!Array<number>}
+ */
+os.style.DEFAULT_LINE_STYLE = [];
+
+
+/**
+ * @typedef {{
+ *   name: string,
+ *   pattern: !Array<number>,
+ *   csPattern: number
```
schmidtk: This is for Cesium, and should be in `plugin.cesium`.

schmidtk:
Tests are failing due to a state schema change. Do you have a PR for `opensphere-state-schema` to add `lineDash`?

src/os/ui/layer/labelcontrols.js
welchyd: I'm not sure what that was for either, I'll remove it

src/os/ui/featureedit.js
welchyd: I just realized that I don't actually use this function ... I guess it makes sense to have for features, but I wasn't sure if you would always know if a feature is a polygon or line (ie a geometry collection)

src/os/style/style.js
welchyd: Does this go for the highlight and select styles too?
I noticed that cesium doesn't show any select or highlight style without the line dash specified in those configs

welchyd: Also - removing this seems to get the state test to pass.

welchyd:
The tests seem to be passing now ... I think removing the lineDash from the default config 'fixed' it

src/os/style/style.js
schmidtk: Yes, it should be undefined by default for all styles. That sounds like a bug introduced in these changes, given we don't currently use `lineDash` anywhere.

src/os/ui/featureedit.js
schmidtk: If the function isn't used, you can remove it. We already have `os.geo.isGeometryPolygonal` for detecting polygons, and line detection should go in the same place (if/when needed). For collections, you can iterate over geometries in the collection and determine if it contains lines/polygons. Depending on the use case you might want at least one line/polygon, all lines/polygons, or to ignore collections entirely, so that shouldn't be done in these functions.

src/plugin/cesium/sync/featureconverter.js
```diff
@@ -540,11 +540,18 @@ plugin.cesium.sync.FeatureConverter.prototype.olLineStringGeometryToCesium = fun
  */
 plugin.cesium.sync.FeatureConverter.prototype.createLinePrimitive = function(positions, context, style, opt_type) {
   var type = opt_type || 'PolylineGeometry';
-  var appearance = new Cesium.PolylineColorAppearance();
```
schmidtk: This is why the color changes stopped working. `Cesium.PolylineMaterialAppearance` uses the color on the geometry instance attributes, which is what we update in `updatePrimitive` when the color changes. The new code is using `Cesium.PolylineMaterialAppearance` with a material that has its own `color` property, so the material would need to be updated instead.

I would propose the following to fix this without having to handle both of these cases going forward:

- Change `createColoredPrimitive` to use `Cesium.MaterialAppearance` as the `appearance`, setting `options.material = Cesium.Material.fromType(Cesium.Material.ColorType, {color: color})`. This will make the default material for lines/polygons use the same type of appearance as dashed lines, so they can be updated in the same fashion.
- Remove the `color` param from `createGeometryInstance`. We will no longer be using color on the geometry instance attributes.
- Change `updatePrimitive` to update the color on the material. It should no longer need to get the attributes with `getGeometryInstanceAttributes`, and instead get the appearance/material and update that.
- Test lines, polygons, extruded polygons (load a KML with extrude), ellipses/ellipsoids, etc to make sure color/selection/highlight still works everywhere.

src/plugin/cesium/sync/featureconverter.js
```diff
@@ -540,11 +540,18 @@ plugin.cesium.sync.FeatureConverter.prototype.olLineStringGeometryToCesium = fun
...
 
   var width = this.extractLineWidthFromOlStyle(style);
   var color = this.extractColorFromOlStyle(style, true);
   color.alpha *= context.layer.getOpacity();
+  var lineDash = this.getDashPattern(style);
+
+  var appearance = new Cesium.PolylineMaterialAppearance({
+    material: Cesium.Material.fromType(Cesium.Material.PolylineDashType, {
```
schmidtk: This should only be using `Cesium.Material.PolylineDashType` when there is actually a line dash. Otherwise it should use `Cesium.Material.ColorType`, which is more efficient to render because it doesn't have to consider a dash pattern. I would suggest adding a call like `getLineMaterial = function(options)`, with `color` and `dashPattern` in the options for now. That allows us to expand support in the future without API changes or a parameter train.

src/plugin/cesium/cesiumplugin.js
```diff
@@ -40,6 +40,53 @@ goog.inherits(plugin.cesium.Plugin, os.plugin.AbstractPlugin);
...
+/**
+ * Line dash configurations for Cesium
+ * Patterns based on 16 bit number to make it look consistent between map engines
+ * @type {!Array<!plugin.cesium.styleLineDashOption>}
+ * @const
+ */
+plugin.cesium.LINE_STYLE_OPTIONS = [
+  {
+    id: 0,
+    csPattern: parseInt('1111111111111111', 2) // []
```
schmidtk: The default (no dash) pattern should be `undefined`, indicating that a dash material should not be used.

src/plugin/cesium/sync/featureconverter.js
welchyd: So keep it as a `PolylineMaterialAppearance` and vary the material?

src/plugin/cesium/sync/featureconverter.js
welchyd: Did you mean `Cesium.PolylineMaterialAppearance ` instead of `Cesium.MaterialAppearance` ?

schmidtk: No, `createColoredPrimitive` is used for creating polygons so it should just use `MaterialAppearance`. I believe all of the polyline paths have their own appearance.

src/plugin/cesium/sync/featureconverter.js
schmidtk: Yep! Let me know if you have any issues. I tinkered with it quick, and it seemed to work as described.

src/plugin/cesium/sync/featureconverter.js
welchyd: I set it as `PolylineMaterialAppearance` and it seems to be working ... tried with just `MaterialAppearance` and can't tell the different actually, so I can stick with the `MaterialAppearance`

schmidtk: Sorry my idea was incorrect, and I didn't notice because we usually render polygons with polylines so their line width will be correct on Windows.

I tested with extruded polygons, which goes down the `olPolygonGeometryToCesium` path and `MaterialAppearance` does not work there. You can test this with extruded polygons. We may have to leave these functions as they currently are, and `updatePrimitive` will need to determine how to update the color appropriately.

welchyd:
I can't seem to push to this guy anymore so I created #628 
I messed with the history to undo some changes and then wrote to it ... so that probably is keeping me from pushing here

src/os/ui/file/kml/abstractkmlexporter.js
```diff
@@ -616,6 +616,15 @@ os.ui.file.kml.AbstractKMLExporter.prototype.processPlacemark = function(element
...
     }
+
+    // Add line dash info as custom element since kml doesn't support it
+    var lineDash = this.getLineDash(item);
+    if (lineDash && Array.isArray(lineDash) && lineDash.length) {
+      var lineDashDataEl = os.xml.appendElementNS('Data', this.kmlNS, ed, undefined, {
+        'name': 'lineDash'
+      });
+      os.xml.appendElementNS('value', this.kmlNS, lineDashDataEl, JSON.stringify(lineDash));
+    }
```
jsalankey: This may cause a merge conflict. I am adding an API for handling our internal fields that contain JSON data, and this should be incorporated into that once both are merged.

src/os/ui/layer/vectorstylecontrols.js
```diff
@@ -80,6 +82,10 @@ os.ui.layer.VectorStyleControlsCtrl = function($scope) {
     this.scope['showCenterIcon'] = true;
   }
 
+  this.scope['lineDashOptions'] = os.style.LINE_STYLE_OPTIONS;
+  var name = os.style.dashPatternToOptions(this.scope['lineDash']);
+  this.scope['lineDashName'] = name ? name : this.scope['lineDashOptions'][0];
```
jsalankey: These are very confusing without type declarations. It seems like we have 3 different variables with different names that are all of type `os.style.styleLineDashOption`...?

src/plugin/cesium/sync/featureconverter.js
```diff
@@ -1892,3 +1938,16 @@ plugin.cesium.sync.FeatureConverter.prototype.setEyeOffset = function(primOffset
...
+/**
+ * Convert a style's line dash to 16 bit int
+ * @param {!ol.style.Style} style
+ * @return {number}
+ */
+plugin.cesium.sync.FeatureConverter.prototype.getDashPattern = function(style) {
+  var stroke = style.getStroke();
+  var dashPattern = stroke != null ? stroke.getLineDash() : undefined;
+  var id = /** @type {os.style.styleLineDashOption} */ (os.style.dashPatternToOptions(dashPattern)).id;
+  return plugin.cesium.LINE_STYLE_OPTIONS[id].csPattern;
```
jsalankey: This seems like a somewhat risky return statement. Is `plugin.cesium.LINE_STYLE_OPTIONS[id]` always guaranteed to be defined?

src/plugin/file/kml/kmlnodelayerui.js
```diff
@@ -1,11 +1,13 @@
 goog.provide('plugin.file.kml.KMLNodeLayerUICtrl');
+goog.provide('plugin.file.kml.KMLNodeLayerUICtrl.UIEventType');
```
jsalankey: Adding in these kinds of missing provide/requires can be risky and result in circular dependencies. The original addition of this event type enum to this class was incorrect. These sorts of event types should really be added to utility classes.

welchyd: I think I'll leave this for now

src/plugin/cesium/sync/featureconverter.js
welchyd: Okay, I've reverted those changes

PR #618, opened 2019-06-18
build(gcc): Upgrade Closure helper to use the JS compiler API.

schmidtk:


bradh:
`Error: Cannot find module '/home/travis/build/ngageoint/opensphere/--compilation_level'`

schmidtk:
Ah right, I missed the other places we're using `os-compile`. Marking this as WIP until I get those args written out as JSON via the resolver.

schmidtk:
https://github.com/ngageoint/opensphere-build-resolver/pull/38 will write test compilation args to JSON.

wallw-bits:
Looks like `src/os/xt/build-xt.js` is getting picked up by one of the globs in `karma.conf.js` and that is causing the test failure.

PR #619, opened 2019-06-18
Fix(Fix-592): Updated field sizes -flex

adamstja:
Styling changes added including "col" and "flex" for smaller window sizes. Windows addressed: "Enter coordinates" and " Add to places..."

1. In Opensphere right click in map to "Add to places...".
2. Compress window size below 680x670 PX.
3. Notice "Enter a location as DD, DMS, DDM or MGRS" no longer falls into position field. 

1. Select "Enter coordinates" from drop down in top right corner (third button from left).
2. Enter any incorrect field ( ex. "ee" or e3").
3. After incorrect coordinate has been entered, field should turn red with no size difference. 

Test link : http://callous-stem.surge.sh/


mgettman:
Looks good to me. Ready for merge.

PR #620, opened 2019-06-18
fix(style): fix merge to allow fill and outline deletions

### LINKED ISSUE #604: Support filled polygon without an outline
There are two issues at play here:

1. The KML importer [merges the default style config](https://github.com/ngageoint/opensphere/blob/master/src/plugin/file/kml/kmlparser.js#L1430) into the style config generated from the KML style. This means KML styles that do not have an outline defined will get a white outline from the default style. This line should either be removed, or improved to verify the minimum default styles to render the geometry are available in the config.
2. In `os.style.createFeatureConfig`, we merge the stroke style from the layer config if one isn't present on the feature config. Determine why this was needed and remove it if possible, otherwise make the implementation more specific to the problem it was intended to solve.

wallw-bits:
This changes the style config merging to allow for a definition of `null` to remove the parent style rather than inheriting it (`undefined`). Additionally, the `readStyle_` parser in `ol.format.KML` was replaced with a custom implementation which returns style configs rather than OpenLayers style objects.

Example: create a style config that removes a stroke rather than inheriting the default stroke:
```
var styleConfig = {
  'fill': {
    'color': 'rgba(255,0,0,1)',
  },
  'stroke': null
};
```

Testing:
Compare [PolygonStyles.txt](https://github.com/ngageoint/opensphere/files/3302618/PolygonStyles.txt) to the same file in Google Earth (making labels always on helps with identifying the correct style)

Google Earth result:
Screen Shot 2019-06-13 at 12 15 52 PM

resolves #604

### DIFF
--- a/src/os/style/style.js
+++ b/src/os/style/style.js
@@ -980,7 +980,7 @@ os.style.createFeatureConfig = function(feature, baseConfig, opt_layerConfig) {
           opt_layerConfig[os.style.StyleField.IMAGE], os.style.DEFAULT_FEATURE_SIZE);
       imageConfig['radius'] = mergedSize;
       imageConfig['scale'] = Math.max(os.style.sizeToScale(mergedSize), 0.01);
-    } else if (opt_layerConfig[os.style.StyleField.IMAGE]) {
+    } else if (imageConfig != null && opt_layerConfig[os.style.StyleField.IMAGE]) {
       // ensure the feature has an image config
       featureConfig[os.style.StyleField.IMAGE] = {};
       os.style.mergeConfig(opt_layerConfig[os.style.StyleField.IMAGE],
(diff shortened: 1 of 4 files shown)


justin-bits:
Build failure addressed by #623

src/plugin/file/kml/kmlparser.js
```diff
@@ -1396,41 +1392,42 @@ plugin.file.kml.KMLParser.reduceStyles_ = function(config, set) {
...
+  var styles = [];
   var highlightStyle = null;
 
   // style from style url
   var styleUrl = /** @type {string} */ (feature.get('styleUrl'));
   if (styleUrl) {
     var styleId = this.getStyleId(decodeURIComponent(styleUrl));
-    styleSets.push(styleId in this.styleMap_ ? this.styleMap_[styleId] : null);
+    styles.push(styleId in this.styleMap_ ? this.styleMap_[styleId] : null);
     highlightStyle = styleId in this.highlightStyleMap_ ? this.highlightStyleMap_[styleId] : null;
```
schmidtk: With the `readStyle` change, `this.highlightStyleMap_` will now be `<string, Object>`. That needs to be updated in the JSDoc and accounted for when setting the custom highlight on the feature.

src/os/style/style.js
```diff
@@ -1251,6 +1251,10 @@ os.style.createFeatureStyle = function(feature, baseConfig, opt_layerConfig) {
 os.style.mergeConfig = function(from, to) {
   for (var key in from) {
     var fval = from[key];
+    if (fval === os.object.IGNORE_VAL) {
+      continue;
+    }
+
     if (fval && typeof fval == 'object' && !(typeof fval.length == 'number')) {
       // clone objects into the target
       if (!(key in to)) {
```
schmidtk: This needs to handle `to[key]` being `null` or `undefined`. This is causing an error when hovering the polygons without a stroke, because it tries merging the highlight stroke into `null`.

src/plugin/file/kml/kmlparser.js
wallw-bits: I'm surprised the compiler didn't complain about that.

src/os/style/style.js
wallw-bits: Fixed and added tests. `undefined` was already handled.

PR #621, opened 2019-06-18
test(cypress): Skips layer-features.spec.js until failure can be inve…

justin-bits:
…stigated


justin-bits:
Fix is in #623, OBE. Closed.

PR #622, opened 2019-06-18
Increases image comparison delay

justin-bits:
Also has two timeouts for frequent failure spots due to time required to load state files.


justin-bits:
Build failure addressed by #623

wallw-bits:
Closed because of #626

PR #623, opened 2019-06-18
test(cypress): Fixes issue with intermittent click failure

justin-bits:


cypress/integration/regression-tests/layers-dialog/layer-features.spec.js
```diff
@@ -91,10 +91,12 @@ describe('Feature grid', function() {
       });
     });
 
-    cy.get(os.layerFeaturesDialog.DESCRIPTION_CELL).should('be.visible');
-    cy.get(os.layerFeaturesDialog.DESCRIPTION_CELL).contains('Show');
-    cy.get(os.layerFeaturesDialog.DESCRIPTION_CELL).contains('Show').click();
-    cy.get(os.descriptionInfoDialog.DIALOG, {timeout: 5000}).should('be.visible');
+    cy.wait(100); // Link not immediately ready when dialog loads
```
bradh: An option (instead of a fixed wait) would be to check if the link has the right characteristics (e.g. an `<a href=...>`). If that ends up being worth-while, could refactor it to a common command.

wallw-bits: Correct. I'm going to look into something like https://github.com/NoriSte/cypress-wait-until to help with that.

bradh: Given the breakage on travis, can this go in as an interim fix?

wallw-bits: I prefer #626 since we are having other issues getting those tests to run consistently in a local environment.

wallw-bits:
Closed because of #626

PR #624, opened 2019-06-19
fix(Fix-616): Drag Cursor

adamstja:
Drag of cursor has been removed to select labels. Only selects input fields.

1. Open any dialog box.
2. Attempt to highlight a label (or multiple labels).
3. Labels should not be highlighted globally.

windy-kite.surge.sh

### DIFF
--- a/scss/os/_overrides_bootstrap.scss
+++ b/scss/os/_overrides_bootstrap.scss
@@ -120,6 +120,6 @@ body {
 }
 
 // Blocks cursor select outside of input field
-* label {
+div {
   user-select: none;
 }
--- a/scss/os/_overrides_bootstrap.scss
+++ b/scss/os/_overrides_bootstrap.scss
@@ -118,3 +118,8 @@ body {
     cursor: pointer;
   }
 }
+
+// Blocks cursor select outside of input field
+* label {
+  user-select: none;
+}
--- a/index-template.html
+++ b/index-template.html
@@ -6,7 +6,7 @@
 <link rel="icon" href="@version@images/favicon.ico?v=1" sizes="16x16 24x24 32x32 48x48 64x64" type="image/x-icon"/>
 <!--VENDOR_CSS-->
 <!--APP_CSS-->
-<div id="ng-app" class="unselectable os-app d-flex flex-fill" ng-init="version='@packageVersion@';versionPath='@version@'">
+<div id="ng-app" class="ol-unselectable os-app d-flex flex-fill" ng-init="version='@packageVersion@';versionPath='@version@'">
   <div ng-view></div>
 </div>
 <noscript>Javascript is disabled in your browser. Turn it on and refresh the page.</noscript>
--- a/views/layers.html
+++ b/views/layers.html
@@ -1,4 +1,4 @@
-<div class="d-flex flex-column flex-fill u-user-select-none">
+<div class="d-flex flex-column flex-fill">
   <div class="form-row align-items-center flex-shrink-0 mt-1">
     <div class="col-auto col-form-label">
       Group By
--- a/views/menu/actionmenuitemlist.html
+++ b/views/menu/actionmenuitemlist.html
@@ -11,7 +11,7 @@
       <!-- action item -->
       <div class="d-flex">
         <div class="text-truncate flex-fill d-inline">
-          <a class="unselectable" ng-if="!(isSeparator || isSeparatorHeader)">
+          <a class="ol-unselectable" ng-if="!(isSeparator || isSeparatorHeader)">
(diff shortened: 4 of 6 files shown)


wallw-bits:
The issue still exists in your demo:

Screen Shot 2019-06-20 at 10 30 07 AM

Additionally, the "scope" value in the commit message (in parens) should be the package in which you were working. "fix(scss): ..." would have worked nicely. If you are unsure what to put, you can skip it entirely. The issue reference should go at the end of the message:

```
fix(scss): some title

some other specifics

fixes #614
```

The formatting on those helps keep some of our automated notes/logs nice.

scss/os/_overrides_bootstrap.scss
```diff
@@ -118,3 +118,8 @@ body {
     cursor: pointer;
   }
 }
+
+// Blocks cursor select outside of input field
+* label {
+  user-select: none;
+}
```
wallw-bits: What about windows, tabs, title bars, or every other non-form element?

adamstja:
?Thanks for the clarification on the commit message. I will be sure to follow this convention.


I'm working to clean up the remainder of the highlight issue. Should be done before COB.

________________________________

wallw-bits:
This is better but we still need the Layer/Feature description window content and also the description panel in the Add Data window to be selectable/copyable.

gregroyal:
Dont add bootstrap overrides really ever. Those are extremely core styles and should be avoided unless you know exactly what you are doing

Options:
- Do the global no select how opensphere used to be (unselectable is already there but doesnt work. ol-unselectable does however)
- Use the already existing util class u-user-select-none where you dont want selection (best choice, most work)

adamstja:
> Dont add bootstrap overrides really ever. Those are extremely core styles and should be avoided unless you know exactly what you are doing
> 
> Options:
> 
> * Do the global no select how opensphere used to be (unselectable is already there but doesnt work. ol-unselectable does however)
> * Use the already existing util class u-user-select-none where you dont want selection (best choice, most work)

Sure, I will go through and do this. Might take a little bit of time.

welchyd:
Bringing back the `ol-unselectable` seems to be pretty effective to me
Only thing that's standing out right now is the menus/drop downs, I'll keep checking though
I'm able to click and drag from the submenu divider bar or any hint text (like ctrl+o) and get things to highlight

DropDown

views/window/window.html
```diff
@@ -1,4 +1,4 @@
-<div class="js-window modal d-flex c-window o-window" ng-class="{'active': active}">
+<div class="js-window modal d-flex c-window o-window u-user-select-none" ng-class="{'active': active}">
```
gregroyal: dont add this to window. all windows wont be selectable. I prefer everything to be selectable and some stuff not, rather than nothing be selectable and some stuff is.

wallw-bits: If it is simpler to define everything as unselectable and then undo it for the few items that should be, why is that not the better option?

adamstja: @wallw-bits That was the original plan. I have gotten push back on changing CSS.

I'd like to know if I should continue adding u-user-select-none to individuals divs, or if I should declare rules within CSS. 

If you would prefer I do the CSS rules, let me know of a few areas to go off of. (Add layer description etc)

gregroyal: Because in normal webpages everything is selectable by default. Its abnormal to have something not selectable. Its better to just make things you dont want selectable, not selectable. So you dont have users questioning why something isnt selectable

I dont even see why this ticket is even needed. I see no problem with the users selecting the text Areas on the areas tab. If the reason is when they are shift clicking on the map or something and stuff gets selected, that issue should be addressed instead of the bandaid fix of just making everything unselectable

gregroyal: How about when the map is moused over, call
window.getSelection().removeAllRanges()

This will remove selected text

wallw-bits: That's no good because I could have a legitimate selection which that function would remove.

gregroyal: It would only remove if you are mousing over the map, at which you've already left the area you are trying to select. It does fix the issue of trying to select some text and all the text gets selected which is not expected. There might be a better fix too, I think thats where the effort should be applied

Just because there is a big map in the background, doesn't mean it gets to break rules of the internet

Just checkout google maps https://www.google.com/maps/place/Washington,+DC/@38.8935755,-77.0846151,12z/data=!3m1!4b1!4m5!3m4!1s0x89b7c6de5af6e45b:0xc2524522d4885d2a!8m2!3d38.9071923!4d-77.0368707
They have no problem with allowing selection while having a big map in the background

welchyd: Google Maps doesn't have a drop target for text though - which is what makes the text selection a nuisance.

PR #633, opened 2019-06-24
feat(states): Add timeline lock to states

### LINKED ISSUE #597: Timeline lock not saved to state
Save the animation lock to saved states as a part of Time.
THIN-13358

smithtb:
THIN-13358
Resolves #597 

This requires ngageoint/opensphere-state-schema#11 to work

### DIFF
--- a/package.json
+++ b/package.json
@@ -267,7 +267,7 @@
     "opensphere-build-docs": "^1.0.0",
     "opensphere-build-index": "^2.1.0",
     "opensphere-build-resolver": "^5.2.0",
-    "opensphere-state-schema": "^2.3.0",
+    "opensphere-state-schema": "^2.4.0",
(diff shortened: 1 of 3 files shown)


welchyd:
Looks like that fixes the test!
Another test is failing though - setting the lock changes the timestamps

smithtb:
Go to http://measly-bread.surge.sh/

Test steps:
1. Set up the timeline and take note of the values. 
2. Enable the "lock" on the timeline with the lock button.
3. Save the state (be sure to include time)
4. Import the state and clear application data
5. Verify that the values from the state match what the timeline was like before the state was saved
6. Try some different combinations (with skips, etc.)

arwatts16:
I tested it and everything worked exactly as I expected it to.

mgettman:
Looks good!

PR #634, opened 2019-06-25
chore: Upgrade eslint to 6.0.0

bradh:


config/settings.json
```diff
@@ -171,4 +201,4 @@
       "type": "trueNorth"
     }
   }
-}
+}
```
wallw-bits: Were the non-whitespace changes here accidental?

bradh: The ogcapi part was an inadvertent commit. I thought I reverted it all, but now see I missed some.

metrics is just whitespace:
```
    "metrics": {
      "enabled" : true
      "enabled": true
    },
```

bradh: the ogcapi part should now be gone.

PR #636, opened 2019-06-25
fix(ScreenOverlay): Screen overlays do not have black border around t…

### LINKED ISSUE #629: Black border on screen overlays
When a screen overlay is present, it has a black border - before we upgraded to bootstrap 4 there was no border on screen overlays and the top margin looks strange because of it.
Please remove the border.

OverlayBorder

Here's an example kmz with a screen overlay: https://sites.google.com/site/kmltouring/code-examples/simple-screen-overlay/GE%20Logo%20bottom%20left.kmz?attredirects=0&d=1

lilysikes:
…hem.

resolves #629

### DIFF
--- a/src/os/ui/screenoverlay.js
+++ b/src/os/ui/screenoverlay.js
@@ -64,7 +64,8 @@ os.ui.launchScreenOverlay = function(options) {
     'show-hide': !!options.showHide,
     'show-close': !!options.showClose,
     'no-scroll': true,
-    'overlay': true
+    'overlay': true,
+    'border': '0'
   };
 
   var template = '<screenoverlay></screenoverlay>';
--- a/src/os/ui/screenoverlay.js
+++ b/src/os/ui/screenoverlay.js
@@ -65,7 +65,7 @@ os.ui.launchScreenOverlay = function(options) {
     'show-close': !!options.showClose,
     'no-scroll': true,
     'overlay': true,
-    'border': '0'
+    'border': false
   };
 
   var template = '<screenoverlay></screenoverlay>';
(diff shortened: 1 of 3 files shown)


lilysikes:
Live Link: sour-scissors.surge.sh

lilysikes:
Test Steps: 
1. Download google screen overlay kmz from here: https://sites.google.com/site/kmltouring/code-examples/simple-screen-overlay/GE%20Logo%20bottom%20left.kmz?attredirects=0&d=1
2. Import kmz file into Opensphere
3. Move the screen overlay to the globe
4. The screen overlay should not have a border around the image.

src/os/ui/screenoverlay.js
```diff
@@ -64,7 +64,8 @@ os.ui.launchScreenOverlay = function(options) {
     'show-hide': !!options.showHide,
     'show-close': !!options.showClose,
     'no-scroll': true,
-    'overlay': true
+    'overlay': true,
+    'border': '0'
```
wallw-bits: `'0'` is a poor value here. Another developer might assume that you mean to specify the border width (in which case an actual number would be better). However, it appears that you mean for this to be a boolean and this should be `false`.

views/window/window.html
```diff
@@ -1,5 +1,5 @@
 <div class="js-window modal d-flex c-window o-window" ng-class="{'active': active}">
-  <div class="modal-content d-flex flex-column flex-fill" ng-class="{'bg-transparent': overlay}">
+  <div class="modal-content d-flex flex-column flex-fill" ng-class="{'bg-transparent': overlay, 'border-0': border == '0'}">
```
wallw-bits: Once the above comment is fixed, the expression here can just be `!border`.

mgettman:
Good to go

PR #638, opened 2019-06-26
fix(PlacesExport): If a place has an icon with rotation the rotation …

### LINKED ISSUE #603: Icon Rotation Doesn't Export with Places
When exporting Places, any place with an icon and rotation doesn't import with the original rotation - it is always set to 0.

lilysikes:
…is exported.

resolves #603

### DIFF
--- a/src/plugin/file/kml/kmlexport.js
+++ b/src/plugin/file/kml/kmlexport.js
@@ -52,20 +52,21 @@ plugin.file.kml.export.getBalloonOptions = function(feature) {
  * Get the icon rotation column for an OpenLayers feature.
  *
  * @param {ol.Feature} feature The feature.
- * @return {string|null|undefined} The column.
+ * @return {string|undefined} The column.
  */
 plugin.file.kml.export.getRotationColumn = function(feature) {
+  var rotColumn;
   if (feature) {
+    rotColumn = /** @type {string|undefined} */ (feature.get(os.style.StyleField.ROTATION_COLUMN));
+
     var layerConfig = os.style.getLayerConfig(feature);
-    if (layerConfig) {
-      var rotColumn = layerConfig[os.style.StyleField.ROTATION_COLUMN];
-      if (!layerConfig[os.style.StyleField.REPLACE_STYLE]) {
-        rotColumn = rotColumn || /** @type {string} */ (feature.get(os.style.StyleField.ROTATION_COLUMN));
-      }
-      return rotColumn;
+    if (!rotColumn || layerConfig && layerConfig[os.style.StyleField.REPLACE_STYLE]) {
+      rotColumn = layerConfig[os.style.StyleField.ROTATION_COLUMN];
     }
   }
-  return undefined;
+
+  // don't return an empty string
+  return rotColumn;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


lilysikes:
Live Link: http://conscious-kitten.surge.sh/

lilysikes:
Test Steps:
1. Right click anywhere on the globe and click 'Save to Places' from the drop down menu
2. From the add place dialogue open the 'Style Options' menu
3. Click on the drop down menu next to style and choose icon
4. Click the button next to the drop down menu and choose an icon that isn't perfectly symmetrical
5. Type in a degree of rotation in the rotation text field right above the 'Style Options' menu
6. Click 'OK' at the bottom of the 'Add Place' menu
7. Click on the Places tab from the layers menu
8. Select the place you just created and click 'Export'
9. Delete the place you made
10. From the Places tab hit the import button and select the file you exported.
11. Hit 'OK'
12. The place you created with the icon with the correct rotation should be on the globe
13. Right click the place and select 'Edit Place'
14. The rotation text field should have the correct rotation
15. Delete the place again and drag and drop or use the 'Add Data' button to import the file not from Places
16. The place should have the correct icon with the correct rotation

src/plugin/file/kml/kmlexport.js
```diff
@@ -57,8 +57,14 @@ plugin.file.kml.export.getBalloonOptions = function(feature) {
...
-    if (layerConfig && layerConfig[os.style.StyleField.SHOW_ROTATION]) {
-      return layerConfig[os.style.StyleField.ROTATION_COLUMN];
+    if (layerConfig) {
+      if (layerConfig[os.style.StyleField.REPLACE_STYLE]) {
+        if (layerConfig[os.style.StyleField.SHOW_ROTATION]) {
+          return layerConfig[os.style.StyleField.ROTATION_COLUMN];
+        }
+      } else if (feature.get(os.style.StyleField.SHOW_ROTATION)) {
+        return /** @type {string} */ (feature.get(os.style.StyleField.ROTATION_COLUMN));
+      }
```
wallw-bits: This only uses the layer config when `REPLACE_STYLE` is true. That is not correct. I think you want something more like this:
```
var rotColumn = layerConfig[...];
if (!layerConfig[REPLACE_STYLE]) {
  rotColumn = rotColumn || feature.get(...);
}
```

welchyd: Doesn't that ignore if the show rotation box is checked or not though?
The kml exports with rotation whether or not that box is checked now ... I guess that could be tracked separately (ie remember the rotation column selection even if it isn't used)

wallw-bits: > Doesn't that ignore if the show rotation box is checked or not though?

It is just pseudocode. You'll need to add that in too.

schmidtk: This function is only concerned with getting the column, and doesn't need to consider the `SHOW_ROTATION` value. Code using this function can choose whether or not to do something with that column based on `SHOW_ROTATION`.

schmidtk:
The core problem here was that the rotation column was not being included in KML export. Places worked around this because `os.ui.FeatureEditCtrl.updateFeatureStyle` was setting icon rotation as long as the `BEARING` column was set on the feature.

I made the following changes:
- Always export the rotation column and show rotation flag to KML.
- Add the appropriate supporting columns in `os.ui.FeatureEditCtrl.updateFeatureStyle`, so old places will export properly.
- Clean up logic in `plugin.file.kml.export.getRotationColumn`.

schmidtk:
Reworded commits to meet style guidelines.

PR #642, opened 2019-06-27
feat(ClearPlaces): The clear menu now can hide places and annotations.

### LINKED ISSUE #632: Annotations/places should be added to the clear menu
Please add the option to remove annotations, all saved places, or both from this menu:

Screen Shot 2019-06-24 at 9 22 54 AM

Do what we do for areas, _turn them off, don't actually remove them_. The most common action with the clear button is to check all and we don't want people to accidentally remove all of the places they have painstakingly created.

lilysikes:
resolves #632

### DIFF
--- a/src/os/mainctrl.js
+++ b/src/os/mainctrl.js
@@ -159,7 +159,6 @@ goog.require('plugin.osm.nom.NominatimPlugin');
 goog.require('plugin.overview.OverviewPlugin');
 goog.require('plugin.params.ParamsPlugin');
 goog.require('plugin.pelias.geocoder.Plugin');
-goog.require('plugin.places.PlacesHide');
 goog.require('plugin.places.PlacesPlugin');
 goog.require('plugin.position.PositionPlugin');
 goog.require('plugin.suncalc.Plugin');
@@ -275,8 +274,6 @@ os.MainCtrl = function($scope, $element, $compile, $timeout, $injector) {
       'Clear all spatial query areas'));
   os.ui.clearManager.addEntry(new os.ui.clear.ClearEntry('states', 'States', os.ui.state.cmd.StateClear,
       'Deactivate all states'));
-  os.ui.clearManager.addEntry(new os.ui.clear.ClearEntry('places', 'Places/Annotations', plugin.places.PlacesHide,
-      'Clear all Places and Annotations'));
 
   // set up search
   var searchManager = os.search.SearchManager.getInstance();
--- a/src/os/mainctrl.js
+++ b/src/os/mainctrl.js
@@ -159,6 +159,7 @@ goog.require('plugin.osm.nom.NominatimPlugin');
 goog.require('plugin.overview.OverviewPlugin');
 goog.require('plugin.params.ParamsPlugin');
 goog.require('plugin.pelias.geocoder.Plugin');
+goog.require('plugin.places.PlacesHide');
 goog.require('plugin.places.PlacesPlugin');
 goog.require('plugin.position.PositionPlugin');
 goog.require('plugin.suncalc.Plugin');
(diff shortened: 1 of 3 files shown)


lilysikes:
Live Link: http://legal-teaching.surge.sh/

lilysikes:
Test Steps:
1. Right click the globe and select 'Save to Places'
2. Click 'OK'
3. Right click the globe somewhere else and select 'Create Annotation'
4. Click 'OK'
5. At the top left of the screen hit the red 'x' button
6. Select the 'Places/Annotations' checkbox
7. Hit 'OK'
8. The place and annotation should be gone from the globe
9. Click on the Places tab from the Layer menu and make sure the annotation and place are still there just unchecked
10. Press CTRL+Z or CMD+Z and the place and annotation should reappear on the globe and they should be checked on the places tab
11. Now uncheck either the annotation or the place from the Places tab
12. Repeat steps 5-7
13. The place you left checked should be removed from the globe
14. Press CTRL+Z or CMD+Z and only the place you had checked should return

src/os/mainctrl.js
```diff
@@ -273,6 +274,8 @@ os.MainCtrl = function($scope, $element, $compile, $timeout, $injector) {
       'Clear all spatial query areas'));
   os.ui.clearManager.addEntry(new os.ui.clear.ClearEntry('states', 'States', os.ui.state.cmd.StateClear,
       'Deactivate all states'));
+  os.ui.clearManager.addEntry(new os.ui.clear.ClearEntry('places', 'Places/Annotations', plugin.places.PlacesHide,
```
schmidtk: The Places plugin should be adding this in its `init` method, so it doesn't show up in the Clear menu when the plugin is disabled.

src/plugin/places/placeshidecmd.js
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * @inheritDoc
+ */
+plugin.places.PlacesHide.prototype.revert = function() {
+  this.state = os.command.State.REVERTING;
+  var rootNode = plugin.places.PlacesManager.getInstance().getPlacesRoot();
+  if (rootNode) {
+    var source = rootNode.getSource();
+    if (source) {
+      if (!this.visibleNodes_.length == 0) {
```
schmidtk: This expression isn't quite what you're shooting for. I think you want `this.visibleNodes_.length != 0` instead.

Currently it's converting the length to a boolean (`!this.visibleNodes_.length`) and comparing the result to 0. If the length is non-zero, the expression is essentially `false == 0`, which happens to be true and the check passes.

src/plugin/places/placeshidecmd.js
```diff
@@ -0,0 +1,129 @@
...
+ */
+plugin.places.PlacesHide.prototype.state = os.command.State.READY;
+
+
+/**
+ * @inheritDoc
+ */
+plugin.places.PlacesHide.prototype.execute = function() {
+  this.state = os.command.State.EXECUTING;
+  var rootNode = plugin.places.PlacesManager.getInstance().getPlacesRoot();
```
schmidtk: This should clear `this.visibleNodes_` first, or it may result in expected behavior with undo/redo. It will also infinitely grow the list on each `execute` call.

src/plugin/places/placeshidecmd.js
```diff
@@ -0,0 +1,129 @@
...
+  if (rootNode) {
+    var source = rootNode.getSource();
+    if (source) {
+      var features = source.getFeatures();
+      if (features) {
+        this.storeVisibleChildren(rootNode);
+        source.hideFeatures(features);
```
schmidtk: You're already changing the node state in `storeVisibleChildren`, so you shouldn't have to do this. `plugin.file.kml.KMLSource.prototype.updateVisibilityFromNodes` will hide the features automatically based on the node state change.

PR #643, opened 2019-06-28
fix(LineDash): The border drop down on places is now consistent acros…

### LINKED ISSUE #639: Dash border select doesn't look consistent
When picking a border style, the select doesn't show the style consistently between browsers and operating systems. Find a way to give it a consistent look. This could be something like using a select2 and svg elements to show the line dashes

DashSelectStyles

lilysikes:
…s browsers.

resolves #639

### DIFF
--- a/scss/os/_vectorstylecontrol.scss
+++ b/scss/os/_vectorstylecontrol.scss
@@ -1,5 +1,5 @@
 .c-vectorstylecontrol__borderdash {
   fill: none;
-  stroke: $gray-500;
+  stroke: $input-color;
   stroke-width: 3;
 }
--- a/scss/os/_vectorstylecontrol.scss
+++ b/scss/os/_vectorstylecontrol.scss
@@ -1,0 +1,5 @@
+.c-vectorstylecontrol__borderdash {
+  fill: none;
+  stroke: $gray-500;
+  stroke-width: 3;
+}
--- a/src/os/ui/layer/vectorstylecontrols.js
+++ b/src/os/ui/layer/vectorstylecontrols.js
@@ -64,12 +64,11 @@ os.ui.Module.directive('vectorstylecontrols', [os.ui.layer.vectorStyleControlsDi
  *
  * @param {!angular.Scope} $scope The Angular scope.
  * @param {!angular.JQLite} $element
- * @param {!angular.$timeout} $timeout
  * @extends {goog.Disposable}
  * @constructor
  * @ngInject
  */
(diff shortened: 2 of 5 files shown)


lilysikes:
Live Link: http://parallel-star.surge.sh/

lilysikes:
Test Steps: 
1. Open in a chrome browser
2. Right click on the globe and click save to places
3. Under the style tab click the border dropdown menu
4. Now open the link in Firefox
5. Repeat Step 2 and 3
6. Compare the lines in the dropdown, they should match

bradh:
Version 75.0.3770.100 (Official Build) (64-bit) on Win 10:

image

Firefox Quantum 67.0.4 (64-bit) on Win 10:
image

src/os/ui/layer/vectorstylecontrols.js
```diff
@@ -160,3 +173,23 @@ os.ui.layer.VectorStyleControlsCtrl.prototype.onLineDashChange = function() {
...
+ * Search result formatter. The select is actually storing the ID of each
+ * descriptor. This function allows us to display the actual layer title.
+ * @param {Object} item
+ * @param {angular.JQLite} ele
+ * @return {string|angular.JQLite}
+ * @private
+ */
+os.ui.layer.VectorStyleControlsCtrl.prototype.select2Formatter_ = function(item, ele) {
+  if (item) {
+    var val = '<svg height="2" width="80"><g fill="none" stroke="white" stroke-width="3"><path ';
```
bradh: I think the hard-coded `stroke` colour and possibly the `d` parameter might be theming problems. For example, if you put OpenSphere into "Light" theme, you can only see the line style for the one you have selected (because the line style and background are both white otherwise). I don't remember what they are though, sorry.

If you were changing the `d`, you could break out the magic numbers as symbolic variables (or constants if applicable).

welchyd: The `d` seems to be okay to keep
I moved the stroke fill to a class so it updates with the theme
Also the search box didn't make sense so that has been removed

scss/os/_vectorstylecontrol.scss
```diff
@@ -0,0 +1,5 @@
+.c-vectorstylecontrol__borderdash {
+  fill: none;
+  stroke: $gray-500;
```
schmidtk: I would suggest using `$input-color` here, to match the font color used for other form inputs.

PR #644, opened 2019-06-28
Add Image Rotation Support

wallw-bits:
This adds image rotation support and then uses it in KML GroundOverlays to address some of our issues with displaying those.

Additionally, if the GroundOverlay contains an unsupported `LatLonQuad` value, then a warning is issued. I can write a ticket for further support of that. It requires non-affine transforms which are not supported in `CanvasRenderingContext2D` but could be done by us since the image would only need to be redrawn once.

I have not decided what warnings to show or not show regarding the lack of projection information for the raster image. Right now it is simply assumed to be in the application projection (probably a bad assumption).

Test Files (test in 2D/3D and in multiple projections):
[GroundOverlays_etna_rotation_compare.txt](https://github.com/ngageoint/opensphere/files/3341083/GroundOverlays_etna_rotation_compare.txt)
[GroundOverlay_unsupported_quad.txt](https://github.com/ngageoint/opensphere/files/3341084/GroundOverlay_unsupported_quad.txt)
[GroundOverlay_rotation_test.txt](https://github.com/ngageoint/opensphere/files/3341085/GroundOverlay_rotation_test.txt)

- [x] Extent reduction from the KML tree is not working as expected, so I'm going to see if I can clean that up here since it is affecting the ease of viewing GroundOverlay items.
- [x] 3D Image Layer Sync

### DIFF
--- a/src/plugin/file/kml/kml.js
+++ b/src/plugin/file/kml/kml.js
@@ -15,6 +15,7 @@ goog.require('ol.style.IconOrigin');
 goog.require('ol.style.Style');
 goog.require('ol.xml');
 goog.require('os.data.RecordField');
+goog.require('os.geo');
 goog.require('os.mixin');
 goog.require('os.mixin.polygon');
 goog.require('os.object');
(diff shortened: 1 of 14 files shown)


src/plugin/file/kml/kml.js
```diff
@@ -446,17 +447,26 @@ plugin.file.kml.readLatLonBox_ = function(node, objectStack) {
...
-        ol.extent.extendCoordinate(extent, coordinate);
-      });
+      var extent = coordinates.reduce(function(extent, coord) {
+        ol.extent.extendCoordinate(extent, coord);
+        return extent;
+      }, ol.extent.createEmpty());
 
       var targetObject = /** @type {Object} */ (objectStack[objectStack.length - 1]);
-      targetObject['extent'] = extent;
+      if (os.geo.isRectangular(coordinates, extent)) {
```
schmidtk: `os.geo.isRectangular` won't work for `LatLonQuad` coordinates because they aren't a closed polygon. They specify the corners as `[bottom left, bottom right, top right, top left]`. Push the first coordinate onto the array before the check?

wallw-bits:
> When animating ground overlays with a timestamp in 3D, the overlay flickers on each timeline change.

This is an unrelated issue with `KMLSource`'s override of `updateAnimationOverlay()`.

> When switching projections in 3D, the ground overlay image clips the globe.

Ignoring all the other projection assumptions we have already made for this (... ew), I am not getting that artifact. Your screenshot that you sent me (which you should post) looks like raster reprojection from OpenLayers, which we should have off by default. Is that enabled in your environment?

src/plugin/file/kml/kml.js
wallw-bits: Should be fixed.

schmidtk:
I'm seeing the artifacts in both Chrome and Firefox on macOS, in both debug and compiled builds. It doesn't seem to matter which projection I start in (only trying 3857 and 4326), I get the same result. I do not have raster reprojection enabled.

Here's the screenshot:

Screen Shot 2019-07-05 at 2 17 41 PM

schmidtk:
One more issue in 3D mode:

If a ground overlay isn't within the view extent when loaded, it will not be visible on the globe when the view extent does contain the overlay.

wallw-bits:
The latest commit adds a synchronizer specific to static image sources. This essentially just loads and adds the primitive once and controls visibility only with `primitive.show`, which makes it much slicker when animating and also means that it is no longer concerned with the view extent.

src/plugin/file/kml/kmlparser.js
```diff
@@ -1195,38 +1195,46 @@ plugin.file.kml.KMLParser.prototype.readGroundOverlay_ = function(el) {
...
+      node.setImage(image);
 
-    return node;
-  } else {
-    return null;
+      return node;
+    } else if (obj['error']) {
+      var msg = obj['error'].toString();
+      goog.log.warning(plugin.file.kml.KMLParser.LOGGER_, msg);
+      os.alertManager.sendAlert(msg, os.alert.AlertEventSeverity.WARNING);
```
jsalankey: This alert is in a function called to parse each `<GroundOverlay>` element, and could spam the user with repeated warnings if there are several non-rectangular LatLonQuad's present. This is probably an edge case, but it would be better to collect all the instances we can't handle and then fire one alert at the end.

wallw-bits: That logic is in the `AlertManager` itself. It defaults to no more than 5 duplicate messages, or we can call `sendAlertOnce` here instead for a single one.

jsalankey: That's fair. I didn't think about the fact that the errors were all exact dupes. I still don't like depending on the AlertManager's dedupe logic in cases where we can control it because it's still super ugly when it happens. That should be mostly a fallback for unanticipated cases.

PR #650, opened 2019-07-05
feat(annotation): Add label-like abilities to annotations

### LINKED ISSUE #535: Popout Labels
Investigate making labels as a popout annotation or text on the globe
The idea would be that the user can choose to show a label or annotation on a place, ie give a popout option for labels that would show it as an annotation instead
Also see if the label position can be configured either by dragging or by providing some position options with a live preview (ie clock position + inside and out)
Additionally make sure the label doesn't over it's shapes defining boundaries
THIN-13367
3 story points

MarcelloNicoletti:
This adds label-like style and text binding to annotations so they can be used as moveable labels.

Known Issue: Using the label controls from the places accordion does not update the annotation text in "label" mode.

fix #535

### DIFF
--- a/src/os/annotation/abstractannotationctrl.js
+++ b/src/os/annotation/abstractannotationctrl.js
@@ -20,26 +20,27 @@ os.annotation.UI_TEMPLATE =
   '<div class="c-annotation u-hover-container" ' +
     'ng-class="{\'c-annotation__editing\': ctrl.editingDescription || ctrl.editingName}">' +
     '<svg class="c-annotation__svg">' +
-      '<path ng-style="{ fill: ctrl.options.showDescription ? ctrl.options.bodyBG : ctrl.options.headerBG}" />' +
+      '<path ng-style="{ fill: ctrl.options.showDescription ? ctrl.options.bodyBG : ctrl.options.headerBG }"/>' +
     '</svg>' +
-    '<div class="u-card-popup position-absolute text-right animate-fade u-hover-show"' +
+    '<div class="u-card-popup position-absolute text-right animate-fade u-hover-show" ' +
         ' ng-show="ctrl.options.editable">' +
-      '<button class="btn btn-sm btn-outline-primary border-0 bg-transparent"' +
-          'title="Hide text box"' +
+      '<button class="btn btn-sm btn-outline-primary border-0 bg-transparent" ' +
+          'title="Hide text box" ' +
           'ng-click="ctrl.hideAnnotation()">' +
         '<i class="c-glyph fa fa-fw fa-comment"></i>' +
       '</button>' +
-      '<button class="btn btn-sm btn-outline-primary border-0 bg-transparent"' +
-          'title="Edit text box"' +
+      '<button class="btn btn-sm btn-outline-primary border-0 bg-transparent" ' +
+          'title="Edit text box" ' +
           'ng-click="ctrl.launchEditWindow()">' +
         '<i class="c-glyph fa fa-fw fa-pencil"></i>' +
       '</button>' +
     '</div>' +
-    '<div class="js-annotation c-window card h-100">' +
-      '<div class="card-header flex-shrink-0 text-truncate px-1 py-0 js-annotation__header" title="{{ctrl.name}}"' +
-          'ng-show="ctrl.options.showName" ' +
+    '<div class="js-annotation c-window card h-100" ' +
+      'ng-class="!ctrl.options.showBackground && \'bg-transparent u-border-show-on-hover u-text-stroke\'">' +
+      '<div class="card-header flex-shrink-0 text-truncate px-1 py-0 js-annotation__header" title="{{ctrl.name}}" ' +
+          'ng-show="ctrl.options.showName && ctrl.options.showBackground" ' +
           'ng-class="!ctrl.options.showDescription && \'h-100 border-0\'" ' +
-          'ng-style="{background: ctrl.options.headerBG }" ' +
+          'ng-style="{ background: ctrl.options.headerBG }" ' +
           'ng-dblclick="ctrl.editName()">' +
         '<div ng-show="!ctrl.editingName">{{ctrl.name}}</div>' +
         '<div class="form-row p-1" ng-if="ctrl.editingName">' +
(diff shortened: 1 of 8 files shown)


MarcelloNicoletti:
The reasoning behind only showing the description is that in this mode the description text matches what the label text would be. This does require editing using the label controls. (see also the know issue in the OP) If it isn't very clear I'd like some help with the help text to make it more clear.

schmidtk:
So I don't think we should be blending the concepts of labels and annotations, as I can see that being confused to the user. There should be a clear distinction between the two, or it will be unintuitive when each should be used. From out point of view, labels are a performant way to display limited metadata for a lot of features. Annotations are a way to show formatted text, images, etc, but are not efficient for displaying in bulk.

With respect to the new annotation option, it's unnecessarily limiting to the user to only display the description when the background is hidden. The behavior was confusing to me, and I would expect users would be equally confused. They should be able to hide the border/background and show whichever portion of the annotation they want.

If we want to add a way to pre-populate an annotation with label text, that can be easily accommodated without imposing this limitation on the show/hide background feature.

CC @welchyd

welchyd:
This started off as a request to allow control over the position of a label on a Place.
However we found that the position is done differently between 2D and 3D - plus another request was made to be able to drag the label around to set its' position.
Annotations seemed like reasonable to meet those requests since it is consistent between 2D and 3D and has the drag positioning capability. Also, since this is specific to a Place it is not intended to be used in a bulk scenario like a label might be.

I'm wondering if maybe we're trying to do too many things with one checkbox? Perhaps we could have separate options for hiding the header background, hiding the description background, and using the label text for the description?

schmidtk:
Yes definitely too many things, given some of them are already controlled by other checkboxes.

I don't see the utility in separately hiding the two BG's... that looks odd in practice. I think this checkbox should simply toggle the BG/border for the entire annotation, nothing further.

> using the label text for the description

The description should support inserting metadata fields, which is something we've discussed previously because it's related to the [KML BalloonStyle](https://developers.google.com/kml/documentation/kmlreference#balloonstyle) specification. That spec allows insert metadata fields into the description using `$[fieldName]` in the text.

welchyd:
Okay, so one checkbox to hide the background - if the user wants to show only the description or header then those options are already present

> That spec allows insert metadata fields into the description using $[fieldName] in the text.

Would it be okay to translate between the label selection and the kml spec?

schmidtk:
> Would it be okay to translate between the label selection and the kml spec?

Could you elaborate on this part? I'm not sure I'm tracking.

welchyd:
What I was going for is to have a checkbox that takes the label selection and turns it into kml markup, overriding the description to make it look like the label would otherwise.

I guess another way to go about this would be to extend the toast ui to allow additional markup for referencing a column and give the user more control over what it looks like.

schmidtk:
I would suggest avoiding any relationship between Labels and Annotations. I have a strong feeling that will only lead to user confusion, and some unfortunate code to manage that relationship. It would be ideal if the description field had its own built-in support for displaying metadata values from the feature.

schmidtk:
@MarcelloNicoletti FYI, this has merge conflicts to resolve if you're not already aware.

MarcelloNicoletti:
I'm aware. But I wanted to work on getting to my goals before mucking it up with the conflict

welchyd:
So the direction we'll take here is to have the Show Background checkbox toggle the header and description backgrounds to be visible or not, and when they are not visible it'll show the outline on hover plus add a shadow to make the text visible. (This means taking out the connection with the label selection)
Adding support for the markup text is a bigger change and will be done separately

MarcelloNicoletti:
At this point all my branch does is add a toggle for the background and add the outline and shadow like @welchyd says. The label connection has been removed.

src/os/annotation/abstractannotationctrl.js
```diff
@@ -73,7 +75,8 @@ os.annotation.UI_TEMPLATE =
           '</button>' +
         '</div>' +
       '</div>' +
-    '</div>';
+    '</div>' +
+  '</div>';
```
bradh: This looks strange. Where did the extra opening `<div>` come from?

MarcelloNicoletti: When I was working on this I pulled the string out to it's own file. While it was separated, with syntax highlighting I noticed the missing closing tag. I didn't do a blame to see how long it's been like this.

views/annotation/annotationoptions.html
```diff
@@ -32,16 +45,16 @@
           </label>
         </div>
       </div>
-      <div class="col">
+      <div class="col ml-5">
```
schmidtk: Is this intended to indent these color options under Show Background to convey their relationship? Personally I think it looks better with everything inline, and I don't believe we do this in any other forms. I would opt for disabling the color pickers when Show Background is unchecked.

views/annotation/annotationoptions.html
```diff
@@ -55,20 +68,22 @@
               ng-change="ctrl.updateAnnotation()" ng-disabled="!ctrl.annotationOptions.show"
               ng-model="ctrl.annotationOptions.showDescription"
             />
-            <label class="custom-control-label" ng-attr-for="showAnnotationDescription{{ctrl.uid}}">
+            <label class="custom-control-label" ng-attr-for="showAnnotationDescription{{ctrl.uid}}" ng-disabled="disabled">
```
schmidtk: I don't think `disabled` is ever set on the scope, nor do `label` elements have a disabled state. Is this needed?

welchyd:
Demo: http://spiky-calculator.surge.sh
When adding or editing a place with a text box, you can click the checkbox to remove the background of the text box
When you hover over text box with your mouse, it'll show an outline to the box

views/annotation/annotationoptions.html
welchyd: The legend settings used to have indentation like that ... I'm guessing that was lost with the bs4 upgrade though (as well as other examples)

schmidtk: @gregroyal mentioned that we avoid indenting levels in the UI because it's easier for a user to scan if everything is lined up. I think we should do the same here.

views/annotation/annotationoptions.html
MarcelloNicoletti: I'm not sure what this actually does. I was following the example set somewhere else in the code. It seems to be automatically set so that when the element in `ng-attr-for` is disabled the label also gets disabled.

schmidtk: I suspect the other places in this template using `ng-disabled="disabled"` were erroneously copied from another file. The `ngDisabled` directive sets `disabled="disabled"` on an element if the expression is truthy. In this case the expression is watching `$scope.disabled`, which is never set so the expression is always false.

views/annotation/annotationoptions.html
kwilson-bits: Extra padding has been removed.

views/annotation/annotationoptions.html
kwilson-bits: Extra tags have been removed, and the color picker inputs are properly disabled when the background is not shown.

PR #651, opened 2019-07-06
WIP: Update eslint-config-opensphere to 2.0.0.

schmidtk:


jsalankey:
Does this cause an error on not leaving a whitespace in jsdocs before parameters?

schmidtk:
@jsalankey Yes, it requires a blank line between the description and the params.

I could go either way on that rule. It's intended for readability and is fixed with `eslint --fix`, but I'm not fully convinced it's worth requiring developers to do in every comment block.

schmidtk:
Closing until config is finalized.

PR #654, opened 2019-07-08
fix(LayerAccordion): Set Accordion

adamstja:
1. Under the Layers menu select any of the Feature or Map layers.
2. Select Style or Zoom to open the accordion. 
3. Select a different layer (feature or map).
4. Re-select the same tile or layer, your style or zoom should return to open or closed where you left it. Not default to open. 

[whole-bedroom.surge.sh](url)

### DIFF
--- a/views/layer/vector.html
+++ b/views/layer/vector.html
@@ -2,7 +2,7 @@
   <div class="card">
     <div class="btn btn-secondary btn-block btn-sm" title="Style controls for the layer(s)"
         data-toggle="collapse" data-target="#layer-style"
-        ng-click="vector.setOpenSection('#layer-style')">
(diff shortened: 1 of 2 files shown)


views/layer/vector.html
```diff
@@ -2,7 +2,7 @@
   <div class="card">
     <div class="btn btn-secondary btn-block btn-sm" title="Style controls for the layer(s)"
         data-toggle="collapse" data-target="#layer-style"
-        ng-click="vector.setOpenSection('#layer-style')">
+        ng-click="vector.setOpenSection('#layer-style')"></div>
```
schmidtk: Was this accidental? Line 7 closes this div.

adamstja: Yes it was. Amending this now.

PR #663, opened 2019-07-09
fix(LayerUi): fix save of collapse state and add save call to style tab

### LINKED ISSUE #630: Accordion menu sticking open
In the layers window the accordion menus have a tendency to want to be open after they have been clicked on once before - we need to have them stay closed once they've been closed

To get this to happen:
Click on a feature or map layer in the layers menu
Click on Style to open the style menu
Click on Style again to close it
Now click on a different layer
The style menu magically opens again

adamstja:
fix #630

To get this to happen:
Click on a feature or map layer in the layers menu.
Click on Style to open the style menu (or zoom).
Click on Style again to close it (or zoom).
Now click on a different layer...
Click on the previous zoom or style tab, it will either be closed or left open as you had it before.


adamstja:
[http://petite-ring.surge.sh/](url)

PR #670, opened 2019-07-11
chore(release-prep): add manually-run script to kick off a release

wallw-bits:
* Organizes scripts under the `tasks/` folder
* `chmod +x` to `deploy.sh`
* Add `tasks/release.sh`, which will be run manually to kick off a build in Travis that will follow up with a run of `semantic-release`
* `semantic-release` is currently passed `--dry-run` so we can check it running once this hits master. I'll remove that once we verify that the release script is kicking that off correctly.

The PR build is likely to fail due to changes to `.travis.yml`


.travis.yml
```diff
@@ -25,17 +25,17 @@ install:
...
   except:
   - "/^v\\d+\\.\\d+\\.\\d+$/"
 env:
   global:
-  - S3SCREENSHOT=true
-  - S3COMPARISON=true
-  - S3VIDEO=true
+    S3SCREENSHOT: true
+    S3COMPARISON: true
+    S3VIDEO: true
```
wallw-bits: These were changed because the API config merge doesn't play as nicely with the list syntax version.

PR #671, opened 2019-07-12
fix(annotations): fixes annotation positioning/persistence issues

jsalankey:
- Adds a position property to annotation options for persisting their position. I'm not exactly sure why this wasn't persisted before, but keeping it around fixes a lot of issues with show/hide and refresh positioning for annotations. They should now no longer reset their position on toggling show/hide nor on refresh.
- Fixes losing the show/hide state on refresh when clicking the toggle buttons on both the node UI and the annotation hover controls.
- Fixes the bug where clicking the top right X on the placemark edit window would cause the annotation hover controls to not show up.

### DIFF
--- a/externs/os.externs.js
+++ b/externs/os.externs.js
@@ -28,6 +28,7 @@ osx.annotation;
  *   showTail: string,
  *   size: !Array<number>,
  *   offset: !Array<number>,
+ *   position: (Array<number>|undefined),
  *   headerBG: (string|undefined),
  *   bodyBG: (string|undefined)
  * }}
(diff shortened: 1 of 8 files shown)


schmidtk:
Also weird: the issue where polygons in Saved Places are filled on refresh... only happens if an annotation is enabled on the feature.

jsalankey:
Good catch. Both of these problems are 2D mode only and are artifacts of the `ol.Overlay`s weird behavior of not exposing the `setVisibility` method and instead deciding whether to show itself based on whether there's a position defined or not. Originally, we were setting the position to null if the visibility was off in order to hide the overlay. Now, I'm directly controlling the visibility property and checking it before re-rendering in `updatePixelPosition`.

Changing this seems to have fixed both cases. The filled polygon issue is unrelated.

welchyd:
I think #620 introduced the fill issues

jsalankey:
OK, so I think I covered the 50 different visibility states that can apply to the annotation. I went back on trying to control the visibility directly, and instead just set the position to null to hide the overlay/annotation. The position will still be set on the annotation options at all times, but may not be set on the overlay itself as a result.

Issue #676, opened 2019-07-16
Default icon for features imported via kml is now a push pin


justin-bits:
![image](https://user-images.githubusercontent.com/44378016/61315331-c9549f80-a7bb-11e9-9ed8-de135fd74a4b.png)

justin-bits:
Workarounds in feature-list.spec.js and load-state-file-url.kml.spec.js need to be removed after this is fixed.

Issue #667, opened 2019-07-10
Context menus not always opening from grids


justin-bits:
From Layers or Feature List (anywhere there's a grid), do left click, right click, left click, right click.  The context menu will not open on the second right click.  The selected row can even be changed a number of times and the context menu will still fail consistently on the second attempt.  On the third right click, the menu will open.

justin-bits:
Workarounds in feature-list.spec.js need to be removed after this is fixed.

Issue #581, opened 2019-05-31
Expand test coverage for Feature List dialog


justin-bits:
https://github.com/ngageoint/opensphere/pull/567

Expand coverage to fill in the gaps for Layer Features testing. This also appears to be the first table version of slick-grid and will serve as a good full test.

Dialog UI:
- Show Selected Only checkbox
- Row height slider

Context Menu:
- Sort Selected
- Export
- Go To

Mouse/Keyboard Interactions:
- Ctrl+click selections
- Shift+click selections
- Sort via clicking on header
- Moving columns via drag

Additional:
- Multiple instances of the dialog with different layers
- Smoke tests for additional types of data (e.g. csv)

justin-bits:
Ctrl / shift selections, sort selected, export, go to done here: https://github.com/justin-bits/opensphere/tree/test/refactor-selectors

PR #677, opened 2019-07-16
test(cypress): Splits selectors into multiple files

### LINKED ISSUE #580: Refactor Cypress selectors
https://github.com/ngageoint/opensphere/pull/567#issuecomment-497756988

- Break `selectors.js` into multiple files
- Create generic selectors for commonly used elements (e.g OK button)
- Update tests accordingly to reference updated selectors

justin-bits:
Resolves #580 

- Some performance changes to reduce flaky tests and produce a cleaner pattern to follow when using image comparisons
- `selectors.js` is now multiple files
-  The structure of the selectors were modified to reduce the complexity and length.  This results in shorter lines in tests and selectors that are easier to lookup. There was also a significant reduction of duplicate selectors. Dropped the use of `_WILDCARD`.
- Updates to images as workaround for #676 
- Significant rework/additions to `feature-list.spec.js` (which was renamed from `layer-features.spec.js` to more closely match the terminology used in the application).  Some work from #581 completed with this PR.
- Inclusion of workaround for #667
- `toolbar-right.spec.js` now has its own test data
- Added support for running any test case (or group of test cases) in a batch five times in a row to check for flaky tests during test development
- Handles specific console errors that started being thrown by OS.  @jsalankey indicated these specific errors are okay to disregard.
- Removed test selectors from OS as substitutes were identified
- #678 and #680 filed as known issues with the current version of tests/OS


justin-bits:
@bradh, oops! I just removed it. Good catch.

schmidtk:
The `outerWidth` console error will still be present until #691 is merged.

Issue #684, opened 2019-07-17
Allow ability to set NodeUI in the base descriptor


robertbolan-bits:
Allow the ability to set nodeUI in base descriptor.

schmidtk:
Resolved in #685.

PR #685, opened 2019-07-17
Update BaseDescriptor to with new Property of NodeUI

robertbolan-bits:
#684

Added a new  setNodeUI() function to IDataDescriptor
Added a new nodeUI property to BaseDescriptor.
Update getNodeUI on config descriptor to append nodeUI new UI element.

If you set nodeUI on configdescriptor object it will append the new UI Element to the nodeui from LayerSyncDescriptor


src/os/data/idatadescriptor.js
```diff
@@ -268,6 +268,12 @@ os.data.IDataDescriptor.prototype.getSearchText;
 os.data.IDataDescriptor.prototype.getNodeUI;
 
 
+/**
+ * Sets the Node UI for this desceriptor
+ * @param {string} node UI HTML
+ */
+os.data.IDataDescriptor.prototype.setNodeUI;
```
welchyd: `setNodeUI` will need to be implemented anywhere that extends `os.data.IDataDescriptor` ... so `iogcdescriptor.js`, `istatedescriptor.js`, `iarcdescriptor.js` and anything downstream from there

bradh: @welchyd So this will be a breaking change for external plugins?

welchyd: Yes ... @robertbolan-bits might be able to just limit it to `baseDescriptor.js`, but yes it would affect external plugins otherwise

bradh: Depending on how that works out, https://opensphere.readthedocs.io/en/latest/guides/plugin_file_type_guide/descriptor.html (https://github.com/ngageoint/opensphere/blob/master/docs/guides/plugin_file_type_guide/descriptor.rst and included demo code) might need an update.

robertbolan-bits: I will move the new function to just the base descriptor and take it off the interface. I think that will be for the best so the change isnt so wide spread

src/os/data/configdescriptor.js
```diff
@@ -290,7 +290,8 @@ os.data.ConfigDescriptor.prototype.reimport = function() {
  * @inheritDoc
  */
 os.data.ConfigDescriptor.prototype.getNodeUI = function() {
-  return /** @type {string} */ (this.getFirst('nodeUi')) || '<defaultlayernodeui></defaultlayernodeui>';
+  return /** @type {string} */ (this.getFirst('nodeUi')) ||
+      os.data.ConfigDescriptor.base(this, 'getNodeUI') + this.nodeUI;
```
wallw-bits: The `+ this.nodeUI` portion is incorrect. That's what will (most likely) be returned by the super call.

PR #687, opened 2019-07-18
chore(css): Disable sass rules at block scope

bradh:
The sass-lint:disable variant has file scope.

Relates to #141 - this narrows down the scope.


bradh:
An alternative (or adjunct) would be to disable the `extends` rules.

PR #689, opened 2019-07-19
fix(scss): webkit theme autofill

### LINKED ISSUE #655: Auto-complete background doesn't match theme
When selecting from a suggested option, the background to the selection doesn't match the theme.
Also, once a recommendation is selected the background doesn't match the theme

selection

Story points: 1

adamstja:
fix #655

1. Create new place. 
2. Clear input of "Position".
3. Enter one digit to activate autofill
4. Select a coordinate from drop down.
5. Previous behavior filled the Position input with white, this didn't match the theme. Current behavior matches them (black/grey).
6. Works for both Light and Dark mode. 

[http://trashy-steel.surge.sh/](url)


scss/os/_overrides_bootstrap.scss
```diff
@@ -156,3 +156,10 @@ ul {
     margin-bottom: 0;
   }
 }
+
+//sass-lint:disable no-duplicate-properties, no-vendor-prefixes
+input:-webkit-autofill,
+input:-webkit-autofill:active  {
+  -webkit-box-shadow: 0 0 0 30px #555 inset !important;
```
schmidtk: These should use Bootstrap variables so they work in any theme, instead of assuming the dark theme.

scss/os/_overrides_bootstrap.scss
```diff
@@ -156,3 +156,10 @@ ul {
     margin-bottom: 0;
   }
 }
+
+//sass-lint:disable no-duplicate-properties, no-vendor-prefixes
```
schmidtk: Please include a note on why this is needed, so it isn't removed if someone is working on #141. We prefer to avoid vendor prefixes, but this doesn't have an alternate solution.

scss/os/_overrides_bootstrap.scss
adamstja: Bootstrap variables "$input-bg" and "$input-color" have been added

scss/os/_overrides_bootstrap.scss
adamstja: sass-lint:disable-block has been added

schmidtk:
This doesn't seem to work for me (Mac, Chrome 75). Are you seeing a difference in light/dark themes? I also tried the other solution provided in [this thread](https://stackoverflow.com/questions/2781549/removing-input-background-colour-for-chrome-autocomplete), which did not work either. According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:-webkit-autofill), the browser may force `!important` in the user agent stylesheet, preventing overrides in CSS. I'm not convinced this is worthwhile if it's not even fully supported by Chrome/Webkit browsers. @welchyd?

welchyd:
This works for me in older versions of Chrome on Mac ...
The newer versions on Chrome have started becoming responsive to the Mac's theme - maybe that's a factor here? https://www.freecodecamp.org/news/how-to-get-dark-mode-working-with-css-740ad31e22e/

schmidtk:
Perhaps. Looks like Chrome 76 will support the [prefers-color-scheme media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme), but that hasn't been released yet. It's possible they are already forcing overrides of the CSS in the PR.

welchyd:
I'm not seeing issues with Chrome 75 on a Mac, here's a picture:
ThemeAutofill

schmidtk:
I misunderstood what was being colored here. I thought it was the dropdown, but it's the input after an autofill option is selected. That said, I'm still seeing this:

Screen Shot 2019-07-24 at 9 08 49 AM

It does work correctly if I add this block to apply the keyframe the current changes are setting up:
```
input:-webkit-autofill {
  -webkit-animation-fill-mode: both;
  -webkit-animation-name: autofill;
}
```

PR #691, opened 2019-07-19
fix(window): Clean up resize handler and avoid errors if called after dispose.

schmidtk:
Fixes an error introduced if the resize handler was called after disposal (`this.element` was no longer defined), and prevents leaking windows due to not removing the event listener.


schmidtk:
Also fixed leaking settings window, caused by not calling parent destroy functions and leaving listeners around.

PR #692, opened 2019-07-19
fix(kml): Replace the OpenLayers default icon with OpenSphere's.

### LINKED ISSUE #676: Default icon for features imported via kml is now a push pin

schmidtk:
Fixes #676.

OpenLayers uses the yellow pushpin as the default icon style to match Google Earth's behavior. We prefer the placemark circle, so this replaces their defaults with ours.

To test, load [this file](https://github.com/ngageoint/opensphere/files/3412570/DefaultPointStyle.txt). There should be a single point using the white placemark circle as the icon.

Note the KML node color varies because a `PolyStyle` is defined in the KML, and that color is resolved when the node looks for a color in the style config. This is perhaps a bug since that color doesn't impact the rendered feature, but a separate issue if we want that behavior to change.

### DIFF
--- a/src/plugin/file/kml/kml.js
+++ b/src/plugin/file/kml/kml.js
@@ -93,6 +93,38 @@ plugin.file.kml.replaceParsers_ = function(obj, field, parser) {
 
 
 /**
+ * Create default OpenLayers styles along with OpenSphere overrides.
+ * @suppress {accessControls, const}
+ */
+plugin.file.kml.createStyleDefaults = function() {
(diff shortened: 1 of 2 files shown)


bradh:
Would it be worth parameterising this in `config/settings.json`?

wallw-bits:
@bradh this fixes a regression. If you want to write that up as an enhancement, that would be good.

schmidtk:
I agree with Will. That's a great idea, but the majority of places referencing the default icon are defined at the global scope. Changing that to a settings-driven value would require more of a refactor than I'm comfortable combining with this fix.

PR #693, opened 2019-07-19
feat(FillStyle): Add controls for fill color and opacity for shapes

kwilson-bits:
Add controls for fill color and opacity for shapes, vector layers, and feature actions.
re ngageoint#531

Testing link: http://os-fill-controls.surge.sh/


welchyd:
I found a few small things:
* For features with fill - If I set the fill color to something different than the border color, and then change the border color to match the fill color then change the border color, it doesn't change the fill color (this works fine in the nodeui, just not the feature edit dialog)
* When changing the fill color, the node doesn't update (could be pre-existing)
* Still testing, will fill out later

src/os/ui/featureedit.js
```diff
@@ -978,6 +994,12 @@ os.ui.FeatureEditCtrl.prototype.loadFromFeature = function(feature) {
       this['centerIcon'] = icon;
     }
 
+    if (config.fill && config.fill.color) {
+      this['fillColor'] = os.color.toHexString(config.fill.color);
+      var opacity = os.color.toRgbArray(config.fill.color);
+      this['fillOpacity'] = opacity[3];
+    }
+
```
welchyd: I think changing this to a string reference will fix the issue I saw with border color not being in sync with the fill color.

```
if (config['fill'] && config['fill']['color']) {
  this['fillColor'] = os.color.toHexString(config['fill']['color']);
  var opacity = os.color.toRgbArray(config['fill']['color']);
  this['fillOpacity'] = opacity[3];
}
```

src/os/command/feature/featurefillcolorcmd.js
```diff
@@ -0,0 +1,124 @@
...
+ * @param {(Array<number>|string)=} opt_oldColor
+ * @constructor
+ */
+os.command.FeatureFillColor = function(layerId, featureId, color, opt_oldColor) {
+  os.command.FeatureFillColor.base(this, 'constructor', layerId, featureId, color, opt_oldColor);
+  this.title = 'Change Feature Fill Color';
+  this.metricKey = os.metrics.Layer.FEATURE_FILL_COLOR;
+
+  if (!color) {
+    var feature = /** @type {ol.Feature} */ (this.getFeature());
```
jsalankey: This typecast should not be necessary. That function is already typed to return an `ol.Feature`, so this cast doesn't tell the compiler anything it doesn't already know.

src/os/command/feature/featurefillopacitycmd.js
```diff
@@ -0,0 +1,80 @@
...
+};
+
+
+/**
+ * @inheritDoc
+ */
+os.command.FeatureFillOpacity.prototype.finish = function(configs) {
+  // dispatch the fill opacity change event on the source for the histogram
+  var feature = /** @type {ol.Feature} */ (this.getFeature());
+  var source = /** @type {plugin.file.kml.KMLSource} */ (os.feature.getSource(feature));
```
jsalankey: This cast is referencing the KML plugin in core code, and is likely not making a correct typecast assumption about the type of source it's getting in all cases. There should be a better way to fire change events on the features themselves that then get handled by their nodes to do KML-specific stuff. It may also be more appropriate for this change to be notified on the source. @schmidtk would know better than me.

src/os/command/feature/featureopacitycmd.js
```diff
@@ -69,3 +70,21 @@ os.command.FeatureOpacity.prototype.applyValue = function(configs, value) {
...
 };
+
+
+/**
+ * @inheritDoc
+ */
+os.command.FeatureOpacity.prototype.finish = function(configs) {
+  // dispatch the opacity change event on the source for the histogram
+  var feature = /** @type {ol.Feature} */ (this.getFeature());
+  var source = /** @type {plugin.file.kml.KMLSource} */ (os.feature.getSource(feature));
```
jsalankey: Same comment about referencing the KML plugin here.

src/os/command/feature/featurestrokecolorcmd.js
```diff
@@ -0,0 +1,128 @@
...
+};
+
+
+/**
+ * @inheritDoc
+ */
+os.command.FeatureStrokeColor.prototype.finish = function(configs) {
+  // dispatch the color change event on the source for the histogram
+  var feature = /** @type {ol.Feature} */ (this.getFeature());
+  var source = /** @type {plugin.file.kml.KMLSource} */ (os.feature.getSource(feature));
```
jsalankey: Plugin reference.

src/os/command/feature/featurestrokeopacitycmd.js
```diff
@@ -0,0 +1,93 @@
...
+};
+
+
+/**
+ * @inheritDoc
+ */
+os.command.FeatureStrokeOpacity.prototype.finish = function(configs) {
+  // dispatch the stroke opacity change event on the source for the histogram
+  var feature = /** @type {ol.Feature} */ (this.getFeature());
+  var source = /** @type {plugin.file.kml.KMLSource} */ (os.feature.getSource(feature));
```
jsalankey: Plugin reference.

src/os/command/feature/featurefillopacitycmd.js
schmidtk: The command shouldn't have to use the source at all. It should fire a change event on the feature, which should be handled by the source/node. In this case, you should fire an event on the feature and handle it in `plugin.file.kml.ui.KMLNode#onFeatureChange`. The handler should respond by firing a new event from the node.

src/os/command/feature/featurefillopacitycmd.js
```diff
@@ -0,0 +1,80 @@
...
+  var rootNode = source.getRootNode();
+  var children = rootNode.getChildren();
+
+  for (var i = 0; i < children.length; i++) { // update icon color
```
schmidtk: In general, please try to minimize dispatched events to the impacted node(s). The current approach doesn't even fire an event from the feature node, and is instead trying to trigger an update from the immediate children of the root.

src/os/state/v4/baselayerstate.js
```diff
@@ -492,6 +504,22 @@ os.state.v4.BaseLayerState.prototype.configKeyToXML = function(layerConfig, type
         os.xml.appendElement(os.state.v4.LayerTag.ALPHA, layerEl, value);
       }
       break;
+    case 'fillOpacity':
```
jsalankey: This should be added to `os.style.StyleField`.

src/os/ui/file/kml/abstractkmlexporter.js
```diff
@@ -712,10 +712,11 @@ os.ui.file.kml.AbstractKMLExporter.prototype.addGeometryNode = function(item, no
  * @param {T} item The item
  * @param {string} styleId The style id
  * @param {string} color The item color
+ * @param {string} fillColor The item fill color
  * @param {os.ui.file.kml.Icon=} opt_icon The item icon
  * @protected
  */
-os.ui.file.kml.AbstractKMLExporter.prototype.createStyle = function(item, styleId, color, opt_icon) {
+os.ui.file.kml.AbstractKMLExporter.prototype.createStyle = function(item, styleId, color, fillColor, opt_icon) {
```
jsalankey: Adding a non-optional argument to a public function constitutes a breaking API change and needs to be annotated as such in the commit message for the change.

src/plugin/file/kml/kmlnodelayerui.js
```diff
@@ -599,17 +782,18 @@ plugin.file.kml.KMLNodeLayerUICtrl.prototype.onCenterShapeChange = function(even
 
 
 /**
- * Handles changes to color
+ * Handle changes to opacity while it changes via slide controls
  *
- * @param {angular.Scope.Event} event
- * @param {number} value
+ * @param {?angular.Scope.Event} event
+ * @param {?} value
```
jsalankey: ? is very rarely a good type declaration. These should probably be numbers.

schmidtk: Jeremy is correct, the all/any types should be avoided if at all possible. We also generally use the all (`*`) type vs the any (`?`) type, because the any type is completely ignored by the compiler.

https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#type-expressions

src/os/command/feature/featurefillopacitycmd.js
welchyd: Admittedly this seems to be copied from my other implementations of feature commands ... so this would probably need to be fixed for all of those commands

src/os/ui/featureedit.js
kwilson-bits: This fix has been applied and partially fixed the problem. The preview on the screen now updates but the fill color within the Feature Edit form does not update.

src/os/command/feature/featurefillcolorcmd.js
kwilson-bits: This should now be fixed. I will squash commits at the end to add the breaking change message.

src/os/command/feature/featurefillopacitycmd.js
kwilson-bits: This should now be improved; the plugin references have been moved to just plugin code. I will squash commits at the end to add the breaking change message.

src/os/command/feature/featureopacitycmd.js
kwilson-bits: Same comment as previous plugin reference.

src/os/command/feature/featurestrokecolorcmd.js
kwilson-bits: Plugin reference updated.

src/os/state/v4/baselayerstate.js
kwilson-bits: I switched it to use the reference defined in os.style.StyleField

src/plugin/file/kml/kmlnodelayerui.js
kwilson-bits: This has been updated to properly use Number types.

src/os/ui/featureedit.js
```diff
@@ -161,6 +161,18 @@ os.ui.FeatureEditCtrl = function($scope, $element, $timeout) {
    */
   this['size'] = os.style.DEFAULT_FEATURE_SIZE;
 
+  /**
+   * The feature fill color
+   * @type {string}
+   */
+  this['fillColor'] = os.style.DEFAULT_LAYER_COLOR;
```
welchyd: This essentially causes the color issue I found (where the fill color doesn't stay in sync with the stroke color for new features). Basically `os.style.DEFAULT_LAYER_COLOR` is not a hex string, but everywhere that `fillColor` is used expects it that way, including the color picker. The `labelColor` doesn't have that problem because the initial setting gets converted to hex before going to the color picker, while the `fillColor` is only changed when `loadFromFeature()` is called (so `createPreviewFeature()` addresses `labelColor` but not `fillColor` in line 924).

```
this['fillColor'] = os.color.toHexString(os.style.DEFAULT_LAYER_COLOR);
```

With that being said, it would also be wise to initialize `color` and `labelColor` this way too.

src/os/command/feature/featurefillopacitycmd.js
kwilson-bits: Dispatching events from the feature or source nodes does not trigger the updated color/opacity values to persist through screen refreshes, that only happened when I dispatched events on the children. The code has still been moved out to plugin/ files to keep plugin references out of core code.

src/os/ui/file/kml/abstractkmlexporter.js
kwilson-bits: I added the comment in a later commit and then squashed the commits. It _should_ be updated properly so that it gets caught by scripts, but I am not 100% sure I got it right. Please verify.

src/os/ui/featureedit.js
kwilson-bits: I added the logic to convert the globals to hex strings to color, labelColor, and fillColor.

src/plugin/file/kml/ui/kmlnode.js
```diff
@@ -304,6 +304,14 @@ plugin.file.kml.ui.KMLNode.prototype.onFeatureChange = function(event) {
         this.dispatchEvent(new os.events.PropertyChangeEvent('icons'));
         this.dispatchEvent(new os.events.PropertyChangeEvent(os.annotation.EventType.CHANGE));
         break;
+      case 'colors':
+        var source = /** @type {plugin.file.kml.KMLSource} */ (os.feature.getSource(this.feature_));
+        var rootNode = source.getRootNode();
+        var children = rootNode.getChildren();
+        for (var i = 0; i < children.length; i++) { // update icon color
+          children[i].dispatchEvent(new os.events.PropertyChangeEvent('icons'));
```
jsalankey: This is still dispatching change events on all of the children of the root, which seems incorrect. The node that receives the color change event (which is this node, as it is the one tied to the feature) should be the only one that needs to dispatch a `new os.events.PropertyChangeEvent('icons')`.

welchyd: So something more like this? `this.dispatchEvent(new os.events.PropertyChangeEvent('icons'));`

jsalankey:
- Places are not visible while editing them due to the fact that the fill opacity slider is set to 0.
- The fill style on points is overriding the stroke style. Changing the original color picker has no effect if there's a fill color set.
- Having a fill style slider at all with a point is confusing. This is part of a larger issue that also applies to the border style picker, so we may not be able to fix it here. However, they shouldn't conflict with each other.
- The fill color is not reflected in the node icon in the layers window. This is a difficult problem also, to which the best solution is to have those icons show sort of both a stroke color on themselves and a fill color. We should write a followon to address this issue.
- The fill color on a point place is applying and changing properly, but on refresh, it is not applying. The color is still set in the feature style UI on refresh, but it's not being applied.
- Undoing a fill opacity change is always resetting the fill opacity to 100%.

kwilson-bits:
> Testing feedback:
> 
> * Places are not visible while editing them due to the fact that the fill opacity slider is set to 0.
> * The fill style on points is overriding the stroke style. Changing the original stroke color picker has no effect if there's a fill color set.
> * Having a fill style slider at all with a point is confusing. This is part of a larger issue that also applies to the border style picker, so we may not be able to fix it here. However, they shouldn't conflict with each other.
> * The fill color is not reflected in the node icon in the layers window. This is a difficult problem also, to which the best solution is to have those icons show sort of both a stroke color on themselves and a fill color. We should write a followon to address this issue.
> * The fill color on a point place is applying and changing properly, but on refresh, it is not applying. The color is still set in the feature style UI on refresh, but it's not being applied.
> * Undoing a fill opacity change is always resetting the fill opacity to 100%.
> * The overall opacity slider now synchronizes both sliders. These two things should be able to be adjusted totally independently.

The opacity slider behavior should be as follows:
- When the stroke opacity and the fill opacity have the same value and the stroke slider is moved, the fill slider should move to match it. The idea being that if you have set the stroke and fill to be the same value, you want to keep them the same value.
- As part of this, if you slide the stroke opacity to be the same as the fill opacity's value, and then keep sliding the stroke opacity, it will pick up the fill opacity and change them both to be the same.
- Regardless of the stroke opacity value, when moving the fill slider you should only change the fill opacity. This way you can always change one without changing the other.

This should now behave the same for both commands and Feature Edit.

When undoing a fill opacity change, it should now properly reset to the previous value.

src/os/command/feature/featurefillcolorcmd.js
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Changes the fill color of a feature
+ * @extends {os.command.AbstractFeatureStyle}
+ * @param {string} layerId
+ * @param {string} featureId
+ * @param {Array<number>|string} color
+ * @param {(Array<number>|string)=} opt_oldColor
+ * @constructor
+ */
+os.command.FeatureFillColor = function(layerId, featureId, color, opt_oldColor) {
```
schmidtk: It seems there is a lot of duplicate code between this and `os.command.FeatureColor`. Same goes for the opacity commands and the stroke color/opacity. We should either abstract out the common logic, or the fill/stroke variants should extend the originals and replace only what they need. You're welcome to refactor the originals to make this easier/cleaner. It seems with a few small changes we could significantly reduce the code duplication in this PR.

src/os/command/vectorlayerfillcolorcmd.js
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Changes the fill color of a layer
+ * @extends {os.command.AbstractVectorStyle}
+ * @param {string} layerId
+ * @param {Array<number>|string} color
+ * @param {(Array<number>|string)=} opt_oldColor
+ * @constructor
+ */
+os.command.VectorLayerFillColor = function(layerId, color, opt_oldColor) {
```
schmidtk: Same goes for the vector layer versions of these commands. There is significant duplication of code happening, and it seems unnecessary.

src/os/command/feature/featurefillcolorcmd.js
schmidtk: On second thought, it seems entirely possible to add an option to the constructor params (`opt_isFill`) that changes the operation of the command to change the fill color/opacity.

kwilson-bits: I have consolidated the commands and removed the separate Stroke and Fill command files. These are still separate commands and still show up in action history as separate entries, but should only affect the fill, stroke, or both, as dictated by the user's action.

src/os/command/vectorlayerfillcolorcmd.js
kwilson-bits: The vector layer commands have also been reduced from 6 down to 2.

src/plugin/file/kml/kmlnodelayerui.js
```diff
@@ -1092,6 +1110,27 @@ plugin.file.kml.KMLNodeLayerUICtrl.prototype.isFeatureDynamic = function() {
...
+ * @return {boolean}
+ * @export
+ */
+plugin.file.kml.KMLNodeLayerUICtrl.prototype.isFeatureFillable = function() {
+  var features = this.getFeatures();
+  var feature = features.length > 0 ? features[0] : null;
+  if (feature) {
+    var geometry = feature.getGeometry();
+
+    if (geometry instanceof ol.geom.Polygon || geometry instanceof ol.geom.MultiPolygon) {
```
schmidtk: This can be replaced with `os.geo.isGeometryPolygonal`. We also prefer checking the geometry's `getType` value over using `instanceof`.

src/os/ui/featureedit.js
```diff
@@ -739,6 +744,25 @@ os.ui.FeatureEditCtrl.prototype.isEllipse = function() {
...
+ * If a polygon is selected.
+ *
+ * @return {boolean}
+ * @export
+ */
+os.ui.FeatureEditCtrl.prototype.isPolygon = function() {
+  var geometry = this.previewFeature.getGeometry();
+
+  if (geometry) {
+    var type = geometry.getType();
```
schmidtk: This duplicates `os.geo.isGeometryPolygonal`. You can use that instead.

src/plugin/file/kml/kmlnodelayerui.js
kwilson-bits: This has been updated.

src/os/ui/featureedit.js
kwilson-bits: This has been updated.

src/os/ui/layer/vectorlayerui.js
```diff
@@ -320,6 +413,127 @@ os.ui.layer.VectorLayerUICtrl.prototype.onColorReset = function(event) {
...
+ * Handles changes to fill color
+ * @param {angular.Scope.Event} event
+ * @param {string} value
+ * @protected
+ */
+os.ui.layer.VectorLayerUICtrl.prototype.onFillColorChange = function(event, value) {
+  event.stopPropagation();
+
+  // Make sure the value includes the current opacity
+  var colorValue = os.color.toRgbArray(value);
```
schmidtk: This needs to handle the reset case, where the fill color is an empty string.

src/os/ui/layer/vectorlayerui.js
```diff
@@ -290,22 +329,76 @@ os.ui.layer.VectorLayerUICtrl.prototype.reconcileLabelsState_ = function() {
 os.ui.layer.VectorLayerUICtrl.prototype.onColorChange = function(event, value) {
   event.stopPropagation();
 
-  var fn =
+  // Make sure the value includes the current opacity
+  var colorValue = os.color.toRgbArray(value);
```
schmidtk: This needs to handle the reset case, where the fill color is an empty string.

src/plugin/file/kml/kmlnodelayerui.js
```diff
@@ -480,16 +578,117 @@ plugin.file.kml.KMLNodeLayerUICtrl.prototype.onLockChange = function() {
 plugin.file.kml.KMLNodeLayerUICtrl.prototype.onColorChange = function(event, value) {
   event.stopPropagation();
 
-  var fn =
+  // Make sure the value includes the current opacity
+  var colorValue = os.color.toRgbArray(value);
```
schmidtk: This needs to handle falsey values for the color (`!value`).

src/plugin/featureaction/featurestyleaction.js
```diff
@@ -306,7 +306,8 @@ plugin.im.action.feature.StyleAction.prototype.fromXml = function(xml) {
         fillColor = os.style.toRgbaString(fillColorArr);
       }
 
-      os.style.setConfigColor(styleConfig, fillColor, [os.style.StyleField.FILL]);
+      // Only change the fill color without changing the image fill color too
+      styleConfig.fill.color = fillColor;
```
schmidtk: The config properties should use bracket notation (`styleConfig['fill']['color']`), to avoid being renamed by the compiler which would break this functionality in the compiled app.

src/plugin/featureaction/ui/featurestyleactionconfig.js
```diff
@@ -154,9 +154,9 @@ plugin.im.action.feature.ui.StyleConfigCtrl.prototype.initialize = function() {
     var strokeColor = os.style.toRgbaString(color);
     os.style.setConfigColor(this.styleConfig, strokeColor);
 
-    // If we have a fill color, standardize that too
+    // If we have a fill color, set that to our style config
     var fillColor = os.style.toRgbaString(fill);
-    os.style.setConfigColor(this.styleConfig, fillColor, [os.style.StyleField.FILL]);
+    this.styleConfig.fill.color = fillColor;
```
schmidtk: This file also needs bracket notation for style config properties.

src/plugin/file/kml/kmlnodelayerui.js
```diff
@@ -1092,6 +1092,46 @@ plugin.file.kml.KMLNodeLayerUICtrl.prototype.createFeatureCommand = function(com
...
+  var items = /** @type {Array<!plugin.file.kml.ui.KMLNode>} */ (this.scope['items']);
+  if (items) {
+    for (var i = 0; i < items.length; i++) {
+      var feature = items[i].getFeature();
+      var source = items[i].getSource();
+      if (feature && source) {
+        var featureId = feature.getId();
+        var layerId = source.getId();
+        if (featureId && layerId) {
+          if (typeof featureId == 'number') {
```
schmidtk: A simple way to do this is `featureId = String(featureId);`, and skip the type check. If it's already a string, it won't be changed.

The prior condition (`if (featureId ...)`) will also fail if the id is the number `0`, so `featureId != null` would be a better test. That won't catch an empty string but... I'm hoping we would never do that.

src/os/command/feature/featurecolorcmd.js
```diff
@@ -41,6 +39,8 @@ os.command.FeatureColor = function(layerId, featureId, color, opt_oldColor, opt_
       break;
   }
 
+  os.command.FeatureColor.base(this, 'constructor', layerId, featureId, color, opt_oldColor);
```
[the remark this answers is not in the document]
kwilson-bits: The value for 'this.changeMode' needed to be set before the parent constructor was run. If it is not set and no 'opt_oldValue' is set, the parent constructor calls this.getOldValue() to determine the previous value. Without that change, any undo action was getting the default opacity instead of properly getting the fill's previous opacity.

That said, the constructor should have been moved to just below setting 'this.changeMode' instead of where I previously moved it. This function has been updated.

src/plugin/file/kml/kmlnodelayerui.js
kwilson-bits: This function will now exit without doing anything if the color value is not an array or valid color string.

src/plugin/file/kml/kmlnodelayerui.js
```diff
@@ -480,16 +578,117 @@ plugin.file.kml.KMLNodeLayerUICtrl.prototype.onLockChange = function() {
...
+         * @return {os.command.ICommand}
+         */
+        function(layerId, featureId) {
+          return new os.command.FeatureColor(layerId, featureId, colorValue);
+        };
+
+      this.createFeatureCommand(fn);
+    } else if (color == fillColor) {
+      // We run these separately so that they retain the different opacities
+      var fn2 =
```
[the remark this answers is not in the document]
kwilson-bits: This should be working with the most recent push.

schmidtk:
I cleaned up remaining issues I've come across in testing, as well as a few bugs from master:
* Undo feature color change will update the label color if it's the same as the feature.
* Quick added places were not persisting their label columns, so labels would disappear after refresh.

schmidtk:
I deployed this to Surge at http://os-fill-controls.surge.sh/, updated in the issue description.

schmidtk:
@welchyd 

> When creating or editing a Place - the border dash options drop down doesn't show the initial selection

This is broken on master, unrelated.

> In the layers window, when the opacity and fill opacity sliders get set to the same value, changing the opacity slider also causes the fill opacity to change. This is not consistent with the behavior when creating/editing a Place

Ah thanks, missed that one when I was removing the opacity/fill opacity linkage. Fixed/deployed.

justin-bits:
-Undo shortcut not working after changing Border.  Seems to be a focus issue as the change is going into the history and undo/redo do work after changing focus to the map or some other element.
-Border not retained through place export/import.

schmidtk:
@justin-bits I believe all of those are resolved, though I didn't see the second bullet prior to my latest changes.

This PR was making a fundamental change to how the layer-level opacity slider was working. Instead of using OL's `get/setOpacity` API's, it was changing the opacity in the layer style config by changing the color alpha values. This presents a few problems:

* A new OL style will be created/cached for each opacity value.
* Layer opacity no longer multiplied by feature opacity. This broke the layer opacity control for any layer with feature-specific styling (KML, layers with feature actions applied, etc). Fixing this would require computing the combined layer/feature opacity internally, which would incur a performance penalty.
* Layer persist/restore behavior changed, impacting how state files are imported/exported.

To resolve these issues, the original layer opacity behavior was restored. This does mean the layer opacity will be multiplied by fill opacity, so if they're both at 50% the fill will be rendered at 25%. This seemed to be a reasonable compromise to avoid the above problems.

justin-bits:
Looks good

PR #702, opened 2019-07-26
chore(conformance): Add conformance violation for goog.isDef()

### LINKED ISSUE #165: Add conformance violations for goog.isDef
`goog.isDef(val)`, `goog.isNull(val)`, and `goog.isDefAndNotNull(val)` are all completely unnecessary in even the oldest browsers we have ever supported.

Make these a conformance violation and use `val !== undefined`, `val === null`, and `val !== undefined && val !== null` instead.

bradh:
Update two users.

Also add violation for unused `goog.isNull()` and `goog.isDefAndNotNull()`.

Resolves #165

### DIFF
--- a/conformance_config.textproto
+++ b/conformance_config.textproto
@@ -158,23 +158,23 @@ requirement: {
 #  error_message: 'goog.isBoolean(value) is not permitted. Use typeof value === \'boolean\' instead.'
 #}
 
-#requirement: {
-#  type: BANNED_NAME
-#  value: 'goog.isDef'
-#  error_message: 'goog.isDef(value) is not permitted. Use value !== undefined instead.'
-#}
+requirement: {
+  type: BANNED_NAME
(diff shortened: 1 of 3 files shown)


wallw-bits:
This is good but it constitutes a breaking build change for other projects, so the commit needs to be marked as such. Additionally, you might mention in the `BREAKING CHANGE` instructions that we have scripts to convert uses of these automatically at https://github.com/schmidtk/opensphere-jscodeshift

bradh:
Power-cycling the PR to see if we can get a passing CI build.

wallw-bits:
@schmidtk already knows what is wrong with the tests and is working it. We'll restart it when that is ready.

schmidtk:
Broken tests will be addressed by the PR's linked in https://github.com/ngageoint/opensphere/issues/709.

PR #712, opened 2019-08-06
Add support for running Cypress under Windows/WSL

justin-bits:
Overall added support for running Cypress tests under Windows.  Note that some of the implementation/usage was impacted by https://github.com/cypress-io/cypress/issues/4145.  Changes were NOT made to test scripts and any issues with those should be handled separately.  (Several tests are currently failing; this is known and unrelated)

Changes:
- Added workaround for a bug with `http-server` where Cypress was stuck in an endless redirect loop under Windows https://github.com/http-party/http-server/issues/525
- Added patch script to allow running Cypress from the command line under Git Bash after being installed under Ubuntu/WSL via `yarn install`. (Yarn operates differently when executed under Windows vs Linux environments.  Git Bash doesn't know how to follow the symlinks it generates under Ubuntu, hence the patch)
- Added a Windows compatible method of detecting running instances of a web server and terminating those instances automatically or on demand (via `yarn stop-server`)
- Added support for running `test:cypress-spec` and `test:cypress-loop` under Windows
- Updated documentation for Windows developers

Recommended Tests (under Windows AND Linux/MacOS):
Note: Windows tests should be executed under Git Bash, not PowerShell or the Command Prompt and definitely not within Ubuntu.
- Cypress tests running via GUI (`cypress:test`) AND CLI (any other cypress script) do not immediately fail with a redirect error
- ALL `package.json` Cypress scripts work (`test:cypress`, `test:cypress-all`, `test:cypress-smoke`, `test:cypress-spec`, `test:cypress-loop`).  To clarify, the script to start/stop the web server and Cypress work, tests are initiated.  This does not mean tests have to pass, just that they can run.
- `package.json` cypress scripts can be run back to back twice without script error (there are no lingering processes, issues with running on an already patched version of windows, etc.) (disregard test failures)
- Cypress tests startup without web server already running AND with web server already running
- Web server can be terminated manually (`yarn stop-server`)
- Ubuntu/WSL, not Git Bash: `yarn install`, `yarn upgrade`, and `yarn build`, all still work correctly after Cypress has patched Windows (after running any one Cypress script from package.json).  Then via Git Bash, Cypress scripts still work.
- Windows/Git Bash only: There are no ALSA errors in the console on running any Cypress scripts (should not be an issue, but certainly was on earlier attempts to run Cypress under Ubuntu)
- Windows/Git Bash only: There are no permissions errors on attempting to run a test again if it failed the first time. (Cypress possesses the necessary permissions on Windows to remove failure artifacts on the second test run) (again, should not be an issue, but certainly was on earlier attempts to run Cypress under Ubuntu)
- If a test fails due to an image comparison, it is NOT due to the window width. (there is a 15 pixel difference between Windows and Linux/MacOS instances of Chrome)


justin-bits:
Note: On the very first execution under Windows, the patch will only be applied on running `yarn test:cypress`, `yarn test:cypress-all` or `yarn test:cypress-smoke`.  So, it doesn't work under `yarn test:cypress-spec` or `yarn test:cypress-loop`.  I updated the documentation to clarify and removed wording indicating that it would occur on any Cypress script execution.

justin-bits:
- Changed patch script from `sh` to `cmd` to fix permissions errors that were occurring when attempting to delete `yarn` artifacts.
- Getting `The file cannot be accessed by the system.` on attempt to copy `run-os` and `run-script-os`.  Need to investigate.

justin-bits:
There was already a file present.  Updated script to remove it first. Issue resolved.

justin-bits:
Discovered image comparisons all failed under GUI when running under Windows due to a 15 pixel window size difference.  Fixed.

justin-bits:
With this change https://github.com/ngageoint/opensphere-build-resolver/pull/46, this PR is largely OBE.  I'll submit a new PR with the change for the image comparisons under Windows. Closed.

Issue #697, opened 2019-07-22
Update tests to reflect push pin fix


justin-bits:
https://github.com/ngageoint/opensphere/pull/692
https://github.com/ngageoint/opensphere/issues/676

PR #714, opened 2019-08-07
Fix menu toggle

schmidtk:
The menu toggle behavior was implemented with a blacklist of elements that shouldn't toggle, and instead always open a new menu. The grid wasn't in the blacklist and doesn't support toggling the menu, causing the menu to be closed while leaving the open flag set to `true`. This changes the behavior to support menu toggle on buttons and button groups.

Updated Cypress tests to remove the workaround for #667, and updated the screenshot following the default icon fix for #676.


justin-bits:
Export test fails: 
1) Feature list Context menu Export:
    Uncaught TypeError: Cannot read property 'firstElementChild' of undefined

justin-bits:
Also, #697 can probably be closed with this PR.

schmidtk:
> Export test fails

That error is unrelated and will randomly happen on `master`, see #716.

PR #717, opened 2019-08-08
Add Cypress Windows Support under Git Bash

justin-bits:
Changes:

- Added support for image comparisons under Windows.  Image comparisons are off by 15 pixels under Windows, and was throwing `Error: Image size (1279x1000) different than saved snapshot size (1264x1000)`.
- Added workaround for a bug with http-server where Cypress was stuck in an endless redirect loop under Windows http-party/http-server#525
- Added a Windows compatible method of detecting running instances of a web server and terminating those instances automatically or on demand (via `yarn stop-server`)

Tests:

Note: There may be failures with the test scripts, those are fixed here: #735 

- Windows: Cypress tests running via GUI (`npm run cypress:test`) AND CLI (any other cypress script; e.g. `npm run cypress:test-spec smoke-tests/smoke-test.spec.js`) do not immediately fail with a redirect error.  Only need to try one test.
- Linux and MacOS: Web server can be stopped via `yarn stop-server`. Windows: Web server can be stopped via `npm run stop-server`. (not using `yarn stop-server` under Windows due to https://github.com/yarnpkg/yarn/issues/5699)
- Windows: Cypress tests startup without web server already running AND with web server already running (GUI or CLI, doesn't matter)
- Windows and Linux/MacOS: Test scripts containing image comparisons work when launched via GUI. (if there is an image failure, it isn't failing due to snapshot size difference)

### DIFF
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -12,6 +12,7 @@
 // the project's config changing)
 
 var addMatchImageSnapshotPlugin = require('cypress-image-snapshot/plugin').addMatchImageSnapshotPlugin;
+os = require('os');
 
 /** This is a description of the foo function.
  *
@@ -22,7 +23,11 @@ module.exports = function(on, config) {
   addMatchImageSnapshotPlugin(on, config);
   on('before:browser:launch', function(browser, args) {
     if (browser.name === 'chrome') {
-      args.push('--window-size=1280,1200');
+      if (os.platform() === 'win32') {
+        args.push('--window-size=1295,1200');
+      } else {
+        args.push('--window-size=1280,1200');
+      }
       return args;
     }
 
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -12,7 +12,7 @@
 // the project's config changing)
 
 var addMatchImageSnapshotPlugin = require('cypress-image-snapshot/plugin').addMatchImageSnapshotPlugin;
-os = require('os');
+var os = require('os');
 
 /** This is a description of the foo function.
  *
--- a/cypress/support/execute-tests.sh
+++ b/cypress/support/execute-tests.sh
@@ -131,7 +131,7 @@ function overrideSettings() {
 
 function startWebServer() {
   if [ "$OSTYPE" == "msys" ]; then
-    webServerProcess="$(netstat -ano | findstr 0.0.0.0:8282 | awk '{print $5}')"
+    webServerProcess="$(netstat -ano | findstr 0.0.0.0:8282 | awk '{print $5}')" # TODO: Use a process name instead after this is fixed: https://github.com/http-party/http-server/issues/333
   else
     webServerProcess=$(ps -ef | grep http-server | grep -v grep)
   fi
@@ -198,7 +198,7 @@ function stopWebServer() {
   if $SERVER_STARTED; then
     echo 'INFO: terminating web server'
     if [ "$OSTYPE" == "msys" ]; then
-      webServerProcess="$(netstat -ano | findstr 0.0.0.0:8282 | awk '{print $5}')"
+      webServerProcess="$(netstat -ano | findstr 0.0.0.0:8282 | awk '{print $5}')" # TODO: Use a process name instead after this is fixed: https://github.com/http-party/http-server/issues/333
       taskkill //PID $webServerProcess //F
     else
       npm run stop-server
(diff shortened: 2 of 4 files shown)


justin-bits:
New issue found that should be added to this PR before merge:

The method of detecting the webserver is failing `cypress/support/execute-tests.sh: line 194: pgrep: command not found` resulting in attempting to start it again when it's already running: `Error: listen EADDRINUSE 0.0.0.0:8282`

justin-bits:
Webserver start/stop changes added.  Should be good to review again.

justin-bits:
Squashed commits, made remaining commit message comply.  Build passing now.

cypress/plugins/index.js
```diff
@@ -12,6 +12,7 @@
 // the project's config changing)
 
 var addMatchImageSnapshotPlugin = require('cypress-image-snapshot/plugin').addMatchImageSnapshotPlugin;
+os = require('os');
```
wallw-bits: Shouldn't this be `var os = ...`?

justin-bits: Yes.  Fixed.

cypress/plugins/index.js
```diff
@@ -22,7 +23,11 @@ module.exports = function(on, config) {
   addMatchImageSnapshotPlugin(on, config);
   on('before:browser:launch', function(browser, args) {
     if (browser.name === 'chrome') {
-      args.push('--window-size=1280,1200');
+      if (os.platform() === 'win32') {
+        args.push('--window-size=1295,1200');
+      } else {
+        args.push('--window-size=1280,1200');
+      }
```
wallw-bits: What's the reason for the difference here?

cypress/support/execute-tests.sh
```diff
@@ -130,7 +130,11 @@ function overrideSettings() {
 }
 
 function startWebServer() {
-  webServerProcess=$(ps -ef | grep http-server | grep -v grep)
+  if [ "$OSTYPE" == "msys" ]; then
+    webServerProcess="$(netstat -ano | findstr 0.0.0.0:8282 | awk '{print $5}')"
```
wallw-bits: This is less specific than the other command and it is possible that it could kill items that are not `http-server`. Is there a way to do this by the name of the process?

cypress/support/execute-tests.sh
```diff
@@ -191,15 +195,16 @@ function runTests() {
...
 
 function stopWebServer() {
-  if pgrep "node" > /dev/null; then
-    if $SERVER_STARTED; then
-      echo 'INFO: terminating web server'
-      npm run stop-server
+  if $SERVER_STARTED; then
+    echo 'INFO: terminating web server'
+    if [ "$OSTYPE" == "msys" ]; then
+      webServerProcess="$(netstat -ano | findstr 0.0.0.0:8282 | awk '{print $5}')"
```
wallw-bits: Same here.

cypress/support/execute-tests.sh
justin-bits: Yes it can be done by name of process, but under Windows that is reported as node.exe.  So I'd have to kill all of those. This would help: https://github.com/http-party/http-server/pull/515

cypress/plugins/index.js
justin-bits: I'm not sure.

cypress/support/execute-tests.sh
justin-bits: I updated the code to link to that issue so that this can be changed in the future.

cypress/plugins/index.js
wallw-bits: I would prefer that we use [cy.viewport](https://docs.cypress.io/api/commands/viewport.html#Syntax) or [set it in config](https://docs.cypress.io/guides/references/configuration.html#Viewport). That should be the most consistent method. This appears to be taking the window skin into account, and that will not be consistent if the user has customized their theme.

justin-bits: `cypress-image-snapshot` doesn't respect `cy.viewport`.  I'm fairly certain I tried to get this to work initially this way as it would be a cleaner approach.  If you see a way to do this, let me know.

The config is static and already set.  I'm not sure how changes here could be dynamic.

Also, I think (but have not tested) Cypress' instance of Chrome might not be affected by user changes to window skin.  It's definitely a modified version of Chrome that doesn't seem to pickup changes from the native instance.

image

PR #723, opened 2019-08-14
perf(columns): only persist non-default values for columns

jsalankey:
This significantly reduces the footprint of our descriptors in settings by only persisting values that differ from the default for columns. Previously we were persisting dozens of key/value pairs that just wasted time and space.

### DIFF
--- a/src/os/data/columndefinition.js
+++ b/src/os/data/columndefinition.js
@@ -182,33 +182,66 @@ os.data.ColumnDefinition.prototype.persist = function(opt_obj) {
     opt_obj = {};
   }
 
+  // always persist/restore these as they are the meat of the column definition
   opt_obj['id'] = this['id'];
   opt_obj['name'] = this['name'];
   opt_obj['field'] = this['field'];
-  opt_obj['type'] = this['type'];
-  opt_obj['resizable'] = this['resizable'];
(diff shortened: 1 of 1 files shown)


src/os/data/columndefinition.js
```diff
@@ -221,30 +254,62 @@ os.data.ColumnDefinition.prototype.restore = function(config) {
...
+  }
+  if (config.hasOwnProperty('toolTip')) {
+    this['toolTip'] = config['toolTip'];
+  }
+  if (config.hasOwnProperty('derivedFrom')) {
+    this['derivedFrom'] = config['derivedFrom'];
+  }
+  if (config.hasOwnProperty('userModified')) {
+    this['userModified'] = config['userModified'];
+  }
```
wallw-bits: This works, but it might be worthwhile for maintainability here to create a new "default" instance and loop over all the fields and compare them to the default. Then you do not have the default values listed in two places.

jsalankey: Good call, I'll switch to something like that.

bradh:
Was this really intended to be merged?

wallw-bits:
It was, which is why I approved it while still providing some feedback.

Issue #724, opened 2019-08-15
Add jest dependency


bradh:
yarn reports:

```
warning "workspace-aggregator-7997fbc4-b19f-4ddf-aac2-ee6a8b8fd831 > opensphere > cypress-image-snapshot > jest-image-snapshot@2.8.2" has unmet peer dependency "jest@>=20 <=24".
```

schmidtk:
[`cypress-image-snapshot`](https://www.npmjs.com/package/cypress-image-snapshot) is using code from `jest-image-snapshot` in a way that doesn't depend on `jest`. This warning can be ignored.

PR #726, opened 2019-08-15
chore(deps): add missing jest peer dependency

bradh:


schmidtk:
As mentioned in #724, I don't believe this is necessary.

PR #733, opened 2019-08-15
fix(wfs): tighten up parsing and handling of WFS URLs and output formats

estadtherr:
Trying to integrate with a [MapServer](https://mapserver.org/ogc/wfs_server.html) WFS server uncovered some discrepancies in the way the capabilities metadata was parsed and handled.

MapServer provides the URLs in the operations metadata in the following way:

```xml
    <ows:Operation name="GetFeature">
      <ows:DCP>
        <ows:HTTP>
          <ows:Get xlink:type="simple" xlink:href="https://osm-wfs-eric.dev.geointservices.io/wfs?"/>
          <ows:Post xlink:type="simple" xlink:href="https://osm-wfs-eric.dev.geointservices.io/wfs?"/>
        </ows:HTTP>
      </ows:DCP>
      …
    </ows:Operation>
```
This failed to parse correctly in the baseline logic, which assumes the `href` attribute will be the first attribute:
```javascript
         // Attr.value is the DOM4 property, while Attr.nodeValue inherited from Node should work on older browsers
        var attr = getFeatureEl.attributes[0];
        this.setWfsUrl(attr.value || attr.nodeValue);
```
The fix is to explicitly look for the `href` attribute by namespace and name, and fall back to the baseline logic if that fails.

MapServer also does not support JSON output format, which is the OpenSphere default. MapServer does support GML3, included in the `GetFeature` metadata in the following way:
```xml
    <ows:Operation name="GetFeature">
      …
      <ows:Parameter name="outputFormat">
        <ows:Value>text/xml; subtype=gml/3.1.1</ows:Value>
      </ows:Parameter>
    </ows:Operation>
```
OpenSphere did not recognize the GML3 support, however, due to a discrepancy in the regular expression used to identify the format:
```javascript
plugin.ogc.wfs.WFSLayerConfig.PREFERRED_TYPES = [
  /^application\/json$/,
  /gml\\?3/i,
  /gml\\?2/i
];
```
This will look for `gml\3` instead of `gml/3` in the format string. The fix is to correct the backslash to a slash.
Additionally, the logic in `plugin.ogc.wfs.WFSLayerConfig.prototype.getBestType` failed to recognize that the default format type (`application/json`) was not supported by the layer. I added logic to that function to double-check the inclusion of the defaulted format in the layer-supported formats.

### DIFF
--- a/src/os/ui/ogc/ogcserver.js
+++ b/src/os/ui/ogc/ogcserver.js
@@ -9,6 +9,7 @@ goog.require('goog.log');
 goog.require('goog.log.Logger');
 goog.require('ol.array');
 goog.require('ol.format.WMSCapabilities');
+goog.require('ol.format.XLink');
 goog.require('os.alert.AlertEventSeverity');
 goog.require('os.alert.AlertManager');
 goog.require('os.color');
@@ -182,13 +183,6 @@ os.ui.ogc.OGCServer.LOGGER_ = goog.log.getLogger('os.ui.ogc.OGCServer');
  * @type {string}
  */
 os.ui.ogc.OGCServer.DEFAULT_WMS_VERSION = '1.3.0';
-
-
-/**
- * @const
- * @type {string}
- */
-os.ui.ogc.OGCServer.XLINK_NAMESPACE = 'http://www.w3.org/1999/xlink';
 
 
 /**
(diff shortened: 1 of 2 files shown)


src/os/ui/ogc/ogcserver.js
```diff
@@ -856,16 +857,24 @@ os.ui.ogc.OGCServer.prototype.parseWfsCapabilities = function(response, uri) {
...
-          // Attr.value is the DOM4 property, while Attr.nodeValue inherited from Node should work on older browsers
-          var attr = getFeatureEl.attributes[0];
-          this.setWfsUrl(attr.value || attr.nodeValue);
+          if (getFeatureEl.hasAttributeNS(ol.format.XLink.NAMESPACE_URI, 'href')) {
+            this.setWfsUrl(getFeatureEl.getAttributeNS(ol.format.XLink.NAMESPACE_URI, 'href'));
+          } else {
+            var attr = getFeatureEl.attributes[0];
+            // Attr.value is the DOM4 property, while Attr.nodeValue inherited from Node should work on older browsers
+            this.setWfsUrl(attr.value || attr.nodeValue);
+          }
```
wallw-bits: I was under the impression that we should replace the old method. Was there a case where that is still necessary?
```
this.setWfsUrl(getFeatureEl.getAttributeNS(ol.format.XLink.NAMESPACE_URI, 'href') || this.getWfsUrl());
```

estadtherr: That would certainly simplify the logic, but there was some safety inherent in the old method in legacy cases where the capabilities document didn't strictly follow the schema. I'm happy to make the change if you're good elevating the strictness of the parsing like that!

PR #742, opened 2019-08-20
fix(ng): use $evalAsync instead of $apply to avoid $rootScope:inprog

### LINKED ISSUE #736: rootScope error thrown on setting date in toolbar-right.spec.js test
```
angular.min.js:127 Uncaught Error: [$rootScope:inprog] http://errors.angularjs.org/1.7.8/$rootScope/inprog?p0=%24apply
```

image

An exception has been added to the tests to ignore the error until this is fixed.

schmidtk:
Resolves #736.

There are a lot of discussions over `$timeout` vs `$evalAsync`, but it seems the latter is better for avoiding the `$rootScope:inprog` error that we can get calling `$apply` directly.

We could possibly replace the local vendor lib with individual libs (like [`angular-ui-date`](https://www.npmjs.com/package/angular-ui-date) for this one), but that's a much larger task than fixing this bug in our copy of the library.

### DIFF
--- a/cypress/integration/smoke-tests/toolbar-right.spec.js
+++ b/cypress/integration/smoke-tests/toolbar-right.spec.js
@@ -100,15 +100,6 @@ describe('Toolbar right', function() {
     });
 
     it('Disable state', function() {
-      // Catch console error we don't care about
-      cy.on('uncaught:exception', function(err) {
-        if (err.message.includes('$rootScope')) {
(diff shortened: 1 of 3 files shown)


schmidtk:
It appears my editor cleaned up extra whitespace in `angular-ui.js`. The relevant change is on line 1092.

PR #744, opened 2019-08-21
Add multiple projection support

justin-bits:
Adds support for overriding the default projection EPSG:3857 Web Mercator with EPSG:4326 Geographic. The override occurs during Cypress startup and reverts immediately after the tests complete (pass or fail).

- The override is accomplished by setting the environment variables: CYPRESS_PROJECTION=4326, STREET_MAP_URL=https://your.domain.com, WORLD_IMAGERY_URL=https://your.domain.com
- If no override variables are present, the defaults will work as is for EPSG:3857.
- Images in /comparisons are now stored under a projection specific folder

Note: This contains changes from ngageoint#735, merge that first.


justin-bits:
- Images for 4326 coming
- Support for using default URLs coming

justin-bits:
Images added.  Default URLs added.  Ready for review.

justin-bits:
Fixed minor issue with attempt to terminate non-existent webserver.

justin-bits:
The default URL for world imagery isn't compatible with EPSG:4326.  This needs to be changed or world imagery needs to be removed altogether.

justin-bits:
World imagery default URL updated.

justin-bits:
In `execute-tests.sh`, `sed` is being used in a way that doesn't work cross platform.  Need to investigate.

justin-bits:
`sed` changed out, works on Windows.  Need to test on Linux and MacOS.

justin-bits:
- `sed` changes verified to work under MacOS.
- Added minor fix for `stop-server` script that was discovered to not run under MacOS.

PR #745, opened 2019-08-22
Move filter/feature action import code to importers

schmidtk:
The code used to import filters and feature actions was within the UI controllers, preventing reuse of the code outside those UI's. To allow importing these elsewhere, all import code has been extracted to importers and the UI's now have a `getImporter` function to provide the appropriate importer.

To test, verify filters and feature actions import as they did before.

Additional changes:
* The filter import UI will no longer show the check/X icons until a layer has been selected, or the text explaining the red X. This is intended to avoid potential confusion because the UI previously showed the red X when no layer was selected.
* The layer picker was using the layer/descriptor `getId` function in the Angular template, and those functions are not exported. This prevented the layer function from working in the compiled app, though the descriptor function was not minified so that part was working. Active layers will now appear in the select dropdown in the compiled app.

### DIFF
--- a/src/os/ui/filter/im/filterimport.js
+++ b/src/os/ui/filter/im/filterimport.js
@@ -6,6 +6,7 @@ goog.require('os.im.Importer');
 goog.require('os.implements');
 goog.require('os.ui.Module');
 goog.require('os.ui.filter');
+goog.require('os.ui.filter.im.FilterImporter');
 goog.require('os.ui.filter.im.filterImportModelDirective');
 goog.require('os.ui.filter.parse.FilterParser');
 goog.require('os.ui.layer.layerPickerDirective');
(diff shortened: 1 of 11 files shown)


src/os/ui/filter/im/filterimport.js
```diff
@@ -208,151 +216,24 @@ os.ui.filter.im.FilterImportCtrl.prototype.init_ = function() {
...
-        if (filterable) {
-          // we matched it by filter key, so clone the filter and set the internal filterable ID as its type
-          var clone = filter.clone();
-          var type = filterable.getFilterableTypes()[0];
-          clone.setId(goog.string.getRandomString());
-          clone.setType(type);
+  if (this.importer) {
+    // assign all the display values
+    this['matched'] = this.importer.matched;
+    this['unmatched'] = this.importer.unmatched;
```
jsalankey: The `unmatched` property was originally an Array on the `FilterImportCtrl`, but the importer now has it as an Object. Is that intentional?

schmidtk: Ah, good call that was not intentional and has been fixed.

PR #749, opened 2019-09-03
build(windows): fix font path in Windows

wallw-bits:
The old setup was only suitable if a project had a symlink from `project/vendor/fonts` to `opensphere/vendor/fonts`. Symlinks do not work in Windows.

Therefore, this change uses a well-defined path to opensphere within the workspace.

Note that I do not believe that this method is compatible with using opensphere as a library within `node_modules`. That will need to be a follow-on for later investigation.

### DIFF
--- a/scss/fontloader.scss
+++ b/scss/fontloader.scss
@@ -17,7 +17,7 @@
 
   // Replace the url to be local
   $web-font-path: str-replace($web-font-path,
-    'https://fonts.googleapis.com/css?family=', '../../vendor/fonts/typeface-');
+    'https://fonts.googleapis.com/css?family=', '../../../opensphere/vendor/fonts/typeface-');
 
   // Replace + with -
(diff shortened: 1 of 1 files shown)


scss/fontloader.scss
```diff
@@ -17,7 +17,7 @@
 
   // Replace the url to be local
   $web-font-path: str-replace($web-font-path,
-    'https://fonts.googleapis.com/css?family=', '../../vendor/fonts/typeface-');
+    'https://fonts.googleapis.com/css?family=', '../../../opensphere/vendor/fonts/typeface-');
```
gregroyal: why does this work, but the current version not work in windows?

schmidtk: This uses a direct path to `opensphere`, instead of a symlink in `<project>/vendor/fonts`. The issue was with symlinks not working on Windows.

PR #751, opened 2019-09-09
fix(measure): measure line should be absolutely positioned

wallw-bits:
Measure lines should be drawn on the mean ellipsoid surface since that is where the calculation is occurring. If we want the line clamped to terrain then we will need to pull the highest quality meshes that intersect the line, clamp the line, and calculate distance from that.

### DIFF
--- a/src/os/interaction/measureinteraction.js
+++ b/src/os/interaction/measureinteraction.js
@@ -14,10 +14,12 @@ goog.require('ol.style.Style');
 goog.require('ol.style.Text');
 goog.require('os.bearing');
 goog.require('os.config.Settings');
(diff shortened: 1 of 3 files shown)


schmidtk:
This accurately depicts what is being measured, but the user interaction in 3D with terrain is not ideal. The mouse cursor is almost always disjoint from the line being drawn. Should 3D/terrain measurements switch to straight line distance and use the exact coordinate at each mouse event?

wallw-bits:
That's possible but we'll need to update the measurement as the terrain is loaded and each vertex is moved due to `clampToGround`.

PR #753, opened 2019-09-20
Add boolean filters

### LINKED ISSUE #645: Support boolean field types in filters
The filter stack currently only handles number and string types. Add support for boolean types and add the operations "is true" and "is false". "is empty" should also be available for boolean types.

In addition to unit tests, add a new layer to GeoServer with a boolean field and filter on that.

glennkitchellcaci:
Closes #645

### DIFF
--- a/src/os/ui/filter/op/isfalseop.js
+++ b/src/os/ui/filter/op/isfalseop.js
@@ -15,8 +15,8 @@ os.ui.filter.op.IsFalse = function() {
   os.ui.filter.op.IsFalse.base(this, 'constructor',
       'And', 'is false', 'false',
       [os.xsd.DataType.BOOLEAN, os.xsd.DataType.INTEGER, os.xsd.DataType.STRING],
-      'hint="os.ui.filter.op.IsFalse"', 'Supports false, 0, and "false" (case insensitive)', 'span', true);
-  this.matchHint = 'os.ui.filter.op.IsFalse';
+      'hint="is false"', 'Supports false, 0, and "false" (case insensitive)', 'span', true);
+  this.matchHint = 'is false';
 };
 
 
--- a/src/os/ui/filter/op/isfalseop.js
+++ b/src/os/ui/filter/op/isfalseop.js
@@ -29,8 +29,7 @@ goog.inherits(os.ui.filter.op.IsFalse, os.ui.filter.op.Op);
  * @inheritDoc
  */
 os.ui.filter.op.IsFalse.prototype.getEvalExpression = function(v, literal) {
-  return '(!(typeof ' + v + '==="undefined"||' + v + '==null||' + v + '.length==0)&&(' +
-      v + '===false||' + v + '==0||(""+' + v + ').toLowerCase()=="false"))';
+  return '(' + v + '===false||' + v + '===0||String(' + v + ').toLowerCase()==="false")';
 };
 
 
--- a/src/os/ui/filter/op/isfalseop.js
+++ b/src/os/ui/filter/op/isfalseop.js
@@ -1,0 +1,62 @@
+goog.provide('os.ui.filter.op.IsFalse');
+
+goog.require('os.ui.filter.op.Op');
+
+
+/**
+ * A 'PropertyIsFalse' operation class.
+ *
+ * @extends {os.ui.filter.op.Op}
+ * @constructor
+ */
+os.ui.filter.op.IsFalse = function() {
+  os.ui.filter.op.IsFalse.base(this, 'constructor',
+      'And', 'is false', 'false', undefined, undefined, 'Supports false, 0, and "false"', 'span', true);
+  this.matchHint = 'false.operation';
+};
+goog.inherits(os.ui.filter.op.IsFalse, os.ui.filter.op.Op);
+
+/**
+ * Because OpenSphere allows isEmpty() comparator for Boolean type properties, fail null/empty tests
+ *
+ * @inheritDoc
+ */
+os.ui.filter.op.IsFalse.prototype.getEvalExpression = function(v, literal) {
+  return '(!(typeof ' + v + '==="undefined"||' + v + '==null||' + v + '.length==0)&&(' +
+      v + '===false||' + v + '==0||(""+' + v + ').toLowerCase()=="false"))';
+};
+
+/**
+ *
+ * @inheritDoc
+ */
+os.ui.filter.op.IsFalse.prototype.getFilter = function(column, literal) {
+  var f = [];
+
+  f.push('<' + this.localName + '>');
+
+  f.push(
+      '<Not><PropertyIsNull>' +
+        '<PropertyName>' + column + '</PropertyName>' +
+      '</PropertyIsNull></Not>'
+  );
+
+  f.push('<Or>');
+  f.push(
+      '<PropertyIsEqualTo>' +
+        '<PropertyName>' + column + '</PropertyName>' +
+        '<Literal><![CDATA[0]]></Literal>' +
+      '</PropertyIsEqualTo>'
+  );
+  f.push(
+      '<PropertyIsEqualTo matchcase="false">' +
+        '<PropertyName>' + column + '</PropertyName>' +
+        '<Literal><![CDATA[false]]></Literal>' +
+      '</PropertyIsEqualTo>'
+  );
+  f.push('</Or>');
+
+  f.push('</' + this.localName + '>');
+
+  return f.join('');
+};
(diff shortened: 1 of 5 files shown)


glennkitchellcaci:
Test on Surge:
http://madly-recess.surge.sh/opensphere/

Tests to Verify:
1. Add Kube Master server to data sources
1a. Click Add Data > Manage Servers > Paste https://kubemaster-trunk-1.udev.six3/ogc/wmsServer and click OK > Change Name to "Kube Master" > Set Feature Server to https://kubemaster-trunk-1.udev.six3/ogc/wfsServer 
2. Show Twitter data layer (Blue Marble will disappear)
3. Select and load data in part of the blob that looks like Japan
3a. Write down the number of features loaded
4. Add Filter > Name: "Hide Truncated" > Expression: TRUNCATED "is false"
4a. Apply the filter and write down the number of features loaded
4b. Double-click some features to verify TRUNCATED is false
5. Edit the Filter > Name: "Show Truncated ONLY" > Expression: TRUNCATED "is true"
5a. Verify the number of features loaded is the difference between the numbers in 3a and 4a
5b. Double-click some features to verify TRUNCATED is true

src/os/ui/filter/op/istrueop.js
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * A 'PropertyIsTrue' operation class.
+ * Based on the OGC Filter Spec
+ *
+ * @extends {os.ui.filter.op.Op}
+ * @constructor
+ */
+os.ui.filter.op.IsTrue = function() {
+  os.ui.filter.op.IsTrue.base(this, 'constructor',
+      'And', 'is true', 'true', undefined, undefined, 'Supports true, >1, and "true"', 'span', true);
```
bradh: `>=1` ?

glennkitchellcaci: Ah, yup that comment is doubly wrong.  I changed it to ==1 to make the implementation less flexible; since this isn't implemented like a coding "boolean" (where non-nulls/non-empty's are true unless explicitly false, etc).

src/os/ui/filter/op/isfalseop.js
```diff
@@ -0,0 +1,62 @@
...
+ * @extends {os.ui.filter.op.Op}
+ * @constructor
+ */
+os.ui.filter.op.IsFalse = function() {
+  os.ui.filter.op.IsFalse.base(this, 'constructor',
+      'And', 'is false', 'false', undefined, undefined, 'Supports false, 0, and "false"', 'span', true);
+  this.matchHint = 'false.operation';
+};
+goog.inherits(os.ui.filter.op.IsFalse, os.ui.filter.op.Op);
+
```
jsalankey: Nit: for consistency, we prefer 2 newlines between top level code blocks. Unfortunately our linter does not enforce this currently (sorry).

src/os/ui/filter/op/isfalseop.js
```diff
@@ -0,0 +1,62 @@
...
+};
+
+/**
+ *
+ * @inheritDoc
+ */
+os.ui.filter.op.IsFalse.prototype.getFilter = function(column, literal) {
+  var f = [];
+
+  f.push('<' + this.localName + '>');
```
glennkitchellcaci: After discussion with @wallw-bits -- Determined that I need the matchHint, etc attributes in the root XML tag so the filter builder can parse back to "falseop" instead of complex "and...or...etc"

src/os/ui/filter/op/isfalseop.js
```diff
@@ -1,21 +1,28 @@
...
  * @constructor
  */
 os.ui.filter.op.IsFalse = function() {
   os.ui.filter.op.IsFalse.base(this, 'constructor',
-      'And', 'is false', 'false', undefined, undefined, 'Supports false, 0, and "false"', 'span', true);
-  this.matchHint = 'false.operation';
+      'And', 'is false', 'false',
+      [os.xsd.DataType.BOOLEAN, os.xsd.DataType.INTEGER, os.xsd.DataType.STRING],
+      'hint="os.ui.filter.op.IsFalse"', 'Supports false, 0, and "false" (case insensitive)', 'span', true);
+  this.matchHint = 'os.ui.filter.op.IsFalse';
```
jsalankey: It's generally preferably not to put a dot-delimited class name in for something like this. These are saved to the raw filter XML for the purposes of how we read and decompose filters out in the filter builder, and the class name could change in the future (or we could integrate with other software where the class name is meaningless).

For the other ops, we literally just do things like `this.matchHint = 'like list';`, `'is false'` would be fine here.

src/os/ui/filter/op/isfalseop.js
```diff
@@ -0,0 +1,71 @@
...
+goog.inherits(os.ui.filter.op.IsFalse, os.ui.filter.op.Op);
+
+
+/**
+ * Because OpenSphere allows isEmpty() comparator for Boolean type properties, fail null/empty tests
+ *
+ * @inheritDoc
+ */
+os.ui.filter.op.IsFalse.prototype.getEvalExpression = function(v, literal) {
+  return '(!(typeof ' + v + '==="undefined"||' + v + '==null||' + v + '.length==0)&&(' +
```
schmidtk: A few things here:
* `prop == null` covers both `null` and `undefined` values, so the `typeof v === 'undefined'` check is redundant. `typeof` is also much slower than a simple value test, so try to avoid it when possible.
* You should be able to simplify this test to `v === false || String(val).toLowerCase() === 'false' || v === 0`, and omit the null/undefined/length checks entirely.

src/os/ui/filter/op/istrueop.js
```diff
@@ -0,0 +1,70 @@
...
+
+goog.inherits(os.ui.filter.op.IsTrue, os.ui.filter.op.Op);
+
+
+/**
+ *
+ * @inheritDoc
+ */
+os.ui.filter.op.IsTrue.prototype.getEvalExpression = function(v, literal) {
+  return '(!(typeof ' + v + '==="undefined"||' + v + '==null||' + v + '.length==0)&&(' +
```
schmidtk: Same comment here. The expression `v === true || String(val).toLowerCase() === 'true' || v === 1` should cover all cases, and return false for all other values.

src/os/ui/filter/op/isfalseop.js
glennkitchellcaci: I recommend not removing the undefined check from code intended to be run through an eval() -- since any unhandled JS exception may cause all JS to cease running.  Also, exception handling is costly as well.

Try typing "blah === null" in your developer tools command line and verify that it doesn't return true or false -- it throws an exception.

This is relevant, because on many JSON-returning backend technologies, it's possible to prune properties that are null to save bandwidth - so some objects may not have all the properties.

Also, not all browsers and all versions of all browsers treat undefined the same way as the others.  Forcing the undefined check through the "typeof" spec is one way to make all browsers treat it the same way.  At one point, undefined wasn't even necessarily equal to null in all browsers -- thus the "redundant" check of "is undefined OR is null."

Since we're pretty much pushing "latest Chrome" and "latest Firefox" I know it's less of a concern with this team -- so let me know if we push code that utilizes features/behaviors that aren't in IE/Safari and/or less-recent versions of Chrome/Firefox.

Anyway, I'll research a newer/faster way of doing the undefined check without a string comparison.  This was just one way I learned that worked in all browsers without throwing exceptions.

schmidtk: All of your points are fair and valid, given either of the following is true:
* The root variable name being tested is not defined in the eval scope.
* The variable name causes an undefined reference exception. For example, `foo = {}` and variable name is `foo.bar.baz`.

OpenSphere assumes that code creating filter functions will protect against both of these. It's entirely possible a leaf property name may not exist on an object (ie, `foo.bar` in the above example), but that will not cause an exception with any of the tests we're doing here.

We make these assumptions because these filter functions could potentially be run against millions of objects, so we value performance with the exchange of putting some burden on the developer to protect against the exceptions you mentioned. If we were to change the core assumptions, we would have to update all of the filter operations to perform the additional undefined variable check.

glennkitchellcaci: Thanks for the break-down!  I was actually thinking about writing a utility function to do dot and array notation resolution on variables (just like you mentioned -- e.g. utils.isNotNull("foo.bar[lookup.idx].baz")) -- but you're right -- the speed requires a trade-off, and assuming there's at least some kind of check upstream then everything downstream can go faster.

PR #759, opened 2019-10-09
fix(arc): fix multipolygon parsing from oriented rings

### LINKED ISSUE #739: ArcGIS parses polygons incorrectly
ArcGIS defines outer/inner rings in a polygon by the winding order of the ring. Clockwise coordinates are outer rings, while counter-clockwise coordinates are inner rings. Update `plugin.arc.ArcJSONParser` to account for this and create an appropriate geometry.

erica-manzer:
resolves #739

Added functionality to compute the WindingOrder of a polygon and use that to sort internal and external Polygons into MultiPolygons. Updated or added corresponding utests.


wallw-bits:
This looks good! Thanks!

Issue #350, opened 2018-12-06
Improve Zip handling


wallw-bits:
We currently only handle zips as file types which are zips (KMZ, Shapefile). However, we should support zip contents that would otherwise be supported if unzipped, such as zipped CSV, GeoJSON, GML, states, 3D tiles, etc.

My initial thought is to check the content type of each zip listing. If there is only one that we support, unzip and import that. Otherwise, present the list of supported files to the user for them to select one and import it.

It would also be nice if the zip file was not duplicated in the filestore for each import.

PR #760, opened 2019-10-09
Zip file enhancements

glennkitchellcaci:
Resolves issue #350 

Added ZIP file support to Add Data > File > Browse... > Select a *.zip file.

Utility will unzip the file and allow User to import one or more of the files inside.

Testing steps:
1. Put a few CSV and KML files into a ZIP file.
2. Click Add Data > Open File/URL > Browse > Select the ZIP file
3. Click Next when contents are unzipped
4. Deselect/Select one or more files and click Done
5. Complete the regular Import Data process for each of the files selected


glennkitchellcaci:
Surge -- http://puzzled-record.surge.sh/opensphere/

src/plugin/file/zip/ui/zipprocessstep.js
```diff
@@ -0,0 +1,200 @@
...
+/**
+ * Controller for the ZIP import file selection step
+ *
+ * @param {!angular.Scope} $scope
+ * @param {!angular.JQLite} $element
+ * @param {!angular.$timeout} $timeout
+ * @constructor
+ * @ngInject
+ */
+plugin.file.zip.ui.ZIPProcessStepCtrl = function($scope, $element, $timeout) {
```
schmidtk: I'm not sure it's necessary to have separate steps for reading the ZIP and allowing the user to pick files. The file picker can tell them how many files are present in the ZIP, or if something bad happened, and reduces the import process by an extra click. I realize that doesn't seem like much, but keeping workflows simple where reasonable should always be a priority.

My suggestion would be to show a "Loading ZIP..." message with a spinner while the file loads, then switch to the checklist UI when loading completes.

src/plugin/file/zip/zip.js
```diff
@@ -0,0 +1,10 @@
+goog.provide('plugin.file.zip');
+
+
+/**
+ * @enum {number}
+ * @const
+ */
+plugin.file.zip.LOOKUP = {
```
schmidtk: This appears to be dead code. Remove?

src/plugin/file/zip/zipdescriptor.js
```diff
@@ -0,0 +1,47 @@
+goog.provide('plugin.file.zip.ZIPDescriptor');
```
schmidtk: Unless you plan on showing the ZIP file in Add Data, you don't need to create a descriptor for the overall ZIP. We're creating individual descriptors for each file imported from the ZIP, so this one should not be necessary.

src/plugin/file/zip/zipplugin.js
```diff
@@ -0,0 +1,65 @@
...
+
+
+/**
+ * @inheritDoc
+ */
+plugin.file.zip.ZIPPlugin.prototype.init = function() {
+  var dm = os.dataManager;
+
+  // register zip provider type
+  dm.registerProviderType(new os.data.ProviderEntry(
```
schmidtk: Given we aren't showing the ZIP file in Add Data, we do not need to create a provider (or a descriptor as previously noted).

src/plugin/file/zip/ui/zipprocessstep.js
```diff
@@ -0,0 +1,200 @@
...
+
+
+/**
+ * @param {number} val
+ * @param {number} oldVal
+ * @param {Object} scope
+ * @public
+ */
+plugin.file.zip.ui.ZIPProcessStepCtrl.prototype.onStatusChange = function(val, oldVal, scope) {
+//  this.validate_();
```
jsalankey: Are these meant to be commented out?

glennkitchellcaci: Ahh, those are old code that's no longer used.  I'm removing this entire step, so those will be gone soon too.

glennkitchellcaci:
At this moment I'm not sure how it was working before -- but somehow the Wizard blocked the deferred chain from completing until each chained Importer finished.

Now that I'm just extending from FileImportUI, all of the sub-imports are popping open at once.  I'll figure out what's going on when I get back on Monday.  Sorry for the delay.

src/os/data/columndefinition.js
```diff
@@ -292,6 +292,9 @@ os.data.ColumnDefinition.prototype.restore = function(config) {
   if (config.hasOwnProperty('selectable')) {
     this['selectable'] = config['selectable'];
   }
+  if (config.hasOwnProperty('formatter')) {
```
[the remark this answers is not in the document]
glennkitchellcaci: In one iteration of my code, I had needed it to quickly rebuild a typed ColumnDefinition object from a basic JS object.  Since then I removed that code and left this change since the restore() function didn't appear to be called by anything that would have a "formatter" on it.

For safety, I'll remove this change -- especially since I no longer need it.

src/plugin/file/zip/ui/zipfilesstep.js
```diff
@@ -0,0 +1,257 @@
...
+
+
+/**
+ * Controller for the ZIP import file selection step
+ *
+ * @param {!angular.Scope} $scope
+ * @constructor
+ * @ngInject
+ */
+plugin.file.zip.ui.ZIPFilesStepCtrl = function($scope) {
```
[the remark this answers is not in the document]
glennkitchellcaci: Removed!

src/plugin/file/zip/ui/zipprocessstep.js
glennkitchellcaci: Removed!

src/plugin/file/zip/zipdescriptor.js
glennkitchellcaci: Removed!

src/plugin/file/zip/zipplugin.js
glennkitchellcaci: Removed!

src/plugin/file/zip/zip.js
glennkitchellcaci: Removed!

glennkitchellcaci:
I added an additional Event listener to the FileImport callback and updated all of the imports to configure it on the config object.  Surge is up to date.

src/plugin/file/zip/ui/zipimport.js
```diff
@@ -0,0 +1,231 @@
...
+ * @return {number}
+ */
+plugin.file.zip.ui.ZIPImportCtrl.prototype.count = function() {
+  return (this['files'] && this['files'].length > 0)
+    ? this['files'].reduce(function(a, v) {
+      if (typeof a == 'object') a = (a.enabled ? 1 : 0);
+      if (v.enabled) a++;
+      return a;
+    })
+    : 0;
```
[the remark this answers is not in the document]
glennkitchellcaci: Niiice!!

src/os/ui/im/abstractimportui.js
```diff
@@ -45,3 +61,11 @@ os.ui.im.AbstractImportUI.prototype.mergeConfig = function(from, to) {
...
   to['title'] = from['title'];
 };
+
+
+/**
+ * @inheritDoc
+ */
+os.ui.im.AbstractImportUI.prototype.onClose = function() {
+  this.eventTarget.dispatchEvent(os.events.EventType.COMPLETE);
+};
```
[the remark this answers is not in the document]
glennkitchellcaci: Removed

src/plugin/file/zip/ui/zipimport.js
```diff
@@ -0,0 +1,295 @@
...
+  var maxWait = 80 * wait;
+
+  if (!this.config_) return;
+
+  var status = this.config_['status'];
+
+  this['loading'] = status != 0;
+  this['valid'] = !this['loading'];
+
+  // this.scope_.$emit(os.ui.wiz.step.WizardStepEvent.VALIDATE, this['valid']);
```
jsalankey: Extra comment here.

src/plugin/file/zip/ui/zipimport.js
```diff
@@ -0,0 +1,295 @@
...
+    if (this['wait'] < maxWait) this.timeout__(this.validate_.bind(this), wait);
+    else {
+      this['loading'] = false;
+      var err = 'Unzipping took too long in browser.  Try unzipping locally first.';
+      os.alert.AlertManager.getInstance().sendAlert(err, os.alert.AlertEventSeverity.ERROR);
+    }
+    this['wait'] += wait;
+  }
+
+  // os.ui.apply(this.scope_);
```
jsalankey: Comment here.

src/plugin/file/kml/mime.js
```diff
@@ -39,7 +41,38 @@ os.file.mime.register(
...
+ *
+ * @param {ArrayBuffer} buffer
+ * @param {os.file.File=} opt_file
+ * @param {*=} opt_context
+ * @return {!goog.Promise<*|undefined>}
+ */
+plugin.file.kml.detectKmz = function(buffer, opt_file, opt_context) {
+  var retVal;
+  var kmzRegex = /\.kmz$/i;
+  var kmlRegex = /\.kml$/i;
```
jsalankey: I am generally leery of filename tests for typing files. They tend to bite us with generate files as the naming schemes are often meant for machine to machine communication, so they leave off file extensions in the names. This might not technically be a valid thing for a KMZ to do, but if we can apply more robust checking of the zip file entries (i.e. our existing KML mime check which looks at the internal element names/namespaces).

wallw-bits: The case here is that it is difficult to determine whether a zip full of KML files is a single KMZ with multiple KML files (a single import) or an actual user-zipped set of several KMLs (zip multi import). The best distinction between those two is the file extension.

src/plugin/file/zip/mime.js
```diff
@@ -0,0 +1,19 @@
...
+plugin.file.zip.mime.TYPE = os.file.mime.zip.TYPE;
+
+
+/**
+ * @param {ArrayBuffer} buffer
+ * @param {os.file.File=} opt_file
+ * @param {*=} opt_context
+ * @return {!goog.Promise<*|undefined>}
+ */
+plugin.file.zip.detect = os.file.mime.zip.detectZip;
```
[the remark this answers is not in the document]
glennkitchellcaci: removed.

src/plugin/file/zip/ui/zipimport.js
```diff
@@ -0,0 +1,295 @@
...
+  };
+
+  this.cur_.importer = this.importers_.splice(0, 1)[0]; // remove the current importer from the queue
+
+  var process = new os.im.ImportProcess(); // new os.ui.im.DuplicateImportProcess();
+  process.setSkipDuplicates(true);
+  process.setEvent(new os.ui.im.ImportEvent(os.ui.im.ImportEventType.FILE, this.cur_.importer['file']));
+
+  // TODO because the duplicate dialog is created by setEvent(), the deferred is not called when
+  // the Duplicate dialog is cancelled... so just use ImportProcess for now
```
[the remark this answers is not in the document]
glennkitchellcaci: updated and moved this comment to above -- let me know if it needs more info

src/os/ui/im/importprocess.js
```diff
@@ -273,7 +273,7 @@ os.ui.im.ImportProcess.prototype.importFile = function(opt_config) {
       }
 
       if (this.deferred_) {
-        this.deferred_.callback(this.file);
+        this.deferred_.callback({'file': this.file, 'ui': ui});
```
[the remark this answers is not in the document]
glennkitchellcaci: Was needed for an old implementation back when this was a Wizard.  Removed.

src/plugin/file/kml/mime.js
glennkitchellcaci: Added the TYPE check.  Now I feel much more comfortable about this too, and it still passes the KML and KMZ mime.test.js tests.

src/plugin/file/zip/ui/zipimport.js
```diff
@@ -0,0 +1,310 @@
...
+  var unsupported = {};
+
+  for (var i = 0; i < entries.length; i++) {
+    var entry = entries[i];
+    if (entry.enabled === true) {
+      var file = entry.file;
+      var type = file.getType();
+      var ui = this.im_.getImportUI(type);
+      if (ui) {
+        var importer = /** @type osx.import.FileWrapper */ ({
```
wallw-bits: If you move the `FileWrapper` typedef into `src/`, then it will not be necessary to use quotes here. Newer syntax rules also let you do this: `{file, ui}`, which is equivalent to `{file: file, ui: ui}`, but I'm not sure when that was added so the compiler may yell at you.

src/plugin/file/zip/zipparser.js
```diff
@@ -0,0 +1,379 @@
...
+      file.setType(type);
+      return file;
+    }
+    return null;
+  };
+
+  var onFile = function(file) {
+    if (file) {
+      // turn this into a better object for the UI
+      return /** @type {osx.import.FileWrapper} */ ({
```
wallw-bits: Same here with the note about the quotes.

src/plugin/file/zip/zipparserconfig.js
```diff
@@ -0,0 +1,58 @@
...
+ * @param {Function} callback
+ */
+plugin.file.zip.ZIPParserConfig.prototype.update = function(callback) {
+  // re-initialize
+  if (this['files'].length > 0) this['files'] = [];
+  this['status'] = -1;
+
+  var parser = new plugin.file.zip.ZIPParser(this);
+
+  goog.events.listenOnce(parser, os.events.EventType.COMPLETE, goog.bind(function() {
```
wallw-bits: What happens if the parser errors instead of completing? I think this would be a memory leak in that scenario.

externs/os.externs.js
```diff
@@ -488,3 +488,24 @@ osx.ui.draw;
...
+  *   id: (string|undefined),
+  *   label: (string|undefined),
+  *   valid: (boolean|undefined),
+  *   enabled: (boolean|undefined),
+  *   msg: (string|undefined),
+  *   file: os.file.File,
+  *   ui: (os.ui.im.IImportUI|undefined)
+  * }}
+  */
+osx.import.FileWrapper;
```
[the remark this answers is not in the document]
glennkitchellcaci: Updated ZipImport chain() to use an internal typedef and removed the extra flexibility from this typedef so the ZipParser builds strongly-typed objects for ZipImport.html

src/plugin/file/zip/zipparserconfig.js
glennkitchellcaci: Moved this code into zipparser.js > unzip() <line 149> and added ERROR throw on all the big errors from the parser with a catch here too.  Should cover everything but JS crashes.

src/plugin/file/zip/zipparser.js
jsalankey: Since this is an extern typedef, using quotes around the keys is unnecessary as the compiler already knows not to minify those symbols.

externs/os.externs.js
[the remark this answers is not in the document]
jsalankey: If this type is being referenced from an HTML file, then a typedef is fine.

src/plugin/file/zip/ui/zipimport.js
```diff
@@ -0,0 +1,347 @@
...
+  var onFailure = function() {
+    this.chain(); // continue
+  };
+
+  this.curImporter_ = this.importers_.splice(0, 1)[0]; // remove the current importer from the queue
+
+  // TODO - Breaking Change -- Fix every place that uses DuplicateImportProcess to utilize events OR alter DuplicateImportProcess
+  // to handle the deferred object in all cases (without breaking any of the current uses).  Currently, the duplicate
+  // dialog is created by setEvent(), so the deferred is not called when that dialog is cancelled.  So for now, we
+  // have to use ImportProcess instead of DuplicateImportProcess here.
```
glennkitchellcaci: This is probably worth a follow-on ticket.  Basically the issue is that there's no feedback to the creator/caller of Importers which says, "I'm done now" -- with event types for imported, failed to read, failed to load, cancelled, etc.  To add it would break all the existing Importers; since we'd probably modify the interface or add the event(s) to the super class which would need to be overridden properly in some of the edge cases.

src/plugin/file/zip/ui/zipimport.js
```diff
@@ -0,0 +1,347 @@
...
+ */
+plugin.file.zip.ui.ZIPImportCtrl.prototype.chain = function() {
+  if (typeof this.importers_ == 'undefined' || this.importers_ == null || this.importers_.length == 0) {
+    if (this.curListener_) this.curListener_(); // kill the listener that was created by the previous chain()
+    this.close();
+    return;
+  }
+
+  var onSuccess = function() {
+    // TODO if there are other paths through importers in the future, then update this code
```
glennkitchellcaci: This is a warning to future developers -- This part of the code is a bit meh, as it only works "nicely" with os.ui.Windows that follow our Importer scope data structure.  Other types of popups or modal frameworks don't throw the same events, so the chain() will just kick off the next importer after 1.35 seconds so least all the file(s) they want will get imported, but it may be annoying to users to have more than one popup going at a time.  

The next TODO talking about making a breaking change to the Importers to enforce that they all throw events would fix this weakness.  That will likely be turned into a follow-on ticket.

src/plugin/file/zip/zipparser.js
glennkitchellcaci: Updated!  That'll save a some bytes.

PR #763, opened 2019-10-14
feat(cursor): add cursor pointer to scss utils

csgibson:


csgibson:
If it needs a pointer, it can almost always be an anchor element. This is not needed

PR #769, opened 2019-10-22
fix tuieditior zindex

jacob-hinkston:
fixed Tui zindex


wallw-bits:
@jacob-hinkston, the sass linter produced an error on that build. Can you please run the build locally and fix that?

PR #770, opened 2019-10-23
Layers Search Grid

glennkitchellcaci:
NOTE: This needs the additional plugin to be testable as part of the original ticket.  This capability can still be used in plain Opensphere when the "supportsGrid" flag on the DrawControlsCtrl is set to 'true'; or if a future plugin/feature wants to reuse this grid code as well.

Testing steps:

1. Set "supportsGrid" to true (this is NOT how it's committed), or use provided Surge link
2. Lower the brightness or opacity of the "Street Map Tiles" layer 
3. Draw a box, circle, and/or polygon
4. Remember, this is NOT how it's committed; the grid will NOT be shown normally
5. Verify a grid snapped to 0.25 degree lat/lon coordinates is drawn around the shape (this can be changed to 1-degree or 0.1-degree grid boxes by changing the setting in settings.json)
6. Verify the grid is not drawn when the search area is large

### DIFF
--- a/config/settings.json
+++ b/config/settings.json
@@ -146,7 +146,11 @@
       }
     },
     "search": {
-      "enableAll": true
+      "enableAll": true,
+      "grid" : {
+        "detail": 0.25,
+        "detailMax": 350
+      }
     },
     "facetedSearch": {
       "categories": {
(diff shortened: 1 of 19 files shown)


glennkitchellcaci:
http://happy-harbor.surge.sh/opensphere/

glennkitchellcaci:
Disregard.  Fixing the commit messages didn't work -- I'm going to create a new branch and re-apply my changes.

PR #771, opened 2019-10-23
Feature Grid

glennkitchellcaci:
NOTE: This needs the additional plugin to be testable as part of the original ticket. This capability can still be used in plain Opensphere when the "supportsGrid" flag on the DrawControlsCtrl is set to 'true'; or if a future plugin/feature wants to reuse this grid code as well.

Testing steps:

1. Set "supportsGrid" to true (this is NOT how it's committed), or use provided Surge link
2. Lower the brightness or opacity of the "Street Map Tiles" layer
3. Draw a box, circle, and/or polygon
4. Remember, this is NOT how it's committed; the grid will NOT be shown normally
5. Verify a grid snapped to 0.25 degree lat/lon coordinates is drawn around the shape (this can be changed to 1-degree or 0.1-degree grid boxes by changing the setting in settings.json)
6. Verify the grid is not drawn when the search area is large


glennkitchellcaci:
http://happy-harbor.surge.sh/opensphere/

config/settings.json
```diff
@@ -146,7 +146,11 @@
       }
     },
     "search": {
-      "enableAll": true
+      "enableAll": true,
+      "grid" : {
+        "detail": 1.0,
```
glennkitchellcaci: Talked to Shaun, set to 1.0 as the default instead of 0.25

src/os/ui/draw/draw.js
```diff
@@ -1,8 +1,122 @@
 goog.provide('os.ui.draw');
+goog.provide('os.ui.draw.utils');
```
jsalankey: The utils namespace is superfluous. We generally prefer to add utilities/static functions directly onto the root namespace, in this case things like `os.ui.draw.getGridFromFeature` would be fine.

glennkitchellcaci:
I like the idea of moving it down to GeoSearch since not all types of searches would benefit from being able to pass the config object back to the UI.  But, since igeosearch is just an add-on interface that doesn't have the nice abstract implementation from which all searches spawn that isearch has, it would be a breaking change to update that interface.  Right now, the grid is opt-in, with no changes required to existing searches -- whether seen or unseen from the OpenSphere repo.  Perhaps I could add a new interface?  Kinda kills the point of the extensible GridConfig object though.

Also, to make it a search-specific thing, I'd have to add some sort of completely-generic-config that all searches pass back.  This completely generic config would contain the same three settings that my GridConfig currently has, but could obviously send back any future configs for any feature.  Maybe that would be a better approach -- remake the gridconfig into a generic searchconfig object, so then it belongs in the top level of the search object hierarchy.  Do we have anything like that already?  Would that be a good workaround to the breaking changes required of messing with igeosearch?  

Or am I missing something?  I'll keep thinking about it just in case.  

Thanks for the feedback!

 - Glenn

wallw-bits:
> with no changes required to existing searches

Any change to an interface is a breaking change. There is no guarantee that all implementations of the interface extend the accompanying abstract class.

As for the implementation here, I'd rather see this as a utility function. Any geo search provider using grid-indexed data should already have a feature representing the search area. Then it can look up its own grid config (say we had several search providers with various grids) and pass the feature and the grid config to the utility function to get the resulting grid/area feature.

glennkitchellcaci:
The grid creation/management is moved now directly to the plugin and the interface+abstract changes are backed out.  Let me know if there's anything else that needs to be moved around!

test/os/ui/draw/draw.test.js
```diff
@@ -0,0 +1,41 @@
+goog.require('ol.Feature');
+goog.require('ol.geom.Polygon');
+goog.require('ol.proj');
+goog.require('os.map');
+goog.require('os.ui.draw');
+
+describe('os.ui.draw', function() {
```
wallw-bits: Thanks for writing tests!

test/os/ui/draw/draw.test.js
```diff
@@ -0,0 +1,41 @@
...
+  window.localStorage.clear();
```
wallw-bits: But I'd like to know why this was necessary.

test/os/ui/draw/draw.test.js
```diff
@@ -0,0 +1,41 @@
...
+  os.net.RequestHandlerFactory.addHandler(os.net.SameDomainHandler);
+
+  beforeEach(function() {
+    window.localStorage.clear();
```
wallw-bits: and this.

test/os/ui/draw/draw.test.js
```diff
@@ -0,0 +1,41 @@
+goog.require('ol.Feature');
+goog.require('ol.geom.Polygon');
+goog.require('ol.proj');
+goog.require('os.map');
+goog.require('os.ui.draw');
+
+describe('os.ui.draw', function() {
+  window.localStorage.clear();
+  os.net.RequestHandlerFactory.addHandler(os.net.SameDomainHandler);
```
wallw-bits: This should already be done, and you're also not using requests below so why is it here?

test/os/ui/draw/draw.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+  beforeEach(function() {
+    window.localStorage.clear();
+  });
+
+  it('should read grid settings as numbers', function() {
+    runs(function() {
+      var gridDetail = os.ui.draw.getGridSetting(os.ui.draw.GRID_DETAIL, 0.1);
+      expect(gridDetail).toBeDefined();
+      expect(typeof gridDetail == 'number').toBeTruthy();
```
wallw-bits: While this works, if this were to fail, the log would contain "Expected 'false' to be 'true'". However, if you write this as `expect(typeof gridDetail).toBe('number')` then the log would be something like "Expected 'string' to be 'number'", which is closer to the test that is actually happening here.

Generally, if you find yourself putting `==` or `===` in the expectation, you probably mean to use `expect(a).toBe(b)` (for non-strict equality or deep equality checks, use `expect(a).toEqual(b)`).

src/os/ui/draw/draw.js
```diff
@@ -1,8 +1,166 @@
...
+
+/**
+ * Helper function to get valid default settings for grid capabilities
+ *
+ * @param {number=} detail The number of degrees squared used to draw the grid
+ * @param {number=} max The maximum number of grid squares to draw
+ * @param {ol.style.Style=} style The style applied to the squares of the grid
+ * @return {osx.ui.draw.GridOptions}
+ */
+os.ui.draw.gridOptionsInstance = function(detail, max, style) {
```
wallw-bits: I don't really like this here. I think it would be better to do the input checking in the function using it (`getGridFromFeature`). 

Since you only have 3 items that make up a `GridOptions` instance, those could just be parameters to `getGridFromFeature`. Not saying that's required, just that it would be simpler in this case.

src/os/ui/draw/draw.js
```diff
@@ -1,8 +1,166 @@
...
+
+
+/**
+ * Helper function; gets a numeric representation of the JSON setting
+ *
+ * @param {string} key
+ * @param {number} defaultValue
+ * @return {number}
+ */
+os.ui.draw.getGridSetting = function(key, defaultValue) {
```
wallw-bits: Let's get rid of the settings stuff entirely in here. Anything that wants to use `os.ui.draw.getGridFromFeature()` should be responsible for creating its own options.

config/settings.json
```diff
@@ -36,6 +36,10 @@
     "metrics": {
       "enabled": true
     },
+    "grid" : {
+      "detail": 0.25,
+      "detailMax": 350
+    },
```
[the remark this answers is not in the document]
glennkitchellcaci: It was used as system-wide defaults here; removed.

test/os/ui/draw/draw.test.js
glennkitchellcaci: Rewrote to no longer pull an actual setting; now it uses a true .toBe as recommended.

src/os/ui/draw/draw.js
glennkitchellcaci: I left this to simplify my code, but valid alternatives are to:
1. Add a "getAs()" function to settings.js which would accept a casting function, OR
2. Add a similar "toFloat()" function to all the places that need it to meet the value tests

I didn't want to do the first option since it's modifying a very high-profile file; and it's overkill for this use-case.
I didn't want to do the second option since it'd end up creating a lot of duplicated casting code as more plugins use the GridOptions.

test/os/ui/draw/draw.test.js
```diff
@@ -0,0 +1,67 @@
+goog.require('ol.Feature');
+goog.require('ol.geom.Polygon');
+goog.require('ol.proj');
+goog.require('os.map');
+goog.require('os.ui.draw');
+
+describe('os.ui.draw', function() {
+  it('should read grid settings as numbers', function() {
```
wallw-bits: This test is superfluous. You've typed `osx.ui.draw.GridOptions` already, the compiler will handle it.

src/os/ui/draw/draw.js
wallw-bits: If I were using this from a plugin, I would do this (since you added an extern for the `GridOptions` type):

```
var options = /** @type {undefined|osx.GridOptions} */ (os.settings.get('whatever'));
if (options) {
  os.ui.draw.getGridFromFeature(feature, options);
}
```

The test using `getGridSetting` is unnecessary.

src/os/ui/draw/draw.js
```diff
@@ -1,8 +1,143 @@
...
+ *
+ * @param {ol.Feature} feature The source feature to trace.
+ * @param {osx.ui.draw.GridOptions} options Config object for grid generation
+ * @return {Array.<ol.Feature>}
+ *
+ * @suppress {accessControls} To allow direct access to feature metadata.
+ */
+os.ui.draw.getGridFromFeature = function(feature, options) {
+  var valid = function(options) {
+    var b = (typeof options != 'undefined' && options != null);
```
wallw-bits: You did not leave `options` as optional in your param list, nor did you specify that it could be `null` or `undefined`, so this is unnecessary.

A better check for neither null or undefined is just `thing != null` (using type coercion). If you want to check specifically for one or the other, use strict equality: `thing === undefined` or `thing === null`. `typeof` should be avoided where possible.

src/os/ui/draw/draw.js
```diff
@@ -1,8 +1,143 @@
...
+      x: (extent[0] < extent[2]) ? extent[2] : extent[0]
+    },
+    lat: {
+      n: (extent[1] < extent[3]) ? extent[1] : extent[3],
+      x: (extent[1] < extent[3]) ? extent[3] : extent[1]
+    }
+  }; // sort the data so the loops later are faster
+
+  var prop = {};
+  goog.object.extend(prop, feature.values_);
```
wallw-bits: To future-proof this, you might want to use `var prop = ol.obj.assign({}, feature.values_);`

src/os/ui/draw/draw.js
glennkitchellcaci: That's what I ended up doing -- should be on the same page now.

test/os/ui/draw/draw.test.js
glennkitchellcaci: Cool cool -- removed.

src/os/ui/draw/draw.js
glennkitchellcaci: I wanted the projection, altitude mode (clamp to ground, etc) and - more importantly - any settings that I didn't know about.

src/os/ui/draw/draw.js
```diff
@@ -1,8 +1,120 @@
...
+ * @param {Object} prop The drawing and altitude modes; copied from the context feature
+ * @param {osx.ui.draw.GridOptions} options The color, line thickness, etc settings for this grid
+ * @return {ol.Feature}
+ * @private
+ */
+os.ui.draw.gridFeatureFromExtent_ = function(extent, prop, options) {
+  // new() is faster than doing os.feature.copyFeature(feature) from the original feature
+  var g = ol.geom.Polygon.fromExtent(extent);
+  var prop_ = {'geometry': g};
+  goog.object.extend(prop_, prop);
```
wallw-bits: This is a little hacky and I would just use `f.setGeometry()` after creating it

PR #772, opened 2019-10-29
test(cypress): adjusts projection handling

justin-bits:
- default urls on default projection (3857) were incorrect; fixed
- default urls now provided for both 3857 and 4326
- `zoomOffset` previously incorrectly always set for 3857; fixed, now handles 3857 and 4326

### DIFF
--- a/cypress/support/execute-tests.sh
+++ b/cypress/support/execute-tests.sh
@@ -161,17 +161,17 @@ function overrideSettings() {
   cp -r $SETTINGS_SOURCE $SETTINGS_TARGET
 
   echo 'INFO: writing projection to settings file'
-  sed -i.bak 's@CYPRESS_PROJECTION@'$CYPRESS_PROJECTION'@g' $OPENSPHERE_CONFIG_TESTER && rm $OPENSPHERE_CONFIG_TESTER.bak
+  sed -i 's@CYPRESS_PROJECTION@'$CYPRESS_PROJECTION'@g' $OPENSPHERE_CONFIG_TESTER
 
   echo 'INFO: writing map urls to settings file'
-  sed -i.bak 's@STREET_MAP_URL@'$STREET_MAP_URL'@g' $OPENSPHERE_CONFIG_TESTER && rm $OPENSPHERE_CONFIG_TESTER.bak
-  sed -i.bak 's@WORLD_IMAGERY_URL@'$WORLD_IMAGERY_URL'@g' $OPENSPHERE_CONFIG_TESTER && rm $OPENSPHERE_CONFIG_TESTER.bak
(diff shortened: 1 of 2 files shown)


cypress/support/execute-tests.sh
```diff
@@ -159,6 +167,13 @@ function overrideSettings() {
   sed -i.bak 's@STREET_MAP_URL@'$STREET_MAP_URL'@g' $OPENSPHERE_CONFIG_TESTER && rm $OPENSPHERE_CONFIG_TESTER.bak
   sed -i.bak 's@WORLD_IMAGERY_URL@'$WORLD_IMAGERY_URL'@g' $OPENSPHERE_CONFIG_TESTER && rm $OPENSPHERE_CONFIG_TESTER.bak
 
+  echo 'INFO: zoom offset to settings file'
+  if [ "$CYPRESS_PROJECTION" = 3857 ]; then
+    sed -i.bak 's@"ZOOM_OFFSET"@'0'@g' $OPENSPHERE_CONFIG_TESTER && rm $OPENSPHERE_CONFIG_TESTER.bak
```
wallw-bits: Why add the `-i.bak` option if you are just going to subsequently delete the `.bak` file in the next statement?

justin-bits: I don't believe there was a good reason for that. Removed.

cypress/support/execute-tests.sh
```diff
@@ -153,11 +161,18 @@ function overrideSettings() {
   cp -r $SETTINGS_SOURCE $SETTINGS_TARGET
 
   echo 'INFO: writing projection to settings file'
-  sed -i.bak 's@CYPRESS_PROJECTION@'$CYPRESS_PROJECTION'@g' $OPENSPHERE_CONFIG_TESTER && rm $OPENSPHERE_CONFIG_TESTER.bak
```
schmidtk: The `-i.bak` flag was for cross-platform support, so we should bring that back for now. Details here:
https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux

cypress/support/execute-tests.sh
schmidtk: The `-i.bak` flag was for cross-platform support, so we should bring that back for now. Details here:
https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux

wallw-bits: Kevin thought it was related to https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux

I have the GNU coreutils installed on mine so I don't see the difference, but out of the box, removing that option may not work.

justin-bits: Commit reverted

PR #776, opened 2019-10-30
feat(progress): added utility for larger progress bars

gregroyal:


scss/os/_utils.scss
```diff
@@ -590,6 +594,11 @@ $somesizes: map-merge((
   top: -1.3rem;
 }
 
+.u-progress__large {
```
schmidtk: Should we follow Bootstrap's convention for utilities like these and use `.u-progress-lg`?

scss/os/_utils.scss
```diff
@@ -590,6 +594,11 @@ $somesizes: map-merge((
...
+  font-size: 1rem;
```
schmidtk: I think it would be best if these values were multipliers on the original variables, so `$progress-height * <multiplier>` and `$progress-font-size * <multiplier>`. That would ensure the progress bar is still larger if one were to change those variables.

Issue #574, opened 2019-05-29
Clicking Clear on application log closes it (Firefox)


welchyd:
Clicking CLEAR on application log closes it only in Firefox.
This seems to work fine in Chrome.
Doesn't happen in all versions of Firefox either, but I can get it consistently in FF 60.6.1esr (64-bit) on Mac and Linux

welchyd:
Fixed via #600

PR #779, opened 2019-10-30
fix(logger): the logger in firefox no longer randomly opens GEOVIZ-1

arwatts16:
This bug was caused by the fix to another one: [issue 574](https://github.com/ngageoint/opensphere/issues/574)

For testing this ticket, make sure to open the branch in firefox. Then check and make sure that:
• Log window doesn't reopen on its own when closed via the X (current bug).
• Log window doesn't close in Firefox when you click Clear (bug from 574).
• Log window closes when the application is closed (additional bug prior to 574).

### DIFF
--- a/src/os/debug/fancierwindow.js
+++ b/src/os/debug/fancierwindow.js
@@ -42,34 +42,22 @@ os.debug.FancierWindow.STYLE_RULES_ = goog.string.Const.from(
  * @inheritDoc
  */
 os.debug.FancierWindow.prototype.writeInitialDocument = function() {
-  goog.events.unlisten(this.win, goog.events.EventType.BEFOREUNLOAD, this.onBeforeUnload_, false, this);
+  goog.events.unlisten(this.win, goog.events.EventType.BEFOREUNLOAD, this.closeLogger, false, this);
   os.debug.FancierWindow.superClass_.writeInitialDocument.call(this);
   // close button should set the debug window to disabled
   if (this.win) {
-    goog.events.listenOnce(this.win, goog.events.EventType.BEFOREUNLOAD, this.onBeforeUnload_, false, this);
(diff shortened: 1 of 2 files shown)


src/os/debug/fancierwindow.js
```diff
@@ -62,6 +63,14 @@ os.debug.FancierWindow.prototype.onBeforeUnload_ = function() {
...
 
 
+/**
+ * @public
+ */
+os.debug.FancierWindow.prototype.closeLogger = function() {
+  // this.exit_ is private (grr), so copy it here
+  this.onBeforeUnload_();
+};
+
```
wallw-bits: `onBeforeUnload_` is defined in `os.debug.FancierWindow` itself. Could just rename it to `closeLogger` and make it public instead of wrapping it if it is all the same. Also, is that comment copy pasta?

schmidtk: I agree with renaming `onBeforeUnload_ -> closeLogger`.

As a cleanup step, we could call `this.exit_(null)` (it doesn't use the event anyway) and `@suppress {accessControls}` instead of copying the function contents. This may have been written before we started making use of `@suppress`.

src/os/debug/fancierwindow.js
```diff
@@ -62,6 +63,14 @@ os.debug.FancierWindow.prototype.onBeforeUnload_ = function() {
 };
 
 
+/**
+ * @public
```
schmidtk: `@public` is the default visibility, so you don't need to include it.

https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#public

src/os/debug/fancierwindow.js
arwatts16: Yeah, I that comment was copied and pasted, I think I was doing something else and then forgot to delete it. I can get rid of it though

PR #784, opened 2019-11-06
feat(list): add ability to pass dynamic attributes to list directive

rothm-caci:
List directive was modified to allow attributes to be passed from a parent scope to a directive with isolated scope compiled in the list directive using list-attr-* convention to map attributes, and with support for two-way binding.

template:
`<div list list-id="my-list" list-attr-my-item="myItem"></div>`

plugin:
`os.ui.list.add('my-list', '<my-directive item="myItem"></my-directive>');`

### DIFF
--- a/src/os/ui/list.js
+++ b/src/os/ui/list.js
@@ -50,7 +50,7 @@ os.ui.list.dispatcher_ = new goog.events.EventTarget();
  * @param {number=} opt_priority The sort priority (lowest to highest)
  */
 os.ui.list.add = function(id, markup, opt_priority) {
-  var map = os.ui.list.map_;
+  const map = os.ui.list.map_;
 
   if (!(id in map)) {
     map[id] = [];
@@ -71,9 +71,9 @@ os.ui.list.add = function(id, markup, opt_priority) {
  * @param {string} markup The directive or markup to delete
  */
 os.ui.list.remove = function(id, markup) {
-  var map = os.ui.list.map_[id];
+  const map = os.ui.list.map_[id];
   if (map) {
-    var item = ol.array.find(map, function(item) {
+    const item = ol.array.find(map, function(item) {
       return item.markup == markup;
     });
     if (item) {
@@ -96,7 +96,7 @@ os.ui.list.remove = function(id, markup) {
  * @param {string} id The list ID to remove
  */
 os.ui.list.removeList = function(id) {
-  var map = os.ui.list.map_[id];
+  const map = os.ui.list.map_[id];
   if (map) {
     map.forEach(function(item) {
       if (item) {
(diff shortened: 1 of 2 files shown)


src/os/ui/list.js
```diff
@@ -328,25 +350,33 @@ os.ui.ListCtrl.prototype.onChange = function(evt) {
...
-  var suffix = this.suffix || '';
+  const prefix = this.prefix || '';
+  const suffix = this.suffix || '';
 
   if (items) {
-    for (var i = 0, n = items.length; i < n; i++) {
-      var item = items[i];
+    let i = 0;
+    const n = items.length;
+    for (; i < n; i++) {
```
gregroyal: why not define i on this line?

src/os/ui/list.js
```diff
@@ -209,17 +210,32 @@ os.ui.Module.directive('list', [os.ui.listDirective]);
...
+      .reduce((acc, key) => {
+        let camel = key.replace(/^listAttr/, '');
+        camel = camel.charAt(0).toLowerCase() + camel.substr(1);
+        ctrl.attrs[camel] = scope.$eval(attr[key]);
+        scope.$watch(function() {
+          const kebab = key.replace(/([A-Z])/g, '-$1').toLowerCase();
+          return scope.$eval(element.attr(kebab));
+        }, function(newValue, oldValue) {
+          ctrl.attrs[camel] = newValue;
+          ctrl.update_();
```
gregroyal: this looks like it re-compiles the directive anytime any of the parameters change

rothm-caci: It now only compiles when !item.element, which is the first time it sees it.

src/os/ui/list.js
rothm-caci: Quite honestly, my IDE made the recommendation and I accepted it. I benchmarked it, and it seems to be marginally faster but significantly uglier.

src/os/ui/list.js
```diff
@@ -328,25 +350,32 @@ os.ui.ListCtrl.prototype.onChange = function(evt) {
...
 
   if (items) {
-    for (var i = 0, n = items.length; i < n; i++) {
-      var item = items[i];
+    const n = items.length;
+    for (let i = 0; i < n; i++) {
+      const item = items[i];
+
+      if (item.scope && Object.keys(this.attrs).length) {
+        goog.object.extend(item.scope, this.attrs);
```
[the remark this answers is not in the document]
schmidtk: We have been using `ol.obj.assign` which has its own polyfill if the native function doesn't exist, but [js-polyfills](https://github.com/inexorabletash/polyfill/blob/cefa69ee683474489685b126f3a74109735d8c76/es6.md#fundamental-objects) does provide it for us so we could switch to always using the native function.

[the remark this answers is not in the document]
rothm-caci: Removed conformance requirement to in order to allow Object.assign in PR

PR #786, opened 2019-11-06
fix(feature actions): using import button on filter actions works

arwatts16:
Before when you would use import from feature actions, it wouldn't find any layers to re import.

### DIFF
--- a/src/os/im/action/importaction.js
+++ b/src/os/im/action/importaction.js
@@ -55,3 +55,24 @@ os.im.action.sortByLabel = function(a, b) {
   var bLabel = b ? b.getLabel() : '';
   return goog.array.defaultCompare(aLabel, bLabel);
 };
+
+
+/**
+ * Get columns from a filterable.
+ *
+ * @param {os.filter.IFilterable} filterable
+ * @return {Array<os.ogc.FeatureTypeColumn>} columns of the  filterable
+ */
+os.im.action.getColumnsFromFilterable = function(filterable) {
+  var columns = null;
+
+  if (filterable instanceof os.layer.Vector) {
+    var source = /** @type {os.source.ISource} */ (filterable.getSource());
+    columns = os.source.getFilterColumns(source, true, true);
+    columns = columns.map(os.source.definitionsToFeatureTypes);
+  } else {
+    columns = filterable.getFilterColumns();
+  }
+
+  return columns;
+};
--- a/src/os/ui/filter/im/filterimporter.js
+++ b/src/os/ui/filter/im/filterimporter.js
@@ -92,6 +92,15 @@ os.ui.filter.im.FilterImporter.prototype.onParsingComplete = function(opt_event)
 
 
 /**
+ * @param {os.filter.IFilterable} filterable
+ * @return {Array<os.ogc.FeatureTypeColumn>}
+ */
+os.ui.filter.im.FilterImporter.prototype.getFilterColumnsFromFilterable = function(filterable) {
+  return filterable ? filterable.getFilterColumns() : null;
+};
+
+
+/**
  * Process the parsed filters.
  * @protected
  */
@@ -115,7 +124,7 @@ os.ui.filter.im.FilterImporter.prototype.processData = function() {
     if (this.layerId) {
       // if we have a layer ID, we were passed some context from a filter window, so use it
       var impliedFilterable = os.ui.filter.getFilterableByType(this.layerId);
-      var columns = impliedFilterable ? impliedFilterable.getFilterColumns() : null;
+      var columns = this.getFilterColumnsFromFilterable(impliedFilterable);
 
       if (impliedFilterable && columns && (this.ignoreColumns || filter.matches(columns))) {
         // this filter matches the columns of the passed in context, so add it as such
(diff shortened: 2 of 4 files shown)


src/os/ui/im/action/filteractionsui.js
```diff
@@ -283,8 +283,6 @@ os.ui.im.action.FilterActionsCtrl.prototype.launchExport = function() {
 os.ui.im.action.FilterActionsCtrl.prototype.launchImport = function() {
   os.metrics.Metrics.getInstance().updateMetric(os.im.action.Metrics.IMPORT, 1);
 
-  var event = new os.ui.im.ImportEvent(os.ui.im.ImportEventType.FILE, undefined, undefined, {
-    'layerId': this.entryType
-  });
+  var event = new os.ui.im.ImportEvent(os.ui.im.ImportEventType.FILE);
```
jsalankey: We don't want to remove the layer ID context here. The layer ID provided here is passed to the filter import directives via this line: https://github.com/ngageoint/opensphere/blob/master/src/os/ui/filter/im/filterimportui.js#L58

The FilterImportCtrl (or in this case, FilterActionImportCtrl) uses that context to configure the UI a little differently. That's where the bug is occurring.

src/os/ui/filter/im/filterimporter.js
```diff
@@ -116,6 +116,11 @@ os.ui.filter.im.FilterImporter.prototype.processData = function() {
       // if we have a layer ID, we were passed some context from a filter window, so use it
       var impliedFilterable = os.ui.filter.getFilterableByType(this.layerId);
       var columns = impliedFilterable ? impliedFilterable.getFilterColumns() : null;
+      var layer = os.map.mapContainer.getLayer(this.layerId);
+
+      if (layer instanceof os.layer.Vector && columns == null) {
+        columns = os.im.action.getVectorLayerColumns(layer);
```
schmidtk: I don't think we want this for generic filters, because filters are applied to server layers and should only match server-provided columns. This is what `getFilterColumns` will provide, where supported. We do not want to match layers that do not support server-side filtering, like file layers.

Filter actions operate on local columns, so in that specific case it makes sense to try getting columns from the layer. That specific case should be handled in `os.ui.im.action.FilterActionImporter`.

src/os/ui/filter/im/filterimporter.js
```diff
@@ -91,6 +91,15 @@ os.ui.filter.im.FilterImporter.prototype.onParsingComplete = function(opt_event)
 };
 
 
+/**
+ * @param {os.filter.IFilterable} filterable
+ * @return {Array}
```
schmidtk: The return type should match `getFilterColumns`, so `Array<os.ogc.FeatureTypeColumn>`.

src/os/ui/im/action/filteractionimporter.js
```diff
@@ -64,6 +64,17 @@ os.ui.im.action.FilterActionImporter.prototype.getFilterTooltip = function(entry
 };
 
 
+/**
+ * @param {os.layer.Vector} vectorLayer
+ * @return {Array}
+ * @override
+ * @suppress {checkTypes}
```
wallw-bits: Using this suppression is typically a big red flag. Let's figure out the proper typing to make the compiler happy instead.

src/os/im/action/importaction.js
```diff
@@ -55,3 +55,20 @@ os.im.action.sortByLabel = function(a, b) {
...
 };
+
+
+/**
+ * Get columns from a vector layer.
+ *
+ * @param {os.layer.Vector} layer
+ * @return {Array} columns of the vector layer
+ */
+os.im.action.getVectorLayerColumns = function(layer) {
```
schmidtk: I would suggest extracting this block and adapting it for use in this new function.
https://github.com/ngageoint/opensphere/blob/master/src/os/ui/im/action/filteractionimport.js#L121-L129

Your function definition will change a bit, as the input param should be a `os.filter.IFilterable` and it should return `Array<os.ogc.FeatureTypeColumn>`. I would also suggest renaming to reflect the new behavior, possibly `os.im.action.getColumnsFromFilterable`.

Once that's done, the function can be used in both `os.ui.im.action.FilterActionImportCtrl#onLayerChange` and `os.ui.im.action.FilterActionImporter#getFilterColumnsFromFilterable`.

src/os/ui/im/action/filteractionimporter.js
schmidtk: Given the previous suggestion to update your convenience function, this should be able to just use `@inheritDoc` with no additional documentation. We like to avoid suppressing `checkTypes` because it's typically avoiding legitimate compiler errors.

src/os/im/action/importaction.js
```diff
@@ -55,3 +55,24 @@ os.im.action.sortByLabel = function(a, b) {
...
+ */
+os.im.action.getColumnsFromFilterable = function(filterable) {
+  var columns = null;
+
+  if (filterable instanceof os.layer.Vector) {
+    var source = /** @type {os.source.ISource} */ (filterable.getSource());
+    columns = os.source.getFilterColumns(source, true, true);
+    columns = columns.map(os.source.definitionsToFeatureTypes);
+  } else {
+    this.columns = filterable.getFilterColumns();
```
schmidtk: Should be `columns` instead of `this.columns`.

PR #787, opened 2019-11-06
feat(vector): add convenience functions to vectorsource

csgibson:


src/os/source/vectorsource.js
```diff
@@ -3217,6 +3234,29 @@ os.source.Vector.prototype.invertSelection = function() {
...
 
 
+/**
+ * Convenience method select by id/array of ids
+ * @param {number|Array<number>} ids
+ * @param {boolean=} opt_deselect
+ * @suppress {accessControls}
+ */
+os.source.Vector.prototype.selectById = function(ids, opt_deselect) {
+  if (ids && !Array.isArray(ids)) {
```
schmidtk: Instead of duplicating logic here and again in `hideById`, we may want another convenience function like `idsToFeatures`.

csgibson:
Seemed like this should have test coverage so I added those

PR #788, opened 2019-11-08
style(utils): added table header sort styling

gregroyal:


scss/os/_utils.scss
```diff
@@ -222,6 +222,10 @@ $user-select-options: all, auto, text, none;
   table-layout: fixed;
 }
 
+.u-table__sortable-header {
```
schmidtk: `.u-cursor-pointer`? We already have several similar classes (grab, default, help, wait, not-allowed).

gregroyal: we might have enough code examples now to avoid people creating font buttons. That was the main driver to avoid it. At this point i could be sway'd to add the utility. But we need to still be careful where we use it

schmidtk: Abuse of any utility class is something that should be dissuaded during the review process, if it happens. Given this class is only changing the cursor to a pointer, it seems reasonable to name it as such.

PR #790, opened 2019-11-10
fix(net): fix double encoding issue

wallw-bits:
The `uri.setQuery()` method takes an argument for whether or not to decode that was not being used. Because of this fix, `BaseServerModifier` is no longer necessary (although the API for configuring it has been left the same to avoid breaking its usage).

BREAKING CHANGE: `UrlModifier` now avoids doubly encoding the query string. Any steps taken on the receiving end of these requests to code around that bug need to be reviewed and possibly removed.

BREAKING CHANGE: `os.net.BaseServerExpression` was removed because it was not fully featured for parsing out domain names. Please use `new goog.Uri(str).getDomain()` or `new URL(str).hostname` instead.

### DIFF
--- a/src/os/net/baseservermodifier.js
+++ b/src/os/net/baseservermodifier.js
@@ -1,68 +1,15 @@
-goog.provide('os.net.BaseServerExpression');
 goog.provide('os.net.BaseServerModifier');
 goog.require('goog.Uri');
 goog.require('os.net.URLModifier');
-
-
-
-/**
- * @extends {os.net.URLModifier}
- * @constructor
- */
-os.net.BaseServerModifier = function() {
-  os.net.BaseServerModifier.base(this, 'constructor', 'baseServer');
-};
-goog.inherits(os.net.BaseServerModifier, os.net.URLModifier);
-
-
-/**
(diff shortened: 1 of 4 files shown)


src/os/net/baseservermodifier.js
```diff
@@ -1,68 +1,15 @@
...
 os.net.BaseServerModifier.configure = function(server) {
-  os.net.BaseServerModifier.replace_.length = 0;
-
   if (server) {
-    os.net.BaseServerModifier.replace_.push({
-      search: new RegExp(os.settings.get('baseServerRegex', '^\\/(.*)$')),
-      replace: server.replace(/\/$/, '') + '/$1'
-    });
+    var config = {};
+    config[os.settings.get('baseServerRegex', '^\\/([^\\/].*)$')] = server.replace(/\/$/, '') + '/$1';
```
wallw-bits: The regex change here fixes scheme-relative URLs (e.g. `//some.domain.com/...`)

gregroyal: Make sure all stakeholders and downstream projects work with this change

PR #791, opened 2019-11-12
feat(slickgrid): add new events and event handling for editing

kwilson-bits:


src/os/ui/slick/slickgrid.js
```diff
@@ -1259,6 +1269,47 @@ os.ui.slick.SlickGridCtrl.prototype.apply = function() {
...
+
+
+/**
+ * The cell is no longer being edited.
+ * This is triggered when an edit has been made or discarded.
+ *
+ * @protected
+ */
+os.ui.slick.SlickGridCtrl.prototype.cellEditorDestroyed = function() {
+  this.scope.$emit(os.ui.slick.SlickGridEvent.CELL_EDITOR_DESTROYED, arguments);
```
welchyd: Not sure how the compiler missed it, but `arguments` isn't defined

src/os/ui/slick/slickgrid.js
```diff
@@ -247,6 +245,18 @@ os.ui.slick.SlickGridCtrl = function($scope, $element, $compile) {
     this.dataView.syncGridSelection(this.grid, false);
   }
 
+  if (options['editable']) {
+    // editable grids have additional CSS styles. add the class to enable them.
+    this.element.addClass('editable');
```
welchyd: Is this a slick-grid specific class? We usually try to be more specific with css class names when deviating from using a bootstrap style. For instance `c-slickgrid__editable` indicates that the class is for the slick grid component. It helps to be specific since `editable` could apply to a number of things

src/os/ui/slick/slickgrid.js
wallw-bits: `arguments` is available on any function that is not an arrow function. See [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) for details.

It does, however, skip typing the function parameters in the compiler, which is not so great. This is completely valid JS, but the typing for the closure compiler could be better.

src/os/ui/slick/slickgrid.js
kwilson-bits: This logic was already in the file before my changes; I only moved it down a few lines rather than create a second `if (options['editable'])` block. Changing class names will have a bigger impact elsewhere and is not part of my scope.

src/os/ui/slick/slickgrid.js
jsalankey: The other Slickgrid listeners we define in here all just have a vaguely defined `args` object that gets passed to them from the Slickgrid notify calls. It's only marginally better, but these should follow that pattern if possible.

kwilson-bits: I pushed a few changes to improve how the listener functions are defined.

src/os/ui/slick/slickgrid.js
welchyd: Good point. Thanks for clarifying!

PR #802, opened 2019-11-25
feat(mixin): add double-transform protection

wallw-bits:
Adds double-transform protection to `ol.geom.Geometry` by caching the destination projection code on the geometry and checking it when subsequently transforming again.

### DIFF
--- a/src/os/geom/geometryfield.js
+++ b/src/os/geom/geometryfield.js
@@ -7,5 +7,6 @@ goog.provide('os.geom.GeometryField');
  */
 os.geom.GeometryField = {
   DIRTY: '_dirty',
-  NORMALIZED: '_normalized'
+  NORMALIZED: '_normalized',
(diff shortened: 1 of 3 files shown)


schmidtk:
I'm a bit concerned about the performance here. With a point geometry, the transform mixin takes over twice as long as the original due to the multiple extra `ol.proj.get` calls.

wallw-bits:
"Twice as long" is still measured in single microseconds:

Master:
```
Transform: Average time per call: 0.0010448499999911292 ms
```
Current Branch:
```
Transform: Average time per call: 0.001634230000010575 ms
Protected: Average time per call: 0.0004169899999978952 ms
```
Nevertheless, I improved it somewhat locally:
```
Transform: Average time per call: 0.001195480000009411 ms
Protected: Average time per call: 0.0007529400000057649 ms
```

PR #807, opened 2019-11-28
Investigate refactor to goog.module

schmidtk:
This branch is being used to work out any kinks with `goog.module` and determine the preferred approach to converting source files. The intent is to require minimal changes when converting from `goog.module` to ES6 modules, once all code has made the initial transition.

The first commit is required for tests to run with `goog.module`, so it can be separately merged if needed.

These PR's are required for the build to pass. They add `goog.module` handling to our ESLint config.
- https://github.com/ngageoint/eslint-config-opensphere/pull/10
- https://github.com/ngageoint/eslint-plugin-opensphere/pull/6

### DIFF
--- a/src/os/alert/alert.js
+++ b/src/os/alert/alert.js
@@ -1,41 +1,48 @@
-goog.provide('os.alert.Alert');
-goog.require('os.alert.AlertEventSeverity');
+goog.module('os.alert.Alert');
+goog.module.declareLegacyNamespace();
 
+const AlertEventSeverity = goog.require('os.alert.AlertEventSeverity'); // eslint-disable-line no-unused-vars
 
 
 /**
- * Object to carry around alert details
- *
- * @param {string} message The alert to send and add to the window
- * @param {os.alert.AlertEventSeverity=} opt_severity Severity of the event, defaults to error
- * @constructor
+ * Object to carry around alert details.
  */
-os.alert.Alert = function(message, opt_severity) {
+class Alert {
   /**
-   * @type {string}
-   * @private
+   * @param {string} message The alert to send and add to the window
+   * @param {AlertEventSeverity=} opt_severity Severity of the event, defaults to error
    */
-  this.message_ = message;
+  constructor(message, opt_severity) {
+    /**
+     * The alert message.
+     * @type {string}
+     * @private
+     */
+    this.message_ = message;
+
+    /**
+     * The alert severity.
+     * @type {AlertEventSeverity|undefined}
+     * @private
+     */
+    this.severity_ = opt_severity;
+  }
 
   /**
-   * @type {os.alert.AlertEventSeverity|undefined}
-   * @private
+   * Get the alert message.
+   * @return {string}
    */
-  this.severity_ = opt_severity;
-};
+  getMessage() {
+    return this.message_;
+  }
 
+  /**
+   * Get the alert severity.
+   * @return {AlertEventSeverity|undefined}
+   */
+  getSeverity() {
+    return this.severity_;
+  }
+}
 
-/**
- * @return {string}
- */
-os.alert.Alert.prototype.getMessage = function() {
-  return this.message_;
-};
-
-
-/**
- * @return {os.alert.AlertEventSeverity|undefined}
- */
-os.alert.Alert.prototype.getSeverity = function() {
-  return this.severity_;
-};
+exports = Alert;
--- a/src/os/alert/alerteventlevel.js
+++ b/src/os/alert/alerteventlevel.js
@@ -1,0 +1,36 @@
+goog.module('os.alert.AlertEventLevel');
+goog.module.declareLegacyNamespace();
+
+
+/**
+ * An alert level.
+ */
+class AlertEventLevel {
+  /**
+   * @param {string} name The name of the level.
+   * @param {number} value The numeric value of the level.
+   */
+  constructor(name, value) {
+    /**
+     * The name of the level.
+     * @type {string}
+     */
+    this.name = name;
+
+    /**
+     * The numeric value of the level.
+     * @type {number}
+     */
+    this.value = value;
+  }
+
+  /**
+   * @return {string} String representation of the alert level.
+   * @override
+   */
+  toString() {
+    return this.name;
+  }
+}
+
+exports = AlertEventLevel;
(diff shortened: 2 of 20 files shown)


schmidtk:
Lessons learned so far:
- Classes are `@struct` by default, which prevents using bracket notation or adding properties outside the constructor. If either of those are needed (ie, using bracket notation to avoid property renaming), the class must have `@unrestricted` in the JSDoc.
- `@typedef` declarations are only for the compiler, but must be exported if they're used outside the file that declares them.

schmidtk:
Google's docs on `goog.module` and ES6:
- [`goog.module`](https://github.com/google/closure-library/wiki/goog.module:-an-ES6-module-like-alternative-to-goog.provide)
- [Migrating from `goog.module` to ES6 modules](https://github.com/google/closure-compiler/wiki/Migrating-from-goog.modules-to-ES6-modules)
- [Closure/ES6 interop](https://github.com/google/closure-library/wiki/ES6-modules-and-Closure-Library)
- [ES6 and the Closure Library](https://github.com/google/closure-library/wiki/ES6-modules-and-Closure-Library)

src/os/alert/alertmanagerinstance.js
```diff
@@ -0,0 +1,12 @@
+goog.module('os.alertManager');
+goog.module.declareLegacyNamespace();
+
+goog.require('os.alert.AlertManager');
+
+/**
+ * Global alert manager reference.
+ * @type {os.alert.AlertManager}
+ */
+const instance = os.alert.AlertManager.getInstance();
```
schmidtk: Ideally I'd like to move this into `alertmanager.js` as:
```
const instance = new os.alert.AlertManager();
Object.freeze(instance);
```
Then add our own `os.alert.AlertManager#getInstance` function that returns the instance.

That would replace the `goog.addSingletonGetter` call, but would be a breaking change. My goal in this PR is to illustrate how we can translate files to modules without breaking changes.

src/os/alert/alert.js
```diff
@@ -1,41 +1,48 @@
-goog.provide('os.alert.Alert');
-goog.require('os.alert.AlertEventSeverity');
+goog.module('os.alert.Alert');
+goog.module.declareLegacyNamespace();
 
+const AlertEventSeverity = goog.require('os.alert.AlertEventSeverity'); // eslint-disable-line no-unused-vars
```
jsalankey: Is this require needed if this file only uses it as a type declaration, but never actually references? Seems annoying to have to require it like this and work around the linter.

src/os/alert/alerteventlevel.js
```diff
@@ -0,0 +1,36 @@
+goog.module('os.alert.AlertEventLevel');
+goog.module.declareLegacyNamespace();
```
jsalankey: Is calling this function in every class needed? What does it actually do? It doesn't have any arguments... so how does it know the context of the declaration we're making?

src/os/alert/alertmanager.js
```diff
@@ -1,163 +1,161 @@
...
   /**
-   * Tracks the number of alerts process by any given client so the client can get all alerts even if it comes online
-   * later.
-   * @type {!Object.<!string, !number>}
-   * @private
+   * Maximum number of alerts to be saved.
+   * @return {number}
    */
-  this.processedByClientCount_ = {};
+  static get MAX_SAVED() {
```
jsalankey: Do we have an example of declaring a setter?

src/os/alert/alert.js
schmidtk: Yes. The Closure compiler will not understand the type unless required in the file. It's possible it would be implicitly required like a lot of our code is doing now, but you would have to use `os.alert.AlertEventSeverity` instead and that will likely stop working once we switch to ES6.

Related: https://github.com/google/closure-compiler/issues/3041

src/os/alert/alertmanagerinstance.js
```diff
@@ -0,0 +1,12 @@
...
+
+const AlertManager = goog.require('os.alert.AlertManager');
+
+/**
+ * Global alert manager reference.
+ * @type {AlertManager}
+ */
+const instance = AlertManager.getInstance();
+
+exports = instance;
```
jsalankey: Is there no way to have multiple different exports from the root class file? Having to add an instance creator like this seems superfluous.

src/os/alert/alerteventlevel.js
schmidtk: Yep. This is what actually allows you to `goog.require` the module in `goog.provide` files. `goog.module` doesn't add anything in the global namespace (part of the point of modules), so `goog.module.declareLegacyNamespace` does that if called within a module.

You can dig into the details in Closure's code if you want, but basically they have a module loader that handles each of these files and calling this function during module load puts the namespace on the global window object.

src/os/alert/alertmanager.js
schmidtk: Yes: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set

Note this is using `static get` to create a constant property on the class itself (`os.alert.AlertManager.MAX_SAVED`). This seems to be the [recommended way](https://stackoverflow.com/a/32647583) to define a constant on an ES6 class.

src/os/alert/alertmanagerinstance.js
schmidtk: I sort of explained this in https://github.com/ngageoint/opensphere/pull/807#discussion_r352179076, but basically I didn't want breaking changes in this PR. There is a _much_ better way to define singletons, but since we have code doing `goog.require('os.alertManager')` I wanted to preserve that namespace and you can't declare multiple modules/namespaces per file.

You can export as many things as you want, for example this is perfect valid:
```
exports.SOME_CONSTANT = 42;
exports.SomeClass = class MyClass { ... };
exports.someFunction = function() { ... };
```

src/os/alert/alertmanagerinstance.js
schmidtk: Scratch the `Object.freeze` line. That prevents modification of any properties on the class.

src/os/alert/alertmanagerinstance.js
schmidtk: I made a change to replace `goog.addSingletonGetter` with a `static getInstance()` function on the class. The second instance file is still there for backward compatibility, with a note up top that we should drop it. I can do that refactor later, but I wanted to limit changes in this PR to converted files where possible.

PR #816, opened 2019-12-08
feat(build): Added html linter to build

gregroyal:


wallw-bits:
This is not a breaking change as it does not affect other plugin or app projects. Please update the commit message accordingly.

gregroyal:
> This is not a breaking change as it does not affect other plugin or app projects. Please update the commit message accordingly.

You're right! updated

PR #818, opened 2019-12-10
docs(modules): add documentation on goog.module and es6 classes

schmidtk:
Describes our transition process to ES6 and explains new code practices with `goog.module` and ES6 `class`.

Live link: http://elite-cap.surge.sh/guides/es6_transition_guide.html


wallw-bits:
Great start! Probably should have a note about how to properly mark classes extending using generics, e.g. `@extends B<X>`. Should also have something on tests with goog modules.

PR #823, opened 2019-12-17
fix(tuieditor): speed-sensitive loading bug

brockwayc:
tuieditor.js had an issue whereby the editor box would not load properly in certain situations. It appears to be speed-sensitive. This may be due to a change in Codemirror. Issue required a one-line fix.

### DIFF
--- a/src/os/ui/text/tuieditor.js
+++ b/src/os/ui/text/tuieditor.js
@@ -466,6 +466,13 @@ os.ui.text.TuiEditorCtrl.prototype.setTargetBlankPropertyInLinks = function() {
  * Codemirror has the same issues that it did in simpleMDE with initalization.
  * Codemirror issue #798. when its put in a textarea with display none it needs a refresh
  * After Codemirror is added, call refresh on it
+ *
+ * Another issue: this broke at some point. The line that reads codeMirror['0'].innerText != 'xxxxxxxxxx'
+ * had codeMirror.length instead. Turns out codeMirror.length is always defined no matter what.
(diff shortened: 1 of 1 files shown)


src/os/ui/text/tuieditor.js
```diff
@@ -466,14 +466,21 @@ os.ui.text.TuiEditorCtrl.prototype.setTargetBlankPropertyInLinks = function() {
...
+ *
  * Commented on tuieditor issue #191
  * @private
  */
 os.ui.text.TuiEditorCtrl.prototype.fixCodemirrorInit_ = function() {
   if (this.element) {
     this.cmFixAttempt_ = this.cmFixAttempt_ ? this.cmFixAttempt_ + 1 : 1;
     var codeMirror = this.element.find('.te-md-container .CodeMirror');
-    if (this['tuiEditor'] && codeMirror.length) {
+    if (this['tuiEditor'] && codeMirror['0'].innerText != 'xxxxxxxxxx') {
```
gregroyal: probably should check the length before using an item from the array. did you find this somewhere? can you document the issue number incase it gets fixed so we can remove it

swtyler: This does seem to fix the problem for me, I refreshed a watchbox edit form like 20 times on trunk and I saw the description not populate 3 or 4 times. On the branch with this change I never saw it happen...

We should still check for `codeMirror.length` like Greg said. This is the code adding in the xxxx https://github.com/codemirror/CodeMirror/blob/master/src/measurement/position_measurement.js#L607, Maybe there is a better way to check for that element?

PR #824, opened 2019-12-18
chore(button): updated outline button color to appear clickable

briedinger:
Make button text different than plain text.

"Ghost buttons" btn-outline-primary border-0 are often use to get an affordance of clicking, with hover effects and all, without having a full button style. This change updates the color for primary to remain blue, so it behaves like a clickable hyperlink and not plain text.

### DIFF
--- a/scss/overrides_slate/_bootswatch.scss
+++ b/scss/overrides_slate/_bootswatch.scss
@@ -28,7 +28,7 @@ $web-font-path: 'https://fonts.googleapis.com/css?family=Open+Sans:400,700' !def
 }
 
 .btn-outline-primary {
-  color: $white;
+  color: $link-color;
 }
 
 .form-control:disabled,


scss/overrides_slate/_bootswatch.scss
```diff
@@ -28,7 +28,7 @@ $web-font-path: 'https://fonts.googleapis.com/css?family=Open+Sans:400,700' !def
 }
 
 .btn-outline-primary {
-  color: $white;
+  color: $link-color;
```
gregroyal: just curious why you picked $link-color over $primary?

briedinger: Primary is a darker blue and was hard to see in dark mode - dark blue on dark gray. Link color was a better choice for a text-only clickable thing, and the animation went well with the actual primary button color on hover. I suppose this could go sideways if we choose link color to be not blue, for branding or whatever. I don't see that happening, though.

gregroyal: ok, just curious!

PR #825, opened 2019-12-30
refactor(punyparent): changes forEach to reduce for readability

rusticpenguin:
Old code was using a forEach that was adding to a variable. Changed it to a reduce to better illustrate the meaning of the code.

Also changed function() to arrow function for the higher order functions.

This should not affect functionality. It produces the same result

### DIFF
--- a/src/os/ui/util/punyparent.js
+++ b/src/os/ui/util/punyparent.js
@@ -97,27 +97,27 @@ os.ui.util.PunyParentCtrl.prototype.onResize_ = function() {
 os.ui.util.PunyParentCtrl.prototype.onThrottleResize_ = function() {
   if (this.element_) {
     var children = this.element_.children().toArray();
-    var childrenWidth = 0;
-    children.forEach(function(child) {
+    var childrenWidth = children.reduce((accumulator, child) => {
       var c = $(child);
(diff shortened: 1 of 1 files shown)


src/os/ui/util/punyparent.js
```diff
@@ -97,27 +97,26 @@ os.ui.util.PunyParentCtrl.prototype.onResize_ = function() {
 os.ui.util.PunyParentCtrl.prototype.onThrottleResize_ = function() {
   if (this.element_) {
     var children = this.element_.children().toArray();
-    var childrenWidth = 0;
-    children.forEach(function(child) {
+    var childrenWidth = children.reduce((accumulator, child) => {
       var c = $(child);
       // ignore the resize trigger since thats the parent size
       if (!c.hasClass('resize-triggers')) {
```
jsalankey: If this `if` case ever fails to pass, there is no `return accumulator` statement after, so the function is assumed to return `undefined`, which will result in `NaN` as the result rather than simply ignoring those elements. We should handle that case.

rusticpenguin: Good catch! I moved the return outside of the if

PR #826, opened 2020-01-06
Update required version of node

bradh:
Node 6 isn't enough.  The `@semantic-release/changelog` dependency wants at least 8.16.


bradh:
Ah, forgot that I need the magic format for the commit message.

PR #829, opened 2020-01-09
feat(contextmenu): allow middle click to launch context menu

csgibson:
Some users with browser configurations set by their IT departments get an annoying browser context menu on top of our app one. This gives them another, similar option to launch that menu. 
I'm open to other suggestions if there are better ideas for things we can fix on our end.

### DIFF
--- a/src/os/ol/events/condition.js
+++ b/src/os/ol/events/condition.js
@@ -14,3 +14,17 @@ os.ol.events.condition.rightClick = function(mapBrowserEvent) {
       mapBrowserEvent.type === ol.MapBrowserEventType.POINTERDOWN) &&
       !!mapBrowserEvent.pointerEvent && mapBrowserEvent.pointerEvent.button === 2;
 };
+
+
+/**
+ * If a map browser event is for a middle or right click.
+ *
+ * @param {ol.MapBrowserEvent} mapBrowserEvent The map browser event
+ * @return {boolean} If the event represents a right click
+ */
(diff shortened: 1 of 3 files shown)


schmidtk:
Are we certain the browser configuration is responsible, or are we not stopping propagation of the event in all cases?

csgibson:
User has no access to `about:config` to verify that it is the issue in firefox but said that it is a setting controlled by group policy script on their network. 
I tested it in Firefox with `dom.evet.contextmenu.enabled` turned to false (normally true). In the meantime I suggested they use double right click, they were just hitting the escape button before.

wallw-bits:
This change only addresses the map right click, so it is not nearly comprehensive enough. Secondly, middle-click already has a (quite useful) function within the browser.

### Solution
Since this is Firefox, they can always override system policy settings in the `user.js` file.

Have those users place the following line in `%APPDATA%\Mozilla\user.js` (which won't exist by default and they will have to create it):
```
user_pref("dom.event.contextmenu.enabled", true);
```
Notes:
* The user.js file is the last place from which settings are loaded, and that even toggling (or attempting to toggle) any setting in this file in about:config will no longer work
* There has been some [discussion on the Mozilla dev site](https://bugzilla.mozilla.org/show_bug.cgi?id=1543752#c44) about removing `user.js` since it is a security issue (other application installs can touch the file and lock the user out of settings, etc.) and also because it is holding up startup a bit. If this happens then we will need a different solution.

PR #835, opened 2020-01-16
fix(maxzoom): reduces maxzoom for streetmap-4326

justin-bits:
The maxzoom is set incorrectly for Street Maps in 4326.


wallw-bits:
https://services.arcgisonline.com/ArcGIS/rest/services/ESRI_StreetMap_World_2D/MapServer says 15. Are we sure we want 13?

justin-bits:
Closing this without merge.  The -1 offset makes the current value of 16 correct.  Some map tiles are available at 15 as indicated by the documentation.

PR #838, opened 2020-01-20
Places persistence fixes

jsalankey:
This adds checks for the sporadic null cases we've seen with saving/loading Places. It also adds another spot where it will loop back and re-initialize if there is an missing places root node.

I also noticed while working on this that the Spatial menu places were under a different menu header and that Quick Add Places was not an available option, so I changed that to be consistent with the Map menu.


schmidtk:
Thanks for adding the alert! The build passed, but status didn't update in GitHub. I restarted it, so hopefully it'll sync this time.

PR #839, opened 2020-01-21
test(tracks): add minor test case for track property changes

bradh:
Something I wrote while trying to understand how to set the track name and label. Not especially useful, but since I had it I thought I'd offer it.


wallw-bits:
Not that we don't love more tests, but isn't that more of a test of `ol.Object` than it is of tracks?

bradh:
Probably. Happy just to close it off if it isn't really providing anything.

PR #849, opened 2020-01-24
feat(bins): Added Empty Bin support

glennkitchellcaci:
Updated OS backend to support graph views of binned Feature data where there are numerical gaps in the data.

NOTE: Histogram and Bins unused by OpenSphere.  

**Unit Tests**
- time.test.js - Tests the use of a new "step" function, which gets the millisecond difference between a given date and the next Month or Year

- uniquebinmethod.test.js - Now includes "getStatsForBin" function to calculate range and step size based on a config.  Also stores the configuration showing/hiding empty bins after crossfiltering calculations
- numericbinmethod.test.js - Updating accordingly
- datebinmethod.test.js - Updating accordingly

See GEOVIZ-33

### DIFF
--- a/externs/os.externs.js
+++ b/externs/os.externs.js
@@ -542,3 +542,22 @@ osx.feature.RingDefinition;
  * }}
  */
 osx.feature.RingOptions;
+
+
+/**
+ * Namespace.
+ * @type {Object}
+ */
+osx.histo;
+
+
+/**
+ * @typedef {{
+  *   range: Array<number>,
+  *   step: number,
+  *   binCount: number,
+  *   binCountAll: number
+  * }}
+  */
+osx.histo.BinMethodStats;
+
--- a/externs/os.externs.js
+++ b/externs/os.externs.js
@@ -542,22 +542,3 @@ osx.feature.RingDefinition;
  * }}
  */
 osx.feature.RingOptions;
-
-
-/**
- * Namespace.
- * @type {Object}
- */
-osx.histo;
-
-
-/**
- * @typedef {{
-  *   range: Array<number>,
-  *   step: number,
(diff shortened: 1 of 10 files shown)


src/os/histo/uniquebinmethod.js
```diff
@@ -348,6 +379,25 @@ os.histo.UniqueBinMethod.prototype.getFilterForBin = function(bin) {
 };
 
 
+/**
+ * Get the range, step size, etc for the bins made by this method.
+ *
+ * @param {!Array<os.histo.Bin>} bins The bins made using this bin method
+ * @return {osx.histo.BinMethodStats|null} The config
+ * @public
```
jsalankey: `@public` tags shouldn't be needed.

externs/os.externs.js
```diff
@@ -542,3 +542,22 @@ osx.feature.RingDefinition;
...
+
+/**
+ * @typedef {{
+  *   range: Array<number>,
+  *   step: number,
+  *   binCount: number,
+  *   binCountAll: number
+  * }}
+  */
+osx.histo.BinMethodStats;
```
jsalankey: These shouldn't need to be externs if they aren't referenced by a directive.

wallw-bits:
@schmidtk @bradh Any idea why `commitlint-travis` passed commit message? I'd expect that subject to fail the lint rules.

bradh:
> @schmidtk @bradh Any idea why `commitlint-travis` passed this commit message? I'd expect that subject to fail the lint rules.

Which commit message? The PR title is upper case (which wouldn't be valid), but all of the individual commits look OK.

wallw-bits:
Doh! Disregard.

PR #850, opened 2020-01-24
fix(cesium): better goto support for primitive layers

wallw-bits:
Adds a general `getExtent` implementation for primitive layers.

This is not comprehensive because the application should really be operating on aggregated bounding volumes specified and implemented by the renderer (extents for OpenLayers and BoundingSpheres for Cesium) rather than always relying on the OL `getExtent()` function.

However, this will at least get the user in the correct vicinity even if it is not precise regarding altitude.

### Testing
1. Clone https://github.com/AnalyticalGraphicsInc/3d-tiles-samples into your workspace
1. Add `<yourWorkspaceUrl>/3d-tiles-samples/tilesets/TilesetWithDiscreteLOD/tileset.json` to OpenSphere
1. Right-click the new layer
    * Verify that "Go To" is an option
1. Select "Go To"
    * Verify that it goes to the big 3D dragon

### DIFF
--- a/src/plugin/cesium/primitivelayer.js
+++ b/src/plugin/cesium/primitivelayer.js
@@ -97,3 +97,26 @@ plugin.cesium.PrimitiveLayer.prototype.removePrimitive = function() {
 
   os.dispatcher.dispatchEvent(os.MapEvent.GL_REPAINT);
 };
+
+
+/**
+ * @inheritDoc
+ */
+plugin.cesium.PrimitiveLayer.prototype.getExtent = function() {
+  var primitive = this.getPrimitive();
+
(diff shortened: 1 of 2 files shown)


src/plugin/cesium/primitivelayer.js
```diff
@@ -97,3 +97,26 @@ plugin.cesium.PrimitiveLayer.prototype.removePrimitive = function() {
...
 };
+
+
+/**
+ * @inheritDoc
+ */
+plugin.cesium.PrimitiveLayer.prototype.getExtent = function() {
+  var primitive = this.getPrimitive();
+
+  if (primitive.ready && primitive.boundingSphere) {
```
schmidtk: This could use a null test.

PR #857, opened 2020-01-30
feat(request): add get method to return a request's modifiers

welchyd:


src/os/net/request.js
```diff
@@ -368,6 +368,16 @@ os.net.Request.prototype.removeModifier = function(modifier) {
 };
 
 
+/**
+ * Returns a copy of the modifiers
+ *
+ * @return {?Array.<os.net.IModifier>}
+ */
+os.net.Request.prototype.getModifiers = function() {
+  return goog.object.clone(this.modifiers_);
```
schmidtk: - `goog.object.clone` on an array is equivalent to `<array>.slice()`. Please prefer native calls over `goog` when available.
- The modifiers can be `null`, which will cause an error here. I would suggest returning `this.modifiers_ ? this.modifiers_.slice() : null`.

PR #860, opened 2020-02-03
build(package): create concurrent build scripts

csgibson:
This introduces a new library and makes a small amount of the build concurrent. 
Expected reduction in build times: ~15%.


wallw-bits:
I believe this is replaced by the webpack build work. @csgibson would you agree?

PR #868, opened 2020-02-10
feat: Color Selected

glennkitchellcaci:
Supporting code for "Color Selected" and "Color Selected Bins" features

### DIFF
--- a/src/os/data/histo/colormodel.js
+++ b/src/os/data/histo/colormodel.js
@@ -180,6 +180,37 @@ os.data.histo.ColorModel.prototype.getResults = function() {
  * Get the bin/color pairs currently applied by this color model.
  *
  * @return {Object<string, string>}
+ * @suppress {accessControls} To allow direct access to bin data.
+ */
+os.data.histo.ColorModel.prototype.getAllBinColors = function() {
+  // start with bins and overrides from this histogram on top of that
+  const colors = Object.assign({}, this.binColors_, this.manualBinColors_);
+  const lookup = {};
+
+  for (const value of Object.values(colors)) {
+    lookup[os.color.toHexString(value)] = true;
+  }
+
+  // get any missing the colors out of the folds from external histograms
+  return this.getResults().reduce(function(all, bin) {
+    if (bin.colorCounts_) {
+      for (var k in bin.colorCounts_) {
+        if (!lookup[k]) {
+          var label = [os.data.histo.OVERRIDE_LABEL, k].join(' ');
+          all[label] = os.style.toRgbaString(k);
+          lookup[k] = true;
+        }
+      }
+    }
+    return all;
+  }, colors);
+};
+
+
+/**
+ * Get the bin/color pairs currently applied by this color model.
+ *
+ * @return {Object<string, string>}
  */
 os.data.histo.ColorModel.prototype.getBinColors = function() {
   return this.binColors_;
@@ -207,7 +238,8 @@ os.data.histo.ColorModel.prototype.setManualBinColors = function(colors) {
   goog.removeUid(colors);
 
   this.manualBinColors_ = colors;
-  this.applyColorMethod();
+
+  // this.applyColorMethod(); // already called by setColorMethod(), so don't double-up the re-draws for all the charts
 };
 
 
@@ -274,6 +306,7 @@ os.data.histo.ColorModel.prototype.setColorMethod = function(value, opt_bins, op
     }
   }
 
+  // keep the underlying method when simply adjusting a few colors manually
   this.colorMethod_ = value === os.data.histo.ColorMethod.MANUAL ? this.colorMethod_ : value;
   this.applyColorMethod();
 };
@@ -471,13 +504,35 @@ os.data.histo.ColorModel.prototype.colorFeatures_ = function(features, opt_color
 
 
 /**
+ * Sets the color on a set of features.
+ *
+ * @param {Array<ol.Feature>} features The features to update
+ * @param {string=} opt_color The new feature color
+ *
+ */
+os.data.histo.ColorModel.prototype.colorFeatures = function(features, opt_color) {
+  if (opt_color) {
+    // add a manual entry for this color
+    var label = [os.data.histo.OVERRIDE_LABEL, os.color.toHexString(opt_color)].join(' ');
+    this.manualBinColors_[label] = opt_color;
+  }
+  this.colorFeatures_(features, opt_color);
+};
+
+
+/**
  * @inheritDoc
  */
 os.data.histo.ColorModel.prototype.persist = function(opt_to) {
   var obj = opt_to || {};
 
   obj['colorMethod'] = this.colorMethod_;
-  obj['manualColors'] = goog.object.clone(this.manualBinColors_);
+  obj['manualColors'] = {};
+
+  // only grab manual bin colors, because features will be different next load/search
+  for (const [label, color] of Object.entries(/** @type {!Object<?, string>} */ (this.manualBinColors_))) {
+    if (label.indexOf(os.data.histo.OVERRIDE_LABEL) == -1) obj['manualColors'][label] = color;
+  }
 
   if (this.histogram_) {
     var binMethod = this.histogram_.getBinMethod();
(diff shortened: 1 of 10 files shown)


src/os/data/histo/colormodel.js
```diff
@@ -207,7 +237,8 @@ os.data.histo.ColorModel.prototype.setManualBinColors = function(colors) {
   goog.removeUid(colors);
 
   this.manualBinColors_ = colors;
-  this.applyColorMethod();
+
+  // this.applyColorMethod(); // already called by setColorMethod(), so don't double-up the re-draws for all the charts
```
schmidtk: This may cause unwanted side effects in `restore`, where `setManualBinColors` is called last. Given the comment there about intentionally calling it last, it seems we can't just move that call. I am curious about that comment because if the color method isn't manual, there shouldn't be manual colors anyway (they would have been cleared when the color model was changed).

src/os/source/vectorsource.js
```diff
@@ -987,6 +987,55 @@ os.source.Vector.prototype.setColumnAutoDetectLimit = function(value) {
...
+os.source.Vector.prototype.setColor = function(items, color) {
+  if (!items || !color) return;
+
+  var cm = this.colorModel;
+  if (cm) {
+    cm.colorFeatures(items, color);
+  } else {
+    // TODO figure out how to do the binning when there are no bins...
+    for (const item of items) {
+      var oldColor = os.color.toHexString(os.feature.getColor(item, undefined, undefined));
```
schmidtk: - I would expect this function to always set `os.data.RecordField.COLOR` on the feature, so even if that matches the current feature color (ie, layer color or feature style config) the override will remain if the layer/feature style are changed.
- Using `os.feature.getColor` is great to avoid unnecessary events, but isn't a very performant function to call on a lot of features. It may be overkill given my previous point.
- Passing `undefined` to a function is superfluous if you aren't providing later arguments, because `undefined` is the default value when a parameter isn't provided.

src/os/source/vectorsource.js
```diff
@@ -987,6 +987,55 @@ os.source.Vector.prototype.setColumnAutoDetectLimit = function(value) {
...
+
+
+/**
+ * Finds the first color in the items so the User has some consistency
+ *
+ * @param {Array<!ol.Feature>|null} items
+ * @return {string} the color
+ */
+os.source.Vector.prototype.getFirstColor = function(items) {
+  var str = 'rgba(255,255,255,1)';
```
schmidtk: Consider using a defined constant like `os.style.DEFAULT_LAYER_COLOR`.

src/os/source/vectorsource.js
```diff
@@ -987,6 +987,55 @@ os.source.Vector.prototype.setColumnAutoDetectLimit = function(value) {
...
+  if (!items) return str;
+  var color;
+
+  for (const item of items) {
+    color = /** @type {string} */ (item.get('color'));
```
schmidtk: Same here - did you intend `os.data.RecordField.COLOR`?

src/os/source/vectorsource.js
```diff
@@ -987,6 +987,55 @@ os.source.Vector.prototype.setColumnAutoDetectLimit = function(value) {
...
+};
+
+
+/**
+ * Finds the first color in the items so the User has some consistency
+ *
+ * @param {Array<!ol.Feature>|null} items
+ * @return {string} the color
+ */
+os.source.Vector.prototype.getFirstColor = function(items) {
```
schmidtk: This function seems better suited as `os.feature.getFirstColor`, as it doesn't relate to the source at all.

src/os/ui/menu/listmenu.js
```diff
@@ -193,6 +215,52 @@ os.ui.menu.list.onSortSelected = function(event) {
...
+ * @param {os.ui.menu.MenuEvent} event The menu event.
+ */
+os.ui.menu.list.onColorSelected = function(event) {
+  var context = event.getContext();
+  if (os.instanceOf(context, os.source.Vector.NAME)) {
+    var source = /** @type {!os.source.Vector} */ (context);
+    if (source && source.getSelectedItems()) {
+      var items = source.getSelectedItems();
+      var callback = goog.bind(
+          function(source, items, color) {
```
schmidtk: Inline functions will inherit variables from the parent closure, so it isn't necessary to bind the function here. Your callback will already have access to `source` and `items` via the parent.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures

When you do need to use bind, prefer the native `bind` call (`function(...) {}.bind(context, args)`) over `goog.bind`.

src/os/source/vectorsource.js
jsalankey: I would also add that calls to `os.color.toHexString` are slow. We should do extensive performance testing with large datasets with these changes.

glennkitchellcaci: Sounds good.  This is the same way that the ColorModel colors features, so whatever improvements we make here should also be applied there and anywhere else that modifies a lot of feature colors at once -- likely making all of them call a new utility function in the Source.

Also, I needed the old color in order to fire the COLOR featureEvent -- which I think is what the UI uses to recolor the Feature in cesium.   Is it okay to just pretend oldColor was null?  (I'll make the change see what happens locally, but even if it works, I'm wondering if it's an OK thing to do)

src/os/data/histo/colormodel.js
glennkitchellcaci: I also moved "this.setColorMethod(...);" to the bottom in restore() for exactly that reason.  Is there a reason why it must be called before the setManualBinColors()?  In my testing I couldn't find any strange cases where it didn't load the manual bin colors properly from a restore with setColorMethod() last instead of setManualBinColors() last.

src/os/source/vectorsource.js
glennkitchellcaci: Here we go -- rather than use the nice getColor() method that goes and looks everywhere, I just do a raw values_[feature....COLOR] comparison.  Best of both worlds.  Also, I removed the toHexString() call and made it so the passed-in color is an RGBA instead of hex so the conversion is a one-time thing instead of a million-time thing.

src/os/data/histo/colormodel.js
glennkitchellcaci: Moving setColorMethod to the bottom was working for AUTO_COLOR, but had to add handler for MANUAL method.

src/os/ui/menu/listmenu.js
glennkitchellcaci: True that.  Updated here and a few other places that I got this from.

src/os/ui/menu/listmenu.js
```diff
@@ -193,6 +215,46 @@ os.ui.menu.list.onSortSelected = function(event) {
...
+os.ui.menu.list.onColorSelected = function(event) {
+  var context = event.getContext();
+  if (os.instanceOf(context, os.source.Vector.NAME)) {
+    var source = /** @type {!os.source.Vector} */ (context);
+    if (source && source.getSelectedItems()) {
+      var items = source.getSelectedItems();
+
+      // call the confirm window from this context so it doesn't appear in the wrong tab
+      os.ui.window.launchConfirmColor(function(color) {
+        source.setColor(items, os.style.toRgbaString(color));
```
glennkitchellcaci: @jsalankey Calling toRgbaString() here, outside the loop, instead of toHexString() in setColor() inside the loop.  Ty ty.

src/os/data/histo/colormodel.js
```diff
@@ -176,6 +176,36 @@ os.data.histo.ColorModel.prototype.getResults = function() {
 };
 
 
+/**
+ * Get the bin/color pairs currently applied by this color model.
+ *
+ * @return {Object<string, string>}
+ * @suppress {accessControls, checkTypes} To allow direct access to bin data.
```
jsalankey: Do you need the `checkTypes` suppression here? We really don't like to use that one unless absolutely required.

src/os/ui/menu/listmenu.js
```diff
@@ -193,6 +215,46 @@ os.ui.menu.list.onSortSelected = function(event) {
...
+ */
+os.ui.menu.list.onResetColor = function(event) {
+  var context = event.getContext();
+  if (os.instanceOf(context, os.source.Vector.NAME)) {
+    var source = /** @type {!os.source.Vector} */ (context);
+    if (source && source.getColorModel()) {
+      var cm = source.getColorModel();
+      if (cm) {
+        cm.setColorMethod(os.data.histo.ColorMethod.RESET);
+        cm.setColorMethod(os.data.histo.ColorMethod.MANUAL);
```
jsalankey: Calling both of these back to back seems odd to me. Should the reset call just set it back to manual automatically?

glennkitchellcaci: Yes and no.  I made some other changes that OBE'd this requirement (it used to have to re-enable the manual colors to pick up the colors off the features).  I put this (and the others) back to only resetting.

PR #870, opened 2020-02-11
feat(fog): fog menu items can be hidden via support setting

Xenotide:
Addition to fog display settings to hide the menu items if fog is unsupported

### DIFF
--- a/src/os/config/displaysettings.js
+++ b/src/os/config/displaysettings.js
@@ -77,7 +77,7 @@ os.config.isTerrainConfigured = function() {
  * @return {boolean}
  */
 os.config.isFogConfigured = function() {
-  var fogSupport = /** @type {boolean|undefined} */ os.settings.get(os.config.DisplaySetting.FOG_UNSUPPORTED);
+  var fogSupport = /** @type {boolean|undefined} */ (os.settings.get(os.config.DisplaySetting.FOG_UNSUPPORTED));
   return !(fogSupport);
 };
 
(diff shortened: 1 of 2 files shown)


src/os/config/displaysettings.js
```diff
@@ -50,6 +50,7 @@ os.config.DisplaySetting = {
   CAMERA_STATE: os.config.DisplaySettings.BASE_KEY + 'cameraState',
   CAMERA_MODE: os.config.DisplaySettings.BASE_KEY + 'cameraMode',
   MAP_MODE: os.config.DisplaySettings.BASE_KEY + 'mapMode',
+  FOG_UNSUPPORTED: os.config.DisplaySettings.BASE_KEY + 'fogUnsupported',
```
wallw-bits: Generally we try to use positive boolean values rather than negative ones. E.g. `fogSupported` rather than `fogUnsupported`.

Also, we generally have "supported" things as feature detections rather than configurable keys. If this was done more correctly (and that's not really on you), then the WebGL renderer would supply its own settings page.

src/os/config/displaysettings.js
```diff
@@ -70,6 +71,15 @@ os.config.isTerrainConfigured = function() {
   return !!(options && options.type);
 };
 
+/**
+ * If fog has been configured in the application.
+ *
+ * @return {boolean}
+ */
+os.config.isFogConfigured = function() {
+  var fogSupport = /** @type {boolean|undefined} */ (os.settings.get(os.config.DisplaySetting.FOG_UNSUPPORTED));
```
wallw-bits: For clarity, this should probably provide a default value if this key is unconfigured: `os.settings.get(SOME_KEY, false)` in this case, or `os.settings.get(FOG_SUPPORTED, true)` if the name is changed to positive.

PR #873, opened 2020-02-11
fix(alert-manager): adds instance setter for the alert manager

jsalankey:
There are cases under the module refactor where the singleton getter
implementation wasn't working quite properly. This allows for setting
a specific instance in those cases.


jsalankey:
Fixes GEOVIZ-109

PR #877, opened 2020-02-14
feat(scope): don't allow os.ui.apply to apply too often

csgibson:
~~This will not allow `os.ui.apply` to apply the scope more often than 5 times a second.~~

~~Open to changing the value for `os.ui.TIME_BETWEEN_SCOPE_UPDATES`.~~

EDIT: This is no longer the approach

### DIFF
--- a/src/os/ui/ui.js
+++ b/src/os/ui/ui.js
@@ -203,7 +203,7 @@ os.ui.apply = function(scope, opt_delay) {
   } else {
     try {
       if (scope && (!scope.$root || !scope.$root.$$phase)) {
-        scope.$apply();
+        scope.$applyAsync();
       }
     } catch (e) {
     }


gregroyal:
This is an interesting change. Whats the reasoning behind it?

csgibson:
> 
> 
> This is an interesting change. Whats the reasoning behind it?

We've been looking into performance testing with our plugin to investigate drops in frame rates. We've seen `os.ui.apply` trigger on intervals as small as 5ms (not often) but frequently it will be called ~10 times per second.

gregroyal:
This might be a nice safety net change overall (maybe lower the time to 100ms?) however its a safety net. You're specific use case might need another fix. We've always fixed this issue on a case-by-case basis where its needed that specific directive has its own delay. I'm a little worried about putting in such a core safety might slow down updates and make the app feel more sluggish just to fix a specfic issue (which with this will be harder to find)

csgibson:
> 
> 
> maybe lower the time to 100ms?
Done

>
>
> We've always fixed this issue on a case-by-case basis where its needed that specific directive has its own delay
I don't see a delay used anywhere, maybe I'm missing something

Also have this to fix a specific case:
https://github.com/ngageoint/opensphere/pull/878

gregroyal:
os.ui.query.BaseCombinatorCtrl.prototype.update is an example, there's a this.updateDelay that throttles the update to be only once every 50 seconds

csgibson:
> 
> 
> os.ui.query.BaseCombinatorCtrl.prototype.update is an example, there's a this.updateDelay that throttles the update to be only once every 50 seconds

I see what you mean, the update is outside of the call to `os.ui.apply` as opposed to using the second parameter to add a delay. 

I've simplified this quite a bit to just switch to applying on the next digest cycle. 
ref: https://code.angularjs.org/snapshot-stable/docs/api/ng/type/$rootScope.Scope#$applyAsync

src/os/ui/ui.js
```diff
@@ -203,7 +203,7 @@ os.ui.apply = function(scope, opt_delay) {
   } else {
     try {
       if (scope && (!scope.$root || !scope.$root.$$phase)) {
-        scope.$apply();
+        scope.$applyAsync();
```
gregroyal: I like this alot more. I'm not sure you need the if statement anymore right? the purpose of that was to check if its in a digest to not call apply

schmidtk: I do like the convenience of a built-in null check, similar to `goog.dispose`. That was initially to protect against attempts to call apply after the controller was destroyed and scope nulled without wrapping in an explicit `if (this.scope)`.

schmidtk:
I'm not certain if we have code that relies on calling `$apply` synchronously to do a digest before executing further code, but the current function also doesn't guarantee a digest will happen immediately. If Angular is already in a digest, nothing will happen. I'll go ahead and merge, but if anyone has input on this let me know.

PR #880, opened 2020-02-17
feat: allow custom wfs queries to be built from an existing request

welchyd:
~~BREAKING CHANGE: added a param to queryhandler's createfilter function~~
Worked around the part that was breaking

### DIFF
--- a/src/os/ui/query/queryhandler.js
+++ b/src/os/ui/query/queryhandler.js
@@ -8,6 +8,16 @@ goog.require('os.filter.IFilterFormatter');
 goog.require('os.filter.ISpatialFormatter');
 
 
+/**
+ * @typedef {{
+ *   entries: !Array<!Object<string, string|boolean>>,
+ *   includes: !Array<!Object<string, string|boolean>>,
+ *   excludes: !Array<!Object<string, string|boolean>>
+ * }}
+ */
+os.ui.query.ActiveEntries;
+
+
 
 /**
  * Class for representing a query consisting of areas and filters. Contains no actual handling logic, just
@@ -307,12 +317,13 @@ os.ui.query.QueryHandler.prototype.getEntries = function(id) {
 
 /**
  * Gets entries for the filter
- * @param {!Array<!Object<string, string|boolean>>} entries
- * @param {!Array<!Object<string, string|boolean>>} includes
- * @param {!Array<!Object<string, string|boolean>>} excludes
+ * @return {os.ui.query.ActiveEntries}
  */
-os.ui.query.QueryHandler.prototype.getActiveEntries = function(entries, includes, excludes) {
+os.ui.query.QueryHandler.prototype.getActiveEntries = function() {
   var qmEntries = this.getEntries(this.getLayerId());
+  var entries = [];
+  var includes = [];
+  var excludes = [];
 
   // clone the entries
   for (var i = 0, n = qmEntries.length; i < n; i++) {
@@ -322,26 +333,26 @@ os.ui.query.QueryHandler.prototype.getActiveEntries = function(entries, includes
   }
 
   // ignore disabled areas
-  entries = entries.filter(this.shownAreas.bind(this));
+  entries = entries.filter(this.shownAreas, this);
 
   if (excludes) { // exclusion areas must be handled differently
-    var excl = entries.filter(this.excludes.bind(this)).map(this.toExclude.bind(this));
+    var excl = entries.filter(this.excludes, this).map(this.toExclude, this);
     for (var i = 0; i < excl.length; i++) {
       excludes.push(excl[i]);
     }
   }
 
   if (includes) { // inclusion areas
-    var incl = entries.filter(this.includes.bind(this));
+    var incl = entries.filter(this.includes, this);
     for (var i = 0; i < incl.length; i++) {
       includes.push(incl[i]);
     }
 
     if (includes.length) {
-      entries = includes.filter(this.filters.bind(this));
+      entries = includes.filter(this.filters, this);
     } else {
       // Theres no inclusions, check to see if we still have filters to apply
-      entries = entries.filter(this.filters.bind(this));
+      entries = entries.filter(this.filters, this);
 
       // Since this is only for exclusions or no areas. We only need 1 of each filter
       goog.array.removeDuplicates(entries, null, function(entry) {
@@ -349,6 +360,8 @@ os.ui.query.QueryHandler.prototype.getActiveEntries = function(entries, includes
       });
     }
   }
+
+  return /** @type {os.ui.query.ActiveEntries} */ ({entries: entries, includes: includes, excludes: excludes});
 };
 
 
@@ -358,11 +371,10 @@ os.ui.query.QueryHandler.prototype.getActiveEntries = function(entries, includes
  * @return {string}
  */
 os.ui.query.QueryHandler.prototype.createFilter = function() {
-  var entries = [];
-  var includes = [];
-  var excludes = [];
-
-  this.getActiveEntries(entries, includes, excludes);
+  var activeEntries = this.getActiveEntries();
+  var entries = activeEntries.entries;
+  var includes = activeEntries.includes;
+  var excludes = activeEntries.excludes;
 
   // we don't care about layer in the order array
   var order = ['area', 'filter'];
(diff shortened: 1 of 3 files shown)


src/os/ui/query/queryhandler.js
```diff
@@ -203,14 +203,15 @@ os.ui.query.QueryHandler.prototype.setSpatialRequired = function(value) {
 
 
 /**
+ * @param {os.query.BaseAreaManager?} areaManager
  * @param {!Object<string, string|boolean>} item
  * @return {boolean}
  */
-os.ui.query.QueryHandler.includes = function(item) {
+os.ui.query.QueryHandler.includes = function(areaManager, item) {
```
schmidtk: This wouldn't be a breaking change if it were an optional second parameter. Same goes for other places this change was made.

src/os/ogc/wfs/wfsformatter.js
```diff
@@ -8,11 +8,13 @@ goog.require('os.ogc');
...
 /**
  * Create a WFS request payload.
- *
+ * @param {Object<string, string>=} opt_params params to override uri format
  * @implements {os.net.IDataFormatter}
  * @constructor
  */
-os.ogc.wfs.WFSFormatter = function() {};
+os.ogc.wfs.WFSFormatter = function(opt_params) {
+  this.params = opt_params ? opt_params : {};
```
schmidtk: Adding param overrides seems like unnecessary bloat to this class.

The `format` call already reads params from `uri.getQueryData()`. If you need to modify an existing `goog.Uri` instance, use `otherUri.clone()` and modify the clone before passing to `format`.

src/os/ui/query/queryhandler.js
```diff
@@ -298,20 +310,22 @@ os.ui.query.QueryHandler.prototype.createFilter = function() {
     entries.push(entry);
   }
 
-  // ignore disabled areas
-  entries = entries.filter(os.ui.query.QueryHandler.shownAreas);
+  if (!opt_ignoreShown) { // ignore disabled areas
+    entries = entries.filter(os.ui.query.QueryHandler.shownAreas.bind(null, am));
```
schmidtk: I'm seeing why you had the manager parameter first. The bind can easily be removed while still having a concise one-liner by using an arrow function:

```
entries = entries.filter((entry) => os.ui.query.QueryHandler.shownAreas(entry, am));
```

src/plugin/arc/query/arcqueryhandler.js
```diff
@@ -69,7 +69,7 @@ plugin.arc.query.ArcQueryHandler.prototype.getActiveEntries = function() {
   }
 
   // ignore disabled areas
-  entries = entries.filter(os.ui.query.QueryHandler.shownAreas);
+  entries = entries.filter(os.ui.query.QueryHandler.shownAreas.bind(null, null));
```
schmidtk: With the manager parameter second, you'll still need to use an arrow function because `filter` passes `item, index, array` params to the callback. So these will be:
```
entries = entries.filter((entry) => os.ui.query.QueryHandler.includes(entry));
```

src/os/ui/query/queryhandler.js
welchyd: I did try that route, but it collided with the use of that function in `Array.filter` since the additional params didn't align

src/plugin/arc/query/arcqueryhandler.js
welchyd: Ahh, that is exactly what I was looking for - didn't want this be a breaking change but couldn't quite find a way around that

src/os/ui/query/queryhandler.js
```diff
@@ -267,22 +274,27 @@ os.ui.query.QueryHandler.filters = function(item) {
...
 };
 
 
 /**
  * Creates the filter
- *
+ * @param {os.query.BaseQueryManager=} opt_qm
+ * @param {os.query.BaseAreaManager=} opt_am
+ * @param {os.filter.BaseFilterManager=} opt_fm
+ * @param {boolean=} opt_ignoreShown whether or not to filter on shown areas
```
schmidtk: This parameter seems like it could use a better name. When true it includes disabled areas, false/undefined it filters out disabled areas. Maybe `opt_includeDisabled`? The default behavior of the function is to only include shown areas.

src/os/ogc/wfs/wfsformatter.js
```diff
@@ -88,7 +87,8 @@ os.ogc.wfs.WFSFormatter.prototype.format = function(uri) {
   data += '>';
 
   var srs = /** @type {string} */ (wfsParams['srsname']) || os.ogc.defaultProjection;
-  data += '<wfs:Query srsName="' + srs + '" typeName="' + /** @type {string} */ (wfsParams['typename']) + '">';
+  data += '<wfs:Query srsName="' + srs + '" typeName="' +
```
schmidtk: All changes in this file seem to only impact whitespace. Can they be dropped from the PR?

welchyd: Done

src/os/ui/query/queryhandler.js
```diff
@@ -269,55 +269,103 @@ os.ui.query.QueryHandler.filters = function(item) {
...
-  var entries = [];
   for (var i = 0, n = qmEntries.length; i < n; i++) {
     var entry = goog.object.clone(qmEntries[i]);
     entry['spatialRequired'] = this.spatialRequired;
     entries.push(entry);
   }
 
   // ignore disabled areas
-  entries = entries.filter(os.ui.query.QueryHandler.shownAreas);
+  entries = entries.filter(this.shownAreas.bind(this));
```
wallw-bits: This is binding a new instance of the function every time. Let's just do that once or alternatively pass the `thisArg` to filter via `entries = entries.filter(this.shownAreas, this)`

src/os/ui/query/queryhandler.js
```diff
@@ -269,55 +269,103 @@ os.ui.query.QueryHandler.filters = function(item) {
...
 
-  // exclusion areas must be handled differently
-  var excludes = entries.filter(os.ui.query.QueryHandler.excludes).map(os.ui.query.QueryHandler.toExclude);
+  if (excludes) { // exclusion areas must be handled differently
+    var excl = entries.filter(this.excludes.bind(this)).map(this.toExclude.bind(this));
```
wallw-bits: Ditto here for both `filter()` and `map()`.

src/os/ui/query/queryhandler.js
```diff
@@ -457,3 +505,20 @@ os.ui.query.QueryHandler.prototype.wrap = function(filter, order, excludes, grou
 os.ui.query.QueryHandler.prototype.wrapAll = function(filter) {
   return this.filterFormatter ? this.filterFormatter.wrapAll(filter) : '';
 };
+
+
+/**
+ * @param {os.ui.query.QueryHandler?} opt_qh another query handler to clone to
```
wallw-bits: This is better named as `opt_result` or `opt_clone`. Also any optional param should be `{Type=}` not `{Type?}`.

src/os/ui/query/queryhandler.js
```diff
@@ -269,55 +269,103 @@ os.ui.query.QueryHandler.filters = function(item) {
...
+
+
+
+/**
+ * Gets entries for the filter
+ * @param {!Array<!Object<string, string|boolean>>} entries
+ * @param {!Array<!Object<string, string|boolean>>} includes
+ * @param {!Array<!Object<string, string|boolean>>} excludes
+ */
+os.ui.query.QueryHandler.prototype.getActiveEntries = function(entries, includes, excludes) {
```
wallw-bits: This is a bit disingenuous because it is a getter that doesn't return anything. I'd prefer returning the entries over forcing the array to be created elsewhere.

src/plugin/arc/query/arcqueryhandler.js
```diff
@@ -125,21 +121,22 @@ plugin.arc.query.ArcQueryHandler.prototype.createGeometry = function() {
...
       return 'inclusion';
     }
 
-    if (os.ui.query.QueryHandler.excludes(entry) && !seenAreas[areaId]) {
+    if (this.excludes(entry) && !seenAreas[areaId]) {
       seenAreas[areaId] = true;
       return 'exclusion';
     }
-  });
+  }.bind(this));
```
wallw-bits: [`goog.array.bucket`](https://google.github.io/closure-library/api/goog.array.html#bucket) also takes a scope argument rather than forcing binds.

src/os/ui/query/queryhandler.js
welchyd: Good catch!

src/os/ui/query/queryhandler.js
welchyd: I was on the fence about that too, will fix

src/os/ui/query/queryhandler.js
```diff
@@ -8,6 +8,16 @@ goog.require('os.filter.IFilterFormatter');
...
 
 
+/**
+ * @typedef {{
+ *   entries: !Array<!Object<string, string|boolean>>,
+ *   includes: !Array<!Object<string, string|boolean>>,
+ *   excludes: !Array<!Object<string, string|boolean>>
+ * }}
+ */
+os.ui.query.activeEntries;
```
wallw-bits: Nit: this would be the only typedef anywhere in the application that is `camelCase` rather than `TitleCase`

PR #882, opened 2020-02-17
feat(controls): throttle info controls to reduce overhead

csgibson:
Increase performance by reducing the number of updates done by the scale line and zoom info controls. These controls manipulate the dom and do a decent amount of math so we should call them less often.


csgibson:
Although the throttle on these could be lowered, the throttle overhead is probably not worth the amount of savings we'd get.

PR #888, opened 2020-02-20
feat(export): adds settings to control enabled export features

jsalankey:
Export of different state types and which persistence methods are
available can be controlled via a setting.

### DIFF
--- a/src/os/state/basestatemanager.js
+++ b/src/os/state/basestatemanager.js
@@ -216,22 +216,26 @@ os.state.BaseStateManager.prototype.deleteStates = function() {
  */
 os.state.BaseStateManager.prototype.getAvailable = function(opt_allVersions) {
   var list = [];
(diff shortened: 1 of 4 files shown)


src/os/state/basestatemanager.js
```diff
@@ -216,22 +216,26 @@ os.state.BaseStateManager.prototype.deleteStates = function() {
  */
 os.state.BaseStateManager.prototype.getAvailable = function(opt_allVersions) {
   var list = [];
+  var disabledStates = /** @type {Object<string, boolean>} */ (os.settings.get('ex.disabledStates', {}));
```
wallw-bits: Nit: We generally try to avoid things like `disabled = true` and instead prefer `enabled = false`. Similar to how we enable/disable plugins etc.

jsalankey: Heh, I actually had it that way initially and then I switched it to this. I can switch it back.

PR #890, opened 2020-02-23
fix: update World Magnetic Model from v2015v2 to v2020

bradh:
v2015v2 expired at the end of 2019, and we're starting to see other applications / data updated so its probably time.

The WMM COF file was sourced from https://www.ngdc.noaa.gov/geomag/WMM/

Test updates are minor.


jsalankey:
Nice, didn't realize a new one was out. Thanks for this.

PR #891, opened 2020-02-26
Support user selection of Electron client certificates

schmidtk:
`opensphere-electron` PR: https://github.com/ngageoint/opensphere-electron/pull/17

- Add an Electron plugin to handle requests for client certificates.
- Add an Electron preload script to establish IPC with the main process.
- Transform `confirm.js` to ES6 so `launchConfirm` can be imported cleanly. The transform was done with `opensphere-jscodeshift`, then `launchConfirm` added as an export. `os.ui.window.launchConfirm` was added to the shim for backward compatibility, and deprecated.
- Replace all deprecated references as a result of the transform.
- Allow passing the window parent to `launchConfirm`. The Electron cert prompt uses the app container (`#ng-app`), because requests made before the default window container is available will not display a prompt.


glennkitchellcaci:
Looks great!  Awesome rewrite with the ConfirmUI.

PR #895, opened 2020-02-28
fix(histogram): added small optimization to histogram updates

glennkitchellcaci:
Updated histogram to avoid throwing the CASCADE_CHANGE event in one typical case; which saves a lot of re-rendering.

### DIFF
--- a/src/os/data/histo/sourcehistogram.js
+++ b/src/os/data/histo/sourcehistogram.js
@@ -341,10 +341,16 @@ os.data.histo.SourceHistogram.prototype.getCascadeValues = function() {
  * @param {Array<*>} value
  */
 os.data.histo.SourceHistogram.prototype.setCascadeValues = function(value) {
-  this.cascadeValues_ = value;
+  // Don't trigger everything re-rendering (usually again) if swapping from null to null;
+  // NOTE: Visible code that calls this are passing in a new Array with the old items, so the
+  // event will still be thrown and all listeners triggered.  It would be more accurate to
+  // run an item-by-item comparison, but that's too slow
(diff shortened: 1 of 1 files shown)


src/os/data/histo/sourcehistogram.js
```diff
@@ -341,10 +341,14 @@ os.data.histo.SourceHistogram.prototype.getCascadeValues = function() {
  * @param {Array<*>} value
  */
 os.data.histo.SourceHistogram.prototype.setCascadeValues = function(value) {
-  this.cascadeValues_ = value;
+  // don't trigger everything re-rendering (usually again) if swapping from null to null;
+  // it would be more accurate running an item-by-item comparison, but that's too slow
+  if (value != null || this.cascadeValues_ != null) {
```
wallw-bits: This is probably better written as `if (this.cascadeValues_ != value)`, since the point here is to avoid dispatching the event if the items are not changing.

That form of the check also wouldn't need a comment since that is much more common.

glennkitchellcaci: The problem with a simple != is that most of the time it's a different Array with the same items in it, so the two arrays are never equal to eachother.  

The "real" way to do it is with null checks followed by length checks followed by item-by-item comparisons when the lengths are the same, but that'd be way too slow for something that is getting hammered on whenever bins are changing.

glennkitchellcaci: Since != covers the null > null case, I just switched it and added a comment about most uses of the code passing in a new Array with the old values (so the event will still be thrown).

PR #898, opened 2020-03-02
feat(colors): increased support for manual coloring

glennkitchellcaci:
Added support for manual coloring and resetting for Layer, and List tools

### DIFF
--- a/src/os/source/vectorsource.js
+++ b/src/os/source/vectorsource.js
@@ -992,27 +992,18 @@ os.source.Vector.prototype.setColumnAutoDetectLimit = function(value) {
  * does a fast-color on the Features directly.
  *
  * @param {Array<!ol.Feature>|null} items
- * @param {string|null} color rgba-color or clear with a null
+ * @param {string=} opt_color rgba-color or clear with a null or undefined (which colormodel treats differently)
  *
  * @suppress {accessControls}
  */
-os.source.Vector.prototype.setColor = function(items, color) {
-  if (!items || !color) return;
-
-  var cm = this.colorModel;
-  if (cm) {
-    cm.colorFeatures(items, color);
-  } else {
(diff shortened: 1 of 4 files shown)


src/os/source/vectorsource.js
```diff
@@ -1671,6 +1689,16 @@ os.source.Vector.prototype.createColorModel = function(opt_histogram, opt_gradie
 };
 
 
+/**
+ * Check for a color model or manually colored items in the Source
+ *
+ * @return {boolean}
+ */
+os.source.Vector.prototype.hasColors = function() {
+  return (this.colorModel != null || (this.getColoredItems() != null && this.getColoredItems().length > 0));
```
jsalankey: Calling `getColoredItems()` twice here is bad as that call does an `Object.values()` on each call. Calling it once and assigning it to a local variable would be cleaner and more performant.

src/os/source/vectorsource.js
```diff
@@ -1689,6 +1717,24 @@ os.source.Vector.prototype.getColorModel = function() {
...
+    var items = this.getColoredItems();
+    if (items && items.length > 0) {
+      this.setColor(items); // this action clears the values in this.coloredItems_
+      notify = true;
+
+      // WORKAROUND: Notify Map that these features need to be recolored.  This is only called if user
+      // is manually coloring Features without a ColorModel
+      // Events that didn't work: STYLE, FEATURE, HIGHLIGHTED, CHANGED; with and without this.changed()
+      this.setSelectedItems(items); // deselect anything selected, then select (to redraw these)
+      this.setSelectedItems([]); // deselect these
```
jsalankey: Try taking a look at what `os.style.notifyStyleChange` does. There is definitely a better way to do this as this will be slow and wipe out a user's selection.

PR #900, opened 2020-03-03
feat(request): add status code enum to aid error handling

dwlueth:
UUOM-763


src/os/net/request.js
```diff
@@ -220,6 +221,20 @@ os.net.Request.METHOD_DELETE = 'DELETE';
 os.net.Request.METHOD_HEAD = 'HEAD';
 
 
+/**
```
dwlueth: I have a task where I handle an error differently based on the status code.  I thought having the codes in a central location will facilitate doing similar things in the future.

PR #908, opened 2020-03-09
feat(feature-limit): adds settings UI for changing max feature counts

jsalankey:
BREAKING CHANGE: adds methods to IWebGLRenderer to support getting and
setting the max feature count for that renderer.

### DIFF
--- a/src/os/config/displaysettings.js
+++ b/src/os/config/displaysettings.js
@@ -127,7 +127,9 @@ os.config.DisplaySettingsCtrl = function($scope) {
         'fog or reducing density may degrade application performance.',
     'sky': 'Show the sky/stars around the 3D globe.',
     'sunlight': 'Light the 3D scene with the Sun.',
-    'terrain': 'Show terrain on the 3D globe.'
+    'terrain': 'Show terrain on the 3D globe.',
+    'help2D': 'Maximum feature count for 2D mode. Valid range: 100-50,000.',
+    'help3D': 'Maximum feature count for 3D mode. Valid range: 100-2,000,000'
   };
 
   /**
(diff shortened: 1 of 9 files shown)


src/os/config/displaysettings.js
```diff
@@ -522,3 +533,21 @@ os.config.DisplaySettingsCtrl.prototype.updateSunlight = function() {
 os.config.DisplaySettingsCtrl.prototype.updateTerrain = function() {
   this.updateSetting_(os.config.DisplaySetting.ENABLE_TERRAIN, this['terrainEnabled']);
 };
+
+
+/**
+ * Update the max feature count.
+ * @param {number} update3D Whether to update 3D or 2D.
```
schmidtk: `boolean`?

src/plugin/cesium/cesiumrenderer.js
```diff
@@ -653,3 +653,19 @@ plugin.cesium.CesiumRenderer.prototype.flyToFeatures = function(features) {
     os.commandStack.addCommand(cmd);
   }
 };
+
+
+/**
+ * @inheritDoc
+ */
+plugin.cesium.CesiumRenderer.prototype.getMaxFeatureCount = function() {
+  return /** @type {number} */ (os.settings.get('maxFeatures.cesium', 150000));
```
schmidtk: This could be implemented in `AbstractWebGLRenderer` without a breaking change by adding a class property for the settings key. Defaulting the key to `maxFeatures.3d` would be backward compatible with existing settings, and renderers would only need to change the property if they wanted their max feature count persisted separately from the default Cesium renderer.

src/os/config/displaysettings.js
jsalankey: Woops. Originally had this passing the new count as the argument.

src/plugin/cesium/cesiumrenderer.js
jsalankey: Removed the subclass functions and switch to using a key on the Abstract.

PR #910, opened 2020-03-10
add ability to position a window by 'bottom'

johnstacy:


johnstacy:
This can be closed, requires more testing.

PR #911, opened 2020-03-11
position a window by 'bottom'

johnstacy:
This ticket does two things:

1. Fix a bug where if the window was not resizable, the double-click action on the header would not work...and generate a console error
2.  Add ability to position a window by the bottom of the window. It now allows the 'bottom' scope variable. This will ignore the 'y' scope variable if both are passed in.

There is currently no way to test the bottom positioning in opensphere itself since no components use it to launch. You can test that it doesn't break any existing windows, however, by launching windows and verifying they behave as they did before.

### DIFF
--- a/src/os/ui/window.js
+++ b/src/os/ui/window.js
@@ -135,7 +135,8 @@ os.ui.windowDirective = function() {
       'windowContainer': '@',
       /* Array.<os.ui.window.HeaderBtnConfig> */
       'headerBtns': '=?',
-      'border': '='
+      'border': '=',
+      'bottom': '=?'
     },
     templateUrl: os.ROOT + 'views/window/window.html',
     controller: os.ui.WindowCtrl,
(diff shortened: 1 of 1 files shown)


src/os/ui/window.js
```diff
@@ -660,7 +661,10 @@ os.ui.WindowCtrl = function($scope, $element, $timeout) {
     $scope['x'] = (container.width() - $scope['width']) / 2;
   }
 
-  if ($scope['y'] == 'center') {
+  if ($scope['bottom'] && $scope['bottom'] != 'auto') {
```
wallw-bits: This check (and all the subsequent checks below) for `$scope['bottom']` do not properly handle the case where `bottom` is `0`, which is falsey. I think you want `$scope['bottom'] != null` instead.

wallw-bits:
I was testing by dropping this in the console:
```javascript
var html = '<div class="modal-body d-flex flex-column flex-fill"><p>This is a test</p></div>';
var options = {
  'id': 'test',
  'width': 400,
  'height': 300,
// remove this section to test collapse fix
  'min-height': 200,
  'min-width': 250,
  'max-width': 1200,
// end section to remove
  'label': 'Test',
  'icon': 'fa fa-cube',
  'show-close': true,
  'modal': false,
  'bottom': 0
};
// Create the window
os.ui.window.create(options, html);
```

src/os/ui/window.js
johnstacy: good catch. fixed.

PR #912, opened 2020-03-12
fix(selection): cleanup related to feature selection

glennkitchellcaci:
Added some null checks and utilized more streamlined methods of selection in some menus.  Part of a cleanup effort.

Test Steps:
1. [opensphere](http://glorious-soccerball.surge.sh/opensphere/)
2. Load some data, Click on a few dots to select them
3. Right click on the Features layer in the Layers modal > Click "Clear Selection" > Verify all selected dots become unselected
4. Shift+Click and drag a box over some loaded dots > Click "Select" in the popup menu > Verify the appropriate dots become selected
5. Right click on the Map > Click "Clear Selection" > Verify all selected dots become unselected

### DIFF
--- a/src/os/ui/menu/listmenu.js
+++ b/src/os/ui/menu/listmenu.js
@@ -1,9 +1,6 @@
 goog.provide('os.ui.menu.list');
 
 goog.require('os.command.FeaturesVisibility');
-goog.require('os.command.InvertSelect');
-goog.require('os.command.SelectAll');
-goog.require('os.command.SelectNone');
 goog.require('os.feature');
 goog.require('os.fn');
 goog.require('os.instanceOf');
--- a/src/os/ui/menu/listmenu.js
+++ b/src/os/ui/menu/listmenu.js
@@ -127,18 +127,18 @@ os.ui.menu.list.handleListEvent = function(event) {
     var eventType = event.type.replace(os.ui.menu.list.PREFIX_REGEXP, '');
     switch (eventType) {
       case os.action.EventType.SELECT:
-        // omit selection events from the stack to reduce clutter
-        new os.command.SelectAll(source.getId()).execute();
+        // don't create and execute a command if it's simple and we don't want it on the stack
+        source.selectAll();
         os.metrics.Metrics.getInstance().updateMetric(os.metrics.FeatureList.SELECT_ALL, 1);
         break;
       case os.action.EventType.DESELECT:
-        // omit selection events from the stack to reduce clutter
-        new os.command.SelectNone(source.getId()).execute();
+        // don't create and execute a command if it's simple and we don't want it on the stack
+        source.selectNone();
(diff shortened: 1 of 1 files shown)


src/os/source/vectorsource.js
```diff
@@ -3344,7 +3344,7 @@ os.source.Vector.prototype.selectAll = function() {
  * @inheritDoc
  */
 os.source.Vector.prototype.selectNone = function() {
-  var i = this.selected_.length;
+  var i = (this.selected_) ? this.selected_.length : 0;
```
jsalankey: This variable is declared as:
```
/**
   * @type {!Array<!ol.Feature>}
   * @private
   */
  this.selected_ = [];
```

Which means it should never be null or undefined. If that's happening, something else is going wrong.

src/os/ui/menu/listmenu.js
```diff
@@ -127,18 +127,18 @@ os.ui.menu.list.handleListEvent = function(event) {
     var eventType = event.type.replace(os.ui.menu.list.PREFIX_REGEXP, '');
     switch (eventType) {
       case os.action.EventType.SELECT:
-        // omit selection events from the stack to reduce clutter
-        new os.command.SelectAll(source.getId()).execute();
+        // don't create and execute a command if it's simple and we don't want it on the stack
+        if (source) source.selectAll();
```
jsalankey: Might be cleaner to wrap this entire thing in `if (source)` if there's a chance that it can be null. All of these calls use it.

src/os/source/vectorsource.js
glennkitchellcaci: Removed!

src/os/ui/menu/listmenu.js
glennkitchellcaci: Removed!  There was a check above it already.

src/os/ui/menu/listmenu.js
```diff
@@ -127,18 +127,18 @@ os.ui.menu.list.handleListEvent = function(event) {
     var eventType = event.type.replace(os.ui.menu.list.PREFIX_REGEXP, '');
     switch (eventType) {
       case os.action.EventType.SELECT:
-        // omit selection events from the stack to reduce clutter
-        new os.command.SelectAll(source.getId()).execute();
+        // don't create and execute a command if it's simple and we don't want it on the stack
```
schmidtk: Minor nit - could you also remove the `goog.require` statements for these? The ES6 transform should remove them automatically if they're not used (once we run it), but we should try to clean those up as file dependencies change.

schmidtk: Thanks for the cleanup!

glennkitchellcaci: Good to clean this up while we're in here 👍

PR #918, opened 2020-03-25
fix(style): added more options when pushing style events

glennkitchellcaci:
Now developers can optionally send event(s) on the Source and ColorModel in addition to the Layer.  Also, a starter module for Text has been added; which will reduce the burden when synchronizing and updating recurring text across multiple widgets, buttons, and menu items.

Testing Steps: 

1. Opensphere|http://reflective-grub.surge.sh/opensphere/
2. Add a layer and load some data > Right-click the layer in the Layers modal > Click "Show Features"
3. On the map, Ctrl+Click-and-drag an area inside your loaded data > Click "Select"
4. In the features modal, scroll to where some of the selected features are listed (they will have a green row highlight) > Right-click
5. Verify there are "Color Selected" and "Reset Color" options in the context menu
6. Verify there is an appropriate on-hover tooltip text for both
7. In the Layers modal, Right-click the layer > Click "Feature Actions" > Create an action that changes the color of some of the features (e.g. based on ALT gte 11000)
8. In the Features modal, verify the color dot of the appropriate features has updated
9. On the map, verify the color color dot of the appropriate features has updated
10. In the Feature Actions modal, uncheck the feature action and click "Apply"
11. In the Features modal, verify the color dot of the appropriate features has updated
12. On the map, verify the color color dot of the appropriate features has updated


src/os/lang/en-us.js
```diff
@@ -0,0 +1,35 @@
...
+  ),
+  MENU_COLOR_SELECTED: new Text(
+      undefined,
+      'Color Selected',
+      'Choose a color for the selected item(s)'
+  )
+};
+// may want to leave "constants" unfrozen so external modules can extend or alter it; either directly or
+// through a yet-to-be-defined interface
+Object.freeze(constants);
```
wallw-bits, line 32: The compiler will enforce this. No need to call `Object.freeze()` on enums. Where you want to use this, you would use `@type {os.lang}`, which I would also name differently so it has ClassCase like the rest of our enums (hopefully, I don't think anything actually checks that).

If you need a place to make these string constants, let's just keep them in a single `@enum {string}` rather than using this `Text` class, since that is more generically useful everywhere.

src/os/text.js
```diff
@@ -0,0 +1,22 @@
...
+class Text {
+  /**
+   * Constructor for a Text
+   * @param {string=} id a key, CSS class, etc to help find the element
+   * @param {string=} label the visible string
+   * @param {string=} tooltip the string shown on hover
+   * @param {string=} description a long explanation of the button or feature
+   */
+  constructor(id = '0', label = '', tooltip = '', description = '') {
+    this.id = id;
```
wallw-bits, line 15: I don't think this is a good association of data. If you want constants for strings so they are reusable, a simple enum will suffice:
```
/**
 * @enum {string}
 */
const StringConstants = {
  MENU_SELECTED_LABEL: 'blah',
  MENU_SELECTED_TOOLTIP: 'blah blah',
  MENU_SELECTED_DESCRIPTION: 'blah blah blah',
};
```
Then this is more reusable for other items too, which may not have ids, tooltips, or descriptions.

src/plugin/featureaction/featurestyleaction.js
```diff
@@ -177,7 +190,20 @@ plugin.im.action.feature.StyleAction.prototype.execute = function(items) {
...
+      source.setColor(items, color); // set the color model's override for these items
+    }
+
+    os.style.notifyStyleChange(
+        layer,
+        items,
+        undefined,
+        undefined,
+        (source && color) // bump the colormodel so dependencies can update/re-render
+    );
```
wallw-bits, line 206: This and the above block look nearly identical. Can we simplify that to a function?

test/os/style/style.test.js
```diff
@@ -379,4 +381,106 @@ describe('os.style', function() {
       expect(os.style.hasNonZeroStrokeOpacity(style)).toBe(false);
     });
   });
+
```
wallw-bits, line 384: Thanks for adding tests! 😍

src/plugin/featureaction/featurestyleaction.js
glennkitchellcaci: The usage of this new colormodel option within "notifyStyleChange()" is pretty embedded in the meat of the function.  I could add a "notifyStyleChangeAndTriggerColorModelListeners()" but because it'd just be calling notifyStyleChange() anyway, it'd be a copy-paste function that'd have to be kept in sync with any future improvements/modifications to notifyStyleChange().

The way it's written now has all of the options to trigger style event(s) on the layer, the source, and/or now the colormodel in a single utility function.  Though often redundant (depending on the interactions of the ways the various plugins set up their listeners), it's pretty nice for getting exactly the event(s) you want.  We'll definitely need to look into how these basic styles and color change events are managed/pushed in the future; but that's asking to refactor all of the listeners in all of the OS plugins... so probably part of the next rewrite.

glennkitchellcaci: Derp.  Looking at the MR, now I see what you're talking about -- I thought you were referring to the before and after.  Not the Do and Undo in the action itself.  I'll make the updated.

PR #919, opened 2020-03-27
fix(filter): support multi-line strings for local is like (list) ops

schmidtk:
The `is like` and `is like list` operations only worked on single line strings. Added the `m` flag to also match the pattern in multi-line strings.

### DIFF
--- a/test/os/ui/filter/op/notlikeop.test.js
+++ b/test/os/ui/filter/op/notlikeop.test.js
@@ -1,6 +1,6 @@
 goog.require('os.ui.filter.op.NotLike');
 
-ddescribe('os.ui.filter.op.NotLike', function() {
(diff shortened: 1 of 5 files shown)


test/os/ui/filter/op/notlikeop.test.js
```diff
@@ -1,6 +1,6 @@
 goog.require('os.ui.filter.op.NotLike');
 
-describe('os.ui.filter.op.NotLike', function() {
+ddescribe('os.ui.filter.op.NotLike', function() {
```
glennkitchellcaci, line 3: Double-d

schmidtk: Thank you! Fixed. I wonder if we could scan for `ddescribe` and `iit` to fail CI/PR builds when present.

Issue #851, opened 2020-01-24
All calls to getExtent should be generic bounding volumes


wallw-bits:
All places which use `getExtent()` (from layers, geometries; aggregated or single, etc.) should really be using a templated bounding volume type. This would be extents in the 2D OpenLayers view and 3D BoundingSpheres in the 3D Cesium View.

Currently, any use of `getExtent()` related to moving/fitting the view for 3D can be incorrect if those features contain altitude (the camera can zoom past the features). We have somewhat addressed this in the feature-specific flyTo (`os.feature.flyTo`) used by KML, selected features context menu in the "Show Features" list, and selected features "Go To" in the Timeline. However, this needs to be done everywhere, especially for the layer menu.

PR #924, opened 2020-04-03
feat(goto): use newer flyto for feature layers

csgibson:
WIP!

Found this issue when importing a csv that had an ellipse at altitude. When the ellipse was styled as such, go to seemed to work fine. When it was just the center point it went straight to the ground below based on the xy extent.

Test Steps:
Load several layers of different types
Ensure that "Go To" from the context menu works on all of them and any combination of them

### DIFF
--- a/src/os/ui/menu/layermenu.js
+++ b/src/os/ui/menu/layermenu.js
@@ -367,11 +367,27 @@ os.ui.menu.layer.onFeatureList_ = function(event) {
  * @private
  */
 os.ui.menu.layer.onGoTo_ = function(event) {
-  var extent = os.ui.menu.layer.getLayersFromContext(event.getContext())
-      .reduce(os.fn.reduceExtentFromLayers, ol.extent.createEmpty());
+  // aggregate the features and execute os.feature.flyTo, in case they have altitude and pure extent wont cut it
+  const layers = os.ui.menu.layer.getLayersFromContext(event.getContext());
+  const features = layers.reduce((feats, layer) => {
+    let source = layer.getSource();
+    if (source instanceof ol.source.Vector) {
+      source = /** @type {ol.source.Vector} */ (source);
+      const newFeats = source.getFeatures();
+      return newFeats.length > 0 ? feats.concat(newFeats) : feats;
+    }
+    return feats;
+  }, []);
 
-  if (!ol.extent.isEmpty(extent)) {
-    os.commandStack.addCommand(new os.command.FlyToExtent(extent));
+  if (features && features.length) {
+    os.feature.flyTo(features);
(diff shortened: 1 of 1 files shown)


schmidtk:
The high level issue here is documented in #851.

src/os/ui/menu/layermenu.js
```diff
@@ -367,11 +367,29 @@ os.ui.menu.layer.onFeatureList_ = function(event) {
...
-  var extent = os.ui.menu.layer.getLayersFromContext(event.getContext())
-      .reduce(os.fn.reduceExtentFromLayers, ol.extent.createEmpty());
+  // aggregate the features and execute os.feature.flyTo, in case they have altitude and pure extent wont cut it
+  const layers = os.ui.menu.layer.getLayersFromContext(event.getContext());
+  const features = layers.reduce((feats, layer) => {
+    let source = layer.getSource();
+    if (source instanceof os.source.Request) {
+      source = /** @type {os.source.Request} */ (source);
+      return feats.concat(
+        source.getFeatureCount() > 0 ?
```
wallw-bits, line 377: This branch is a little bit cleaner little cleaner both speed and memory-wise if you do:
```javascript
return source.getFeatureCount() ?
  feats.concat(source.getFeatures()) :
  feats;
```
since `feats.concat([])` is effectively `feats.slice()`; i.e. copies to a new array.

src/os/ui/menu/layermenu.js
```diff
@@ -367,11 +367,28 @@ os.ui.menu.layer.onFeatureList_ = function(event) {
  * @private
  */
 os.ui.menu.layer.onGoTo_ = function(event) {
-  var extent = os.ui.menu.layer.getLayersFromContext(event.getContext())
-      .reduce(os.fn.reduceExtentFromLayers, ol.extent.createEmpty());
+  // aggregate the features and execute os.feature.flyTo, in case they have altitude and pure extent wont cut it
+  const layers = os.ui.menu.layer.getLayersFromContext(event.getContext());
+  const features = layers.reduce((feats, layer) => {
+    let source = layer.getSource();
+    if (source instanceof os.source.Request) {
```
wallw-bits, line 374: This was probably my fault for not noticing this last time, but why is this checking for `instanceof os.source.Request` over just `ol.source.Vector`?

schmidtk: Possibly a typo, but that should be `os.source.Vector` (not `ol`) given the use of `getFeatureCount`. We could support OL sources as well if we skip the `getFeatureCount` call, assign `source.getFeatures()` to something, and do the `concat` if that's defined and not empty.

PR #937, opened 2020-04-14
WIP: feat(search): additional wildcard search param

swtyler:
We've found multiple situations where it would be useful to "force" wildcards in certain searches rather than expecting users to append a * every time.


swtyler:
We found a more appropriate place for this search parameter/code to live, closing this request

PR #943, opened 2020-04-17
improvement(slickgrid): context menu position not rely on external ids

kwilson-bits:
There was an issue discovered where the wrapper for SlickGrid was using an external HTML ID for the positioning of context menus. That works fine as long as that ID starts at the top left corner of the view port.

### DIFF
--- a/src/os/ui/slick/slickgrid.js
+++ b/src/os/ui/slick/slickgrid.js
@@ -1107,10 +1107,15 @@ os.ui.slick.SlickGridCtrl.prototype.onContextMenu_ = function(event, opt_positio
         os.ui.openMenu(menu, {x: menuX, y: menuY});
       }
     } else if (menu instanceof os.ui.menu.Menu) {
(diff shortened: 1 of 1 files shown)


src/os/ui/slick/slickgrid.js
```diff
@@ -1107,10 +1107,15 @@ os.ui.slick.SlickGridCtrl.prototype.onContextMenu_ = function(event, opt_positio
         os.ui.openMenu(menu, {x: menuX, y: menuY});
       }
     } else if (menu instanceof os.ui.menu.Menu) {
+      const rect = this.element[0].getBoundingClientRect();
+
+      menuX -= rect['left'];
```
wallw-bits, line 1112: This shouldn't require bracket notation. If you were seeing issues in compile mode that is probably indicative of another issue.

kwilson-bits: I removed the bracket notation and it still works in compiled mode.

PR #944, opened 2020-04-18
add ability to add a list by regex

johnstacy:
Example:
`os.ui.list.add('action-list_.*', 'button-directive', null, true)`
then in html: `<div list list-id="action-list_{{someUniqueThing}}">`

When the list is instantiated, it will match the regex list and put a concrete version in the regular list map based on the items in the regex list.

### DIFF
--- a/src/os/ui/list.js
+++ b/src/os/ui/list.js
@@ -30,10 +30,10 @@ os.ui.list.ListEntry;
 
 
 /**
- * @type {Object<string, Array<os.ui.list.ListEntry>>}
+ * @type {Map<string|RegExp, Array<os.ui.list.ListEntry>>}
  * @private
  */
-os.ui.list.map_ = {};
+os.ui.list.map_ = new Map();
 
 
 /**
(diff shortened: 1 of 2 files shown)


src/os/ui/list.js
```diff
@@ -48,9 +55,10 @@ os.ui.list.dispatcher_ = new goog.events.EventTarget();
  * @param {string} id The list ID to which to add
  * @param {string} markup The directive or markup to add
  * @param {number=} opt_priority The sort priority (lowest to highest)
+ * @param {boolean=} opt_isRegex True if id is a regex to match against list IDs
  */
-os.ui.list.add = function(id, markup, opt_priority) {
-  var map = os.ui.list.map_;
+os.ui.list.add = function(id, markup, opt_priority, opt_isRegex) {
```
wallw-bits, line 60: To keep the API simpler, let's just keep a single storage structure and drop the `opt_isRegex` param everywhere. That means that `id` becomes becomes more of a pattern (or optionally `(string|RegExp)`). If you want to treat string input as an exact match, then consider converting it to a `RegExp` by making the pattern `'^' + id + '$'`.

Also, it looks like this is wholly lacking tests. Let's at least add tests for the new behavior and I'll see if I can help catch them up.

johnstacy: sounds good.

johnstacy: API now keeps the same number of args, just accepts a `RegExp` in addition to the `string`. The `RegExp` is stored as such and when a list initializes that matches that regex a new concrete version of that list is added to the map. Think of the RegExp map entry as a template that the concrete version is created from

PR #951, opened 2020-04-29
fix(geo): remove incorrect shortcut in parseLon and parseLat

wallw-bits:
Only do the Decimal Degrees shortcut if the number of digits is <= 3 for lon and <= 2 for lat.


wallw-bits:
Hmmm. This doesn't handle leading zeroes. I was trying really hard to avoid running a regex in the shortcut case. Might have to do that anyway.

I'll add another test for that.

wallw-bits:
I like this update better.

PR #955, opened 2020-05-21
fix(source): better multi-remove protection

wallw-bits:
Previously this would blow up if something attempted feature removal multiple times. Since it is cheap to check if it is been done since we have multiple indices by key/id, we'll do that before removing.

### DIFF
--- a/src/os/source/vectorsource.js
+++ b/src/os/source/vectorsource.js
@@ -1106,7 +1106,7 @@ os.source.Vector.updateScratchExtent_ = function(g) {
  * @suppress {accessControls}
  */
 os.source.Vector.prototype.updateIndex = function(feature) {
-  if (feature) {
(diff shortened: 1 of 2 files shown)


src/os/source/vectorsource.js
```diff
@@ -2158,17 +2159,18 @@ os.source.Vector.prototype.processImmediate = function(feature) {
  * @suppress {accessControls} To allow direct access to feature metadata.
  */
 os.source.Vector.prototype.columnTypeDetection_ = function(feature) {
-  var keys = feature.getKeys();
-  keys.forEach(function(col) {
-    var value = feature.values_[col];
-    var type = typeof (value);
+  const keys = feature.getKeys();
```
wallw-bits, line 2162: This was changed to avoid making a new function in a performance-critical loop.

PR #959, opened 2020-05-28
feat(renderer): add display settings hook up for multiple renderers

csgibson:
In the event that more than one webgl renderer is present: 
allow the associated plugin to register it with the mapContainer
allow users to choose one from the display settings

### DIFF
--- a/src/os/config/displaysettings.js
+++ b/src/os/config/displaysettings.js
@@ -221,6 +221,26 @@ os.config.DisplaySettingsCtrl = function($scope) {
    */
   this['maxFeatures3D'] = 150000;
 
+  /**
+   * Need a button to reset everything if the renderer changes
+   * @type {boolean}
+   */
+  this['resetButtonActive'] = false;
+
+  /**
+   * The available 3D renderers
+   * @type {Array<Object<string, string>>}
+   */
+  this['renderers3D'] = this.parseRenderers_();
+
+  /**
+   * The current 3D renderer
+   * @type {string|undefined}
+   */
+  this['renderer'] = os.settings.get(os.webgl.AbstractWebGLRenderer.ACTIVE_SETTINGS_KEY);
+
+  $scope.$watch('display.renderer', this.updateRenderer_.bind(this));
+
   $scope.$watch('display.fogEnabled', this.updateFog.bind(this));
   $scope.$watch('display.fogDensity', this.updateFog.bind(this));
 
@@ -235,6 +255,14 @@ os.config.DisplaySettingsCtrl = function($scope) {
 
   $scope.$on('$destroy', this.destroy_.bind(this));
 };
+
+
+/**
+ * Use this if something goes wrong with retrieving renderer ids
+ * @type {string}
+ * @const
+ */
+os.config.DisplaySettingsCtrl.DEFAULT_RENDERER_ID = 'unknown';
 
 
 /**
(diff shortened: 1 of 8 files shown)


src/os/config/displaysettings.js
```diff
@@ -431,6 +459,55 @@ os.config.DisplaySettingsCtrl.prototype.getZoom = function() {
...
+ * @private
+ */
+os.config.DisplaySettingsCtrl.prototype.parseRenderers_ = function() {
+  const retArr = [];
+  const renderers = os.MapContainer.getInstance().getWebGLRenderers();
+  if (renderers) {
+    for (const [key, val] of Object.entries(renderers)) {
+      const temp = {};
+      temp['id'] = key == val.id ? key : os.config.DisplaySettingsCtrl.DEFAULT_RENDERER_ID;
+      temp['label'] = temp['id'].charAt(0).toUpperCase() + temp['id'].substr(1);
```
schmidtk, line 474: While this works through convention, it may be better to allow the renderer to provide its own user-facing label along with the description added in this PR.

src/os/mapcontainer.js
```diff
@@ -285,11 +292,72 @@ os.MapContainer.prototype.getWebGLRenderer = function() {
...
+};
+
+
+/**
+ * Remove a WebGL renderer
+ *
+ * @param {os.webgl.IWebGLRenderer} renderer
+ * @return {boolean}
+ */
+os.MapContainer.prototype.removeWebGLRenderer = function(renderer) {
```
schmidtk, line 342: Is there a use case for this? Seems like this and `getWebGLRendererKey_` wouldn't typically be used (plugins will either add the renderer or not on init).

src/os/webgl/abstractwebglrenderer.js
```diff
@@ -18,6 +18,18 @@ goog.require('os.webgl.IWebGLRenderer');
 os.webgl.AbstractWebGLRenderer = function() {
   os.webgl.AbstractWebGLRenderer.base(this, 'constructor');
 
+  /**
+   * The id string of the renderer; probably the same as the plugin
+   * @type {string|undefined}
+   */
+  this.id;
```
schmidtk, line 25: This and `description` should probably be defined on `os.webgl.IWebGLRenderer`.

src/plugin/cesium/cesiumplugin.js
```diff
@@ -94,46 +94,50 @@ plugin.cesium.Plugin.prototype.init = function() {
...
-      plugin.cesium.tiles.Provider,
-      plugin.cesium.tiles.TYPE,
-      plugin.cesium.tiles.TYPE));
-  dm.registerDescriptorType(plugin.cesium.tiles.ID, plugin.cesium.tiles.Descriptor);
-
-  var im = os.ui.im.ImportManager.getInstance();
-  im.registerImportDetails(plugin.cesium.tiles.TYPE, true);
-  im.registerImportUI(plugin.cesium.tiles.mime.TYPE, new plugin.cesium.tiles.TilesetImportUI());
+  // check if cesium is the active renderer
+  if (os.settings.get(os.webgl.AbstractWebGLRenderer.ACTIVE_SETTINGS_KEY) == plugin.cesium.Plugin.ID) {
```
schmidtk, line 98: It might be better to have 3D plugins always register their renderer class with `MapContainer` (just the class, not an instance of it). When `MapContainer` initializes (after plugins have loaded), it will create an instance of the active renderer and call an init function on it. That init function will contain the remaining code in this block. Then plugins don't need to be aware of the internals of how the renderer is selected/activated.

src/os/webgl/abstractwebglrenderer.js
csgibson: I was concerned about a breaking change on the `IWebGLRenderer`. If I don't need to worry about that I can add all three properties and get rid of the random fallback for ids.

wallw-bits: If you're concerned about that (which I don't think I am), then you can just put it in another interface and have the class implement both interfaces.

src/plugin/cesium/cesiumplugin.js
csgibson: Moved the plugin code to the renderer as suggested but didn't do the renderer registration changes yet, still need to test it out first

csgibson: Looks like the MapContainer inits in `onAngularReady_` in `os.ui,map` and I am having trouble hooking this up without regressions within my available time. Let me know if you want me to figure this part out but it may have to wait until next month.

src/plugin/cesium/cesiumrenderer.js
```diff
@@ -108,6 +129,46 @@ plugin.cesium.CesiumRenderer.prototype.initialize = function() {
     return new goog.Promise(function(resolve, reject) {
       plugin.cesium.loadCesium().then(function() {
         try {
+          // register the default set of synchronizers
```
jsalankey, line 132: Moving all of this code to this function makes me slightly nervous. Is there ever a case where `initialize` gets called again in a single session? Calls like these just fit better in a plugin init call where we are pretty certain they won't happen again.

csgibson: The check at the top of this function is essentially a more robust version of `this.isInitialized()` so I think it will be fine. In a basic test, I wasn't able to get it to call twice.

PR #960, opened 2020-06-03
feat(branding): add hooks for opensphere to pickup brand from suite

gregroyal:
of a suite

Opensphere supports application level branding. This change makes it easier to rebrand if its part of a application suite


gregroyal:
I guess a note for anyone who extends opensphere, they'll also have to move their favicon location

index-template.html
```diff
@@ -4,6 +4,7 @@
 <meta charset="UTF-8">
 <!-- here in head area, or Chrome will look in site root before picking it up -->
 <link rel="icon" href="@version@images/favicon.ico?v=1" sizes="16x16 24x24 32x32 48x48 64x64" type="image/x-icon"/>
+<link rel="icon" href="@version@images/brand/favicon/opensphere.ico?v=1" sizes="16x16 24x24 32x32 48x48 64x64" type="image/x-icon"/>
```
gregroyal, line 7: allows choosing where you want your favicon, should work if with the original, or if you get from another plugin

PR #961, opened 2020-06-05
feat(heatmap): moved heatmap cesium sync to cesium

glennkitchellcaci:
Moved implementation of HeatmapSynchronizer to Cesium as a simple workaround to handle the heatmap canvas when swapping back and forth between WebGLRenderer plugins

Testing Steps:
1. Opensphere | http://red-curtain.surge.sh/opensphere/
2. Add a Data Layer and load some data
3. Right click the Feature layer > Click "Generate Heatmap"
4. Verify a heatmap is drawn
5. Zoom in fairly close to the area
6. Verify the heatmap becomes more detailed as you get closer
(No further testing needed here; the change was simply moving the Synchronizer; not changing its functionality)


src/plugin/heatmap/heatmaplayer.js
```diff
@@ -639,3 +639,12 @@ plugin.heatmap.Heatmap.drawShadow = function(context, sizeY, blur) {
...
 };
+
+
+/**
+ * @enum {string}
+ * @const
+ */
+plugin.heatmap.Heatmap.LayerValues = {
+  CANVAS: 'canvas'
+};
```
jsalankey, line 650: Rather than defining a whole new enum for this, you could tack it onto `plugin.heatmap.HeatmapField` instead. The properties on that are currently all keys for features, but a field for a layer doesn't hurt my head there.

PR #964, opened 2020-06-18
feat(export): option to export all,selected or active areas

arwatts16:


src/os/ui/areas.js
```diff
@@ -144,13 +147,25 @@ os.ui.AreasCtrl.prototype.exportDisabled = function() {
 };
 
 
+/**
+ * Default for the export button
+ *
+ * @export
+ */
+os.ui.AreasCtrl.prototype.defaultExport = function() {
+  os.ui.AreasCtrl.export({areaOpt: 'selected', scope: this.scope});
```
jsalankey, line 156: The `areaOpt` and `scope` keys here either need to be referenced with quotes everywhere or without quotes everywhere. There's a reference below that uses them with quotes that won't work in compiled mode with this object.

src/os/ui/areas.js
```diff
@@ -144,13 +147,25 @@ os.ui.AreasCtrl.prototype.exportDisabled = function() {
...
+};
+
+
 /**
  * Pop up area export gui
  *
+ * @param {Object} opt
```
jsalankey, line 163: In a similar vein to the above comment, we generally prefer to define a `typedef` rather than using generic options objects, just to make it clear what the type of each value is on here. If you look at `importmanager.js:16`, there's an example declaration of `os.ui.im.ImporterConfig`. All this does is tell the Closure compiler what type those values are.

src/os/ui/areas.js
```diff
@@ -144,13 +147,25 @@ os.ui.AreasCtrl.prototype.exportDisabled = function() {
...
  * @export
  */
-os.ui.AreasCtrl.prototype.export = function() {
-  var areas = /** @type {Array<os.structs.ITreeNode>} */ (this.scope['selected']).map(
+os.ui.AreasCtrl.export = function(opt) {
+  var areaOpt = /** @type {string} */ (opt['areaOpt'] ? opt['areaOpt'] : 'selected');
+  var areas = /** @type {Array<os.structs.ITreeNode>} */ (opt['scope'][areaOpt]).map(
```
jsalankey, line 168: These are the problematic bracket notations.

src/os/ui/areas.js
```diff
@@ -171,6 +186,40 @@ os.ui.AreasCtrl.prototype.export = function() {
...
+/**
+ * Opens the area export menu.
+ *
+ * @export
+ */
+os.ui.AreasCtrl.prototype.openExportMenu = function() {
+  var target = this.element.find('.js-export-group');
+  var menu = os.ui.menu.areaExport.MENU;
+
+  var menuOptions = menu.getRoot()['children'];
```
jsalankey, line 198: You should use `getChildren()` here rather than referencing `['children']`. This is generally true for all of our typed JS classes. Bracket notation should mostly be used for properties on Angular scopes/controllers that need to be referenced in HTML.

arwatts16: There isn't a function that exists right now on `os.ui.menu.Menu` for `getChildren()`, should I make one/

src/os/ui/menu/areaexportmenu.js
```diff
@@ -0,0 +1,99 @@
+goog.provide('os.ui.menu.areaExport');
```
jsalankey, line 1: One other structural change I would make would be to move all of this menu code into the `areas.js` file. The global, singleton menu design (i.e. instantiated in `MainCtrl` is only necessary for menus that are associated with UIs that are always in the view. Stuff like `MapMenu` and `SpatialMenu` for map interactions, and the menus on the top nav bar fall into this category. This one should be created and destroyed with the `areas.js` class.

src/os/query/areamanager.js
```diff
@@ -191,6 +192,26 @@ os.query.AreaManager.mapOriginalGeoms = function(area) {
 };
 
 
+/**
+ * @return {Array<!os.data.AreaNode>} Array of active areas
+ * @export
```
jsalankey, line 197: The `@export` tag here is unnecessary. That tag causes the compiler to leave the function name the same in compiled mode, which we only need to do for functions referenced by our HTML (and in some other, specialized cases).

I would also recommend moving this function entirely into the areas.js file. The `AreaManager` shouldn't create `AreaNodes` as those are more of a UI construct while this is the model class.

src/os/query/areamanager.js
```diff
@@ -191,6 +192,26 @@ os.query.AreaManager.mapOriginalGeoms = function(area) {
...
+ */
+os.query.AreaManager.prototype.activeAreas = function() {
+  var allAreas = this.getAll();
+  var activeAreas = [];
+  var index = 0;
+
+  allAreas.forEach((area) => {
+    if (area.get('shown')) {
+      activeAreas[index] = /** @type {!os.data.AreaNode} */ (new os.data.AreaNode(area));
```
jsalankey, line 206: You should call `activeAreas.push(area)` here and get rid of the `index` variable altogether.

For future uses of the `forEach` call, it's worth checking the larger function signature for it and other built in JS API calls: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

If you do need the index when using this call at some point in the future, `forEach` passes it to the function you give it as the second argument i.e. `allAreas.forEach((ele, index, arr) => {...}`.

src/os/ui/areas.js
```diff
@@ -167,7 +183,106 @@ os.ui.AreasCtrl.prototype.export = function() {
...
+      label: 'Export Selected',
+      eventType: os.ui.AreasCtrl.EventType.SELECTED,
+      tooltip: 'Export only selected areas',
+      handler: os.ui.AreasCtrl.exportEventHandler_,
+      sort: 2
+    }, {
+      label: 'Export Active',
+      eventType: os.ui.AreasCtrl.EventType.ACTIVE,
+      tooltip: 'Export only active areas',
+      handler: os.ui.AreasCtrl.exportEventHandler_,
```
jsalankey, line 226: You can get rid of the menu context object entirely if you switch `os.ui.AreasCtrl.exportEventHandler_` to be on the prototype and use `this.exportEventHandlier_.bind(this)`. This is a cleaner way than passing around that context object and will allow you to directly reference `this.scope['areas/activeAreas/selected']` from the handler function.

src/os/ui/areas.js
```diff
@@ -167,7 +183,106 @@ os.ui.AreasCtrl.prototype.export = function() {
...
+  var menuRoot = menu.getRoot();
+
+  this.scope['activeAreas'] = this.scope['areas'][0].getLabel() == 'No results' ? [] : this.activeAreas();
+
+  menuRoot.find(os.ui.AreasCtrl.EventType.ALL).enabled = (this.scope['areas'][0].getLabel() != 'No results');
+  menuRoot.find(os.ui.AreasCtrl.EventType.SELECTED).enabled = (this.scope['selected'].length > 0);
+  menuRoot.find(os.ui.AreasCtrl.EventType.ACTIVE).enabled = (this.scope['activeAreas'].length > 0);
+
+  if (menu && target && target.length) {
+    menu.open({all: this.scope['areas'], active: this.scope['activeAreas'], selected: this.scope['selected']}, {
```
jsalankey, line 240: The lack of quotes here on the object keys is an example where the compiler might mess with you. This goes away with my above fix suggestion.

PR #966, opened 2020-06-22
Feat wrapping boxes

jsalankey:
Adds support for boxes that wrap the earth in both 2D and 3D mode. This updates the drawing control to allow it, and also fixes some issues with queries/geometries that cross the antimeridian. A future PR will update the Enter Coordinates dialog.

### DIFF
--- a/src/os/ui/ol/interaction/dragboxinteraction.js
+++ b/src/os/ui/ol/interaction/dragboxinteraction.js
@@ -196,5 +196,9 @@ os.ui.ol.interaction.DragBox.prototype.cleanup = function() {
  * @inheritDoc
  */
 os.ui.ol.interaction.DragBox.prototype.getResultString = function() {
-  return 'TODO: FIX';
+  if (this.startCoord && this.endCoord) {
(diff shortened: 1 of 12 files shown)


jsalankey:
Added a commit to this to fix the Enter Coordinates dialog as well.

src/os/ui/ol/interaction/dragboxinteraction.js
```diff
@@ -156,5 +196,5 @@ os.ui.ol.interaction.DragBox.prototype.cleanup = function() {
  * @inheritDoc
  */
 os.ui.ol.interaction.DragBox.prototype.getResultString = function() {
-  return this.extent ? this.extent.toString() : 'none';
+  return 'TODO: FIX';
```
glennkitchellcaci, line 199: Does this need to be To-Done as part of this ticket and/or is there a follow-on ticket?

jsalankey: Yep, forgot about this. It's actually only called to log the result, but I still need to update it.

PR #968, opened 2020-06-24
improvement(u-link-style): converted u-link-style tags to btn btn-link

Gamerlios123:
Converted div/span/a tags that used "u-link-style" to button tags that use "btn btn-link".

### DIFF
--- a/scss/os/_mixins.scss
+++ b/scss/os/_mixins.scss
@@ -61,7 +61,7 @@
     @content;
   }
 
-  .u-bg-#{$color}-text-a .u-link-style,
+  .u-bg-#{$color}-text-a .btn.btn-link,
   .u-bg-#{$color}-text-a a {
     @content;
   }
--- a/scss/standard/_bootswatch.scss
+++ b/scss/standard/_bootswatch.scss
@@ -20,3 +20,10 @@
 .pagination {
   margin: .5rem;
 }
+
+.btn.btn-link {
+  border: 0;
+  font-size: inherit;
+  padding: 0;
+  vertical-align: baseline;
+}
--- a/scss/os/_utils.scss
+++ b/scss/os/_utils.scss
@@ -132,22 +132,6 @@ $user-select-options: all, auto, text, none;
   text-align: center;
 }
 
-.u-link-style {
-  @include hover {
-    color: $link-hover-color;
-    text-decoration: $link-hover-decoration;
-  }
-
-  background-color: transparent; // Remove the gray background on active links in IE 10.
-  color: $link-color;
-  cursor: pointer;
-  text-decoration: $link-decoration;
-
-  &.nav-link {
-    text-decoration: none;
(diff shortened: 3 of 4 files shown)


scss/standard/_bootswatch.scss
```diff
@@ -20,3 +20,7 @@
 .pagination {
   margin: .5rem;
 }
+
+.btn.btn-link {
+  padding: 0 0 0 0;
```
gregroyal, line 25: padding: 0 should achieve the same thing

gregroyal: actually the build already breaks
Property `padding` should be written more concisely as `0` instead of `0 0 0 0`  shorthand-values

Gamerlios123: Ah, didn't see that build failure for some reason. It's been fixed.

gregroyal: checking out the review, might make sense to also add vertical-align: baseline? since btn sets it to middle, this undoes that and makes it normal text. Do you think that works to apply it everywhere?

scss/standard/_bootswatch.scss
```diff
@@ -20,3 +20,8 @@
 .pagination {
   margin: .5rem;
 }
+
```
gregroyal, line 23: Remove u-link-style also

scss/os/_mixins.scss
```diff
@@ -61,7 +61,6 @@
     @content;
   }
 
-  .u-bg-#{$color}-text-a .u-link-style,
```
gregroyal, line 64: this probably should be replaced with .btn.btn-link
I think the purpose is this colors the link to a color that doesnt clash with the background color (say if the background is blue, and the link is blue)

scss/standard/_bootswatch.scss
```diff
@@ -20,3 +20,8 @@
...
+.btn.btn-link {
+  padding: 0;
+  vertical-align: baseline;
```
gregroyal, line 26: I'd experiment with this, but you might want to ignore the font that .btn adds by doing

font-size: inherit;

Might fix any issues with btn-links in header tags and stuff

gregroyal:
```
.btn.btn-link {
  font-size: inherit;
  padding: 0;
  vertical-align: baseline;

  &:focus,
  &:hover {
    border: 0;
  }
}
```

Gamerlios123: Adding `font-size: inherit;` did allow me to simplify my changes to the slide widgets in bits so that has been added.

For the border, why do we only want it to be 0 on hover and focus and not all the time?

gregroyal: I was assuming bootstrap adds it on hover and focus. so if you dont override it there it might not work? its already not there normally

PR #969, opened 2020-06-25
refactor(theming): cleaned up fonts and allowed subprojects to provid…

gregroyal:
…e fonts


gregroyal:
opened under  a different branch https://github.com/ngageoint/opensphere/pull/972/files

PR #971, opened 2020-06-26
feat(alerts): throttle duplicated alerts

rothm-caci:
UUOM-1063


schmidtk:
Could we mimic the Dev Tools Console behavior and display a count next to repeated messages? That would inform the user that the message occurred multiple times without spamming them.
DevToolsCounter

gregroyal:
I like that idea! obviously more scope creep. And also I think this specific case @rothm-caci was addressing theres 2 alerts that are both valid and exposes a bug in our tool lol. We have a ticket for that bug, but adding a counter would also expose it

schmidtk:
@rothm-caci @gregroyal FYI this broke the `sendAlertOnce` call, which isn't setting up the throttle map. This results in an undefined exception in `sendAlert_`.

PR #975, opened 2020-07-08
feat(featureinfo): toggle hide empty columns in feature info properties

schmidtk:
Features with sparsely populated columns do not present well in the Feature Info dialog, if the user does not care about the empty columns. This is common with layers whose metadata fields vary by feature type, like OSM.

This PR adds a toggle button to show/hide empty columns in Feature Info. This preference defaults to showing empty (current behavior) and is saved to settings so the value persists through app refresh.

I also did some cleanup/rework on the Feature Info properties tab:

- Table header is now fixed so scrolling only impacts data rows. This makes it easy to toggle sort and the new show/hide even when scrolled.
- Changing the displayed properties for any reason (new feature, toggling show/hide empty, etc) will now apply the sort so properties are still displayed in the order indicated by the headers.
- Fixed the first column width to 40% for a consistent layout. Previously width was determined by the browser using column values.
- The Field column now truncates text exceeding the column width, and displays the full value on hover. The Value column continues to wrap so the full value can be seen.

Demo: https://unwieldy-sink.surge.sh/
Test CSV: [Empty Columns.txt](https://github.com/ngageoint/opensphere/files/4891800/Empty.Columns.txt)

### DIFF
--- a/scss/opensphere.scss
+++ b/scss/opensphere.scss
@@ -40,6 +40,7 @@
 @import 'os/overrides_select2';
 @import 'os/overrides_slickgrid';
 @import 'os/overrides_tuieditor';
+@import 'os/propertiestab';
 @import 'os/searchresults';
 @import 'os/singlepage';
 @import 'os/slicktree';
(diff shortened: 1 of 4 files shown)


schmidtk:
I added a status bar to indicate how many fields are displayed in the table, and how many are hidden when "hide empty" is enabled. This is similar to the status bar in the feature list and intended to help inform the user when the table isn't displaying everything. I didn't want to take up extra vertical space with these changes, but it seemed minimal and consistent with the feature list.

PR #977, opened 2020-07-08
feat(reset): update Reset button nameA

ericadeckl:
UCEP-437


views/config/settingswindow.html
```diff
@@ -11,7 +11,7 @@
     <div class="d-flex">
       <button type="button" class="btn btn-danger" ng-click="setCon.reset();">
         <i class="fa fa-fw fa-refresh"></i>
-        Reset
+        Reset {{$root.appSuite.getName('suite')}} Settings
```
schmidtk, line 14: `$root.appSuite` is not something that gets set in OpenSphere. `os.config.getAppName` should be used to get the application name.

PR #979, opened 2020-07-09
chore(storage): show help link in persistent storage warnings

swtyler:
Added config to show a link in the persistent storage alert pop-ups.

### DIFF
--- a/src/plugin/storage/persistplugin.js
+++ b/src/plugin/storage/persistplugin.js
@@ -33,6 +33,7 @@ class PersistPlugin extends AbstractPlugin {
   init() {
     const msgs = [];
     const metrics = Metrics.getInstance();
+    const helpUrl = os.settings.get('plugin.storage.persistenceHelp', 'https://web.dev/persistent-storage/');
 
     if (navigator.storage && navigator.storage.persist) {
       navigator.storage.persisted().then((result) => {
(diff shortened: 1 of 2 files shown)


src/plugin/storage/persistplugin.js
```diff
@@ -97,13 +100,15 @@ class PersistPlugin extends AbstractPlugin {
...
-        msgs.push(`The application is not being served in a secure context. Persistent storage permission
-            will be automatically declined.`);
+        msg = `This application is not being served in a secure context. The persistent storage permission
+            will be automatically declined. `;
       }
 
-      msgs.push('Your application settings may be automatically reset when under storage pressure.');
-      this.log(msgs);
+      msg += `Your application settings may be automatically reset when under storage pressure. Get more info 
+        <a href="` + helpUrl + `"rel="noopener" target="_blank">here</a>`;
```
eric-crowell, line 110: String concatenation isn't necessary for [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). It's a bit more readable to use it like so:
```javascript
msg += `Your application settings may be automatically reset when under storage pressure. Get more info 
        <a href="${helpUrl}" rel="noopener" target="_blank">here</a>`;
```
If OpenSphere doesn't care about it, then either way is okay. Just a bit strange to look at to me.
Same for the other instances in the changes.

schmidtk: Agree, if you're using template literals you should use expressions instead of string concatenation.

swtyler: I made that change

PR #980, opened 2020-07-10
feat(feature actions): indicates active feature actions on a layer

arwatts16:


src/os/data/layernode.js
```diff
@@ -42,6 +42,7 @@ os.data.LayerNode = function() {
   this.layer_ = null;
 
   os.ui.queryManager.listen(goog.events.EventType.PROPERTYCHANGE, this.onNodeChanged_, false, this);
+  os.dispatcher.listen(os.im.action.ImportActionEventType.REFRESH, this.onNodeChanged_, false, this);
```
jsalankey, line 45: This can simply listen on the `ImportActionManager`. Most of the manager classes in OS extend `goog.events.EventTarget`. That base class is where all these classes inherit the `listen/unlisten/dispatchEvent` methods, and we prefer to keep listeners off of `os.dispatcher` as much as possible.

Also, it needs to unlisten in the `disposeInternal` method below, otherwise the `LayerNode` instances will be retained in memory because the listener reference still exists.

src/os/im/action/importactionmanager.js
```diff
@@ -530,6 +530,7 @@ os.im.action.ImportActionManager.prototype.initialize = goog.nullFunction;
  */
 os.im.action.ImportActionManager.prototype.apply = function() {
   this.dispatchEvent(os.im.action.ImportActionEventType.REFRESH);
+  os.dispatcher.dispatchEvent(os.im.action.ImportActionEventType.REFRESH);
```
jsalankey, line 533: This can be removed entirely in favor of simply listening for the event above.

src/plugin/featureaction/featureactionmanager.js
```diff
@@ -168,6 +168,25 @@ plugin.im.action.feature.Manager.prototype.onSourceRemoved_ = function(event) {
...
+plugin.im.action.feature.Manager.prototype.hasActiveActions = function(source) {
+  var featureActions = this.getActionEntries(source.getId());
+  var hasActiveActions = false;
+
+  if (featureActions.length > 0) {
+    for (var i = 0; i < featureActions.length; i++) {
+      if (featureActions[i].enabled) {
+        hasActiveActions = true;
+      }
+    }
```
jsalankey, line 184: This loop can be more cleanly written using `featureActions.some()`. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

src/os/layer/vector.js
```diff
@@ -428,6 +428,10 @@ os.layer.Vector.prototype.getFASet = function() {
     icons.push(os.ui.Icons.FILTER);
   }
 
+  if (plugin.im.action.feature.Manager.getInstance().hasActiveActions(/** @type {os.source.Request} */ (source))) {
```
jsalankey, line 431: The `plugin` namespace should never be referenced outside of other `plugin` files. In this instance, this should instead call `os.im.action.ImportActionManager.getInstance()` as the feature action plugin version extends and replaces that singleton.

I would also recommend moving the `hasActiveActions` function to the `ImportActionManager` as it's a handy utility to have on the core singleton.

src/os/im/action/importactionmanager.js
```diff
@@ -610,3 +610,17 @@ os.im.action.ImportActionManager.prototype.onAddActionEntry_ = function(event) {
...
+ * @param {os.source.Request} source
+ * @return {boolean}
+ */
+os.im.action.ImportActionManager.prototype.hasActiveActions = function(source) {
+  var featureActions = this.getActionEntries(source.getId());
+
+  const hasActiveActions = (action) => action.enabled === true;
+
+  return featureActions.some(hasActiveActions);
+};
```
jsalankey, line 626: Couple things here:
1. This function should take just an ID as an argument, not the whole source. This class generally stores things by ID and all off the function signatures for getters work that way. Also, just better to make an argument to a function like this the simplest possible thing that it can be.

2. Rather than referencing `action.enabled`, we prefer to use the getter whenever possible (in this case, `action.isEnabled()`. The flag is actually defined with bracket notation on `FilterEntry`, so referencing it with a . is potentially a problem when compiled.

3. It's simpler to just run the function as the argument to `some` rather than assigning it to a variable, i.e. 
`featureActions.some((action) => action.isEnabled());`

PR #983, opened 2020-07-15
fix(cesium): fixes case where port was being passed as NaN

jsalankey:


src/plugin/cesium/mixin/cesiummixin.js
```diff
@@ -8,6 +8,31 @@ goog.require('os.MapEvent');
 goog.require('os.net.Request');
 
 
+/**
+ * Gets the port from a URL. Cesium requires passing a valid port to its TrustedServers API.
+ * Returns 443 for https, 80 for http and everything else that isn't specified.
+ * @param {URL} url The URL to get the port from.
+ * @return {number} The port.
+ */
+const getPort = (url) => {
```
schmidtk, line 17: Only `goog.module` files should declare things at the file scope. Outside of a module, you're declaring `window.getPort` to be this function.

src/plugin/cesium/mixin/cesiummixin.js
```diff
@@ -8,6 +8,31 @@ goog.require('os.MapEvent');
...
+  let port;
+
+  if (url.port === '') {
+    // if it's not specified, use 443 for https or 80 else
+    port = url.protocol === 'https:' ? 443 : 80;
```
schmidtk, line 22: This would be a good general fallback, so you may want to try parsing the port and if `port` is still NaN/undefined then do this.

PR #986, opened 2020-07-21
fix(settings): no need to JSON stringify settings

swtyler:
Settings cleanup, there is no needs to JSON-ify settings to save them, that should be done internal to each storage mechanism if necessary. I verified that state files save and load correctly.


michael-roschenwimmer:
Looks good to me!

PR #987, opened 2020-07-23
fix(feature): clean up source listener in properties tab

schmidtk:
The properties tab doesn't clean up its source listener, causing the controller to be leaked and an error to throw if the source columns change.


src/os/ui/feature/tab/propertiestab.js
```diff
@@ -44,6 +44,8 @@ os.ui.Module.directive('propertiestab', [os.ui.feature.tab.propertiesTabDirectiv
  * @ngInject
  */
 os.ui.feature.tab.PropertiesTabCtrl = function($scope, $element) {
+  os.ui.feature.tab.PropertiesTabCtrl.base(this, 'constructor', $scope, $element);
```
schmidtk, line 47: I didn't see a need for this to be called last, and when converted to an ES6 class you must call `super` before referencing `this`.

PR #993, opened 2020-07-27
feat: fix z-index for initializing components on a window

johnstacy:


src/os/ui/query/areafilteradd.js
```diff
@@ -591,8 +591,9 @@ os.ui.query.AreaFilterAddCtrl.prototype.onAreasSelected_ = function(features) {
  * @param {string} label
  * @param {boolean=} opt_hideArea
  * @param {boolean=} opt_hideFilter
+ * @param {boolean=} opt_modal
  */
-os.ui.query.AreaFilterAddCtrl.launch = function(label, opt_hideArea, opt_hideFilter) {
+os.ui.query.AreaFilterAddCtrl.launch = function(label, opt_hideArea, opt_hideFilter, opt_modal) {
```
gregroyal, line 596: This isnt needed anymore right?

src/os/ui/window.js
```diff
@@ -898,7 +898,17 @@ os.ui.WindowCtrl.prototype.disposeInternal = function() {
...
+      if (key == '#' + this.scope['id']) {
+        return false;
+      }
+
+      const win = os.ui.window.getById(key);
+      const scope = win.children().first().scope();
+      return scope && scope['modal'];
+    });
+
+    this.removeModalBg(moreModalsLeft);
```
gregroyal, line 911: shouldn't this only remove the modal background if there are no modals left? or are there multiple backgrounds applied?

src/os/ui/query/areafilteradd.js
johnstacy: it is still needed. it's a window, not a menu.

src/os/ui/window.js
johnstacy: a background is applied for each modal.

src/os/ui/query/areafilteradd.js
gregroyal: oh since the window you are opening is already a modal, you need this to optionally be a modal so it can open on top of it.

johnstacy: correct.

PR #997, opened 2020-08-03
test(cypress): selector update

brant-bits:


justin-bits:
Can you update the selectors for the other tabs on the Layers widget too?

PR #998, opened 2020-08-04
feat(layer-grouping): adds setters to IGroupable interface

jsalankey:
Adds the ability to pass a custom layer group ID/label to a tile layer via its layer config. This change can be adapted in the future to other layer types if we need to add custom grouping functionality to them.


jsalankey:
I screwed up and pushed this branch to ngageoint/feat-layer-grouping, which I think is what created that second build that is now failing. I deleted that branch, but the build didn't go away, but it always won't pass because the branch is gone.

PR #999, opened 2020-08-04
feat(areas): Created generics for handling any OGC FeatureLayer

glennkitchellcaci:
Created a Registry and the base classes for dynamically configuring and using any number of OGC Feature Layers. Default UI is a pop-up dialog with a dropdown and look-ahead text filter.

Testing Steps:
This cannot be tested with OS default configs.

### DIFF
--- a/externs/plugin.externs.js
+++ b/externs/plugin.externs.js
@@ -1,0 +1,42 @@
+/**
+ * @fileoverview Closure compiler externs for plugins. Use this file to define types that need to be left uncompiled to
+ *               avoid bracket notation and enforce type checking.
+ *
+ * @externs
+ */
+
+
+/**
+ * @type {Object}
+ */
+var pluginx;
+
+
+/**
+ * Namespace.
+ * @type {Object}
+ */
+pluginx.areadata;
+
+
+/**
+ * @typedef {{
+ *  label: (string|undefined),
+ *  text: (string|undefined),
+ *  icon: (string|undefined)
+ * }}
+ */
+pluginx.areadata.ListUIOptions;
+
+
+/**
+ * @typedef {{
+ *  id: (!string),
+ *  type: (plugin.areadata.AreaImportType|undefined),
+ *  clazz: (string|undefined),
+ *  menuItemOptions: (os.ui.menu.MenuItemOptions|undefined),
+ *  listUIOptions: (pluginx.areadata.ListUIOptions|undefined),
+ *  ogcSettings: (osx.ogc.OGCSettings|undefined)
+ * }}
+ */
+pluginx.areadata.AreaMenuItemOptions;
--- a/externs/plugin.externs.js
+++ b/externs/plugin.externs.js
@@ -33,7 +33,9 @@ pluginx.areadata.ListUIOptions;
  * @typedef {{
  *  id: (!string),
  *  type: (plugin.areadata.AreaImportType|undefined),
- *  clazz: (string|undefined),
+ *  clazzKey: (string|undefined),
+ *  sort: (number|undefined),
+ *  enabled: (boolean|undefined),
  *  menuItemOptions: (os.ui.menu.MenuItemOptions|undefined),
  *  listUIOptions: (pluginx.areadata.ListUIOptions|undefined),
  *  ogcSettings: (osx.ogc.OGCSettings|undefined)
--- a/externs/plugin.externs.js
+++ b/externs/plugin.externs.js
@@ -21,23 +21,13 @@ pluginx.areadata;
 
 /**
  * @typedef {{
- *  label: (string|undefined),
- *  text: (string|undefined),
- *  icon: (string|undefined)
- * }}
- */
-pluginx.areadata.ListUIOptions;
-
-
-/**
- * @typedef {{
  *  id: (!string),
- *  type: (plugin.areadata.AreaImportType|undefined),
+ *  type: (string|undefined),
  *  clazzKey: (string|undefined),
  *  sort: (number|undefined),
  *  enabled: (boolean|undefined),
- *  menuItemOptions: (os.ui.menu.MenuItemOptions|undefined),
- *  listUIOptions: (pluginx.areadata.ListUIOptions|undefined),
+ *  menuItemOptions: (Object<string, *>|undefined),
+ *  listUIOptions: (Object<string, *>|undefined),
  *  ogcSettings: (osx.ogc.OGCSettings|undefined)
  * }}
  */
--- a/src/os/data/registry.js
+++ b/src/os/data/registry.js
@@ -2,7 +2,8 @@ goog.module('os.data.Registry');
 
 const EventTarget = goog.require('goog.events.EventTarget');
 const PropertyChangeEvent = goog.require('os.events.PropertyChangeEvent');
-
+const {PROPERTYCHANGE} = goog.require('goog.events.EventType');
+const RegistryPropertyChange = goog.require('os.data.RegistryPropertyChange');
 
 /**
  * Object wrapper to serve as agnostic manager of objects/arrays/etc.
@@ -38,7 +39,7 @@ class Registry extends EventTarget {
    * @inheritDoc
    */
   dispose() {
-    this.removeAllListeners('propertychange');
+    this.removeAllListeners(PROPERTYCHANGE);
 
     super.dispose();
   }
@@ -134,13 +135,13 @@ class Registry extends EventTarget {
     });
 
     // then listen for changes to and run listeners as needed
-    this.listen('propertychange', function(event) {
-      switch (event.property_) {
-        case 'add':
-        case 'change':
+    this.listen(PROPERTYCHANGE, function(event) {
+      switch (event.getProperty()) {
+        case RegistryPropertyChange.ADD:
+        case RegistryPropertyChange.UPDATE:
           onAddChange.call(opt_this, event.newVal_);
           break;
-        case 'remove':
+        case RegistryPropertyChange.REMOVE:
           onRemove.call(opt_this, event.newVal_);
           break;
         default:
@@ -161,9 +162,9 @@ class Registry extends EventTarget {
     const had = this.has(key);
     this.map_[key] = [value, opt];
     if (!had) {
-      this.dispatchEvent(new PropertyChangeEvent('add', key));
+      this.dispatchEvent(new PropertyChangeEvent(RegistryPropertyChange.ADD, key));
     } else {
-      this.dispatchEvent(new PropertyChangeEvent('update', key));
+      this.dispatchEvent(new PropertyChangeEvent(RegistryPropertyChange.UPDATE, key));
     }
     return had;
   }
@@ -175,7 +176,7 @@ class Registry extends EventTarget {
    */
   remove(key) {
     if (this.has(key)) {
-      this.dispatchEvent(new PropertyChangeEvent('remove', key));
+      this.dispatchEvent(new PropertyChangeEvent(RegistryPropertyChange.REMOVE, key));
       delete this.map_[key];
       return true;
     }
--- a/src/os/data/registry.js
+++ b/src/os/data/registry.js
@@ -1,0 +1,185 @@
+goog.module('os.data.Registry');
+
+const EventTarget = goog.require('goog.events.EventTarget');
+const PropertyChangeEvent = goog.require('os.events.PropertyChangeEvent');
+
+
+/**
+ * Object wrapper to serve as agnostic manager of objects/arrays/etc.
+ *
+ * Typically, use os.data.CollectionManager instead if your "value" has an "id" property
+ *
+ * The key difference here is that a Registry lets you work with key-value pairs like normal; but
+ * also supports custom collections made from the value AND an array of optional parameters
+ *
+ * e.g.
+ *   myRegistry.register(key, value, option1, option2, ...);
+ * ... which handles additional parameters, versus...
+ *   myManager.add({id: key, data: [value, option1, option2, ...])
+ * ... which requires intermediary typedefs, special comparators, and extra logic to extract the "value" from
+ *     the typedef
+ *
+ * @template T
+ */
+class Registry extends EventTarget {
+  /**
+   */
+  constructor() {
+    super();
+
+    /**
+     * Collection of registered objects
+     * @type {!Object<string, T>}
+     */
+    this.map_ = {};
+  }
+
+  /**
+   * @inheritDoc
+   */
+  dispose() {
+    this.removeAllListeners('propertychange');
+
+    super.dispose();
+  }
+
+  /**
+   * Get the entry from this Registry as an array, just as it was registered
+   *
+   * e.g.
+   * myRegistry.register(k, v, opt1, opt2, ...);
+   * ...
+   * myRegistry.entry(v) >> [k, v, opt1, opt2, ...]
+   *
+   * @param {string} keyOrValue
+   * @return {Array<?>}
+   */
+  entry(keyOrValue) {
+    const key = this.key(keyOrValue);
+    return key ? [key, this.map_[key]].flat(2) : null;
+  }
+
+  /**
+   * Get the entries from this Registry as an array
+   * @return {Array<Array<?>>}
+   */
+  entries() {
+    const arr = [];
+    for (const entry of Object.entries(this.map_)) {
+      arr.push(entry.flat(2));
+    }
+    return arr;
+  }
+
+  /**
+   * Get the Value from this Registry
+   * @param {string} key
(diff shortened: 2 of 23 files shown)


src/os/data/registry.js
```diff
@@ -0,0 +1,86 @@
+goog.module('os.data.Registry');
+
+
+/**
+ * Object wrapper to serve as agnostic manager of objects
+ * @template T
+ */
+class Registry {
```
jsalankey, line 8: This class is similar to https://github.com/ngageoint/opensphere/blob/master/src/os/data/collectionmanager.js. You may even be able to replace it entirely and simply use the existing one.

src/os/query/query.js
```diff
@@ -7,6 +7,8 @@ goog.require('ol.proj');
 goog.require('os.interpolate');
 goog.require('os.metrics.MapMetrics');
 goog.require('os.metrics.Metrics');
+// goog.require('os.ogc.Service');
+// goog.require('os.ogc.registry');
```
jsalankey, line 11: Commented requires. Are these needed here?

src/os/ui/draw/drawpicker.js
```diff
@@ -222,15 +224,16 @@ os.ui.draw.DrawPickerCtrl.prototype.initControlMenu = function() {
...
-      label: 'Country Border',
-      eventType: os.ui.menu.draw.EventType.COUNTRY_BORDER,
-      tooltip: 'Choose a country border',
-      icons: ['<i class="fa fa-fw fa-globe"></i> '],
-      handler: onDraw,
-      sort: 130
-    });
+  // add any configured OGC lookups (e.g. Country Borders)
+  var sort = 130;
+  for (const [, service, settings] of os.ogc.registry.getInstance().entries()) {
```
jsalankey, line 229: Slight performance thing: for a loop like this, the to `getInstance().entries()` is happening on every iteration of the loop. In this case, it's not likely to matter, but for cases where that call has to generate, say, a 10000 element array every time, it's massively inefficient. Better to assign `const entries = os.ogc.registry.getInstance().entries()` and use the preassigned variable.

src/os/ui/draw/drawpicker.js
```diff
@@ -441,3 +447,17 @@ os.ui.draw.DrawPickerCtrl.prototype.onCountryChosen = function(garbage, feature)
     this.scope['callback'](geometry);
   }
 };
+
+
+/**
+ * Handler for water chosen.
+ *
+ * @param {ol.Feature} garbage The chosen water.
```
jsalankey, line 455: Odd variable name...

src/os/ui/draw/basedrawcontrols.js
```diff
@@ -290,15 +294,18 @@ os.ui.draw.BaseDrawControlsCtrl.prototype.initControlMenu = function() {
...
-      icons: ['<i class="fa fa-fw fa-globe"></i> '],
-      handler: os.ui.menu.draw.handleDrawEvent,
-      sort: 130
-    });
+
+  if (!this['hideExtraControls']) {
+    var sort = 130;
+    for (const [, service, settings] of os.ogc.registry.getInstance().entries()) {
+      // must be configured in settings.json and have MenuItemOptions (i.e. settings)
+      if (service.isConfigured() && settings) {
```
jsalankey, line 302: Here and everywhere else that a menu is being registered from the registry should be refactored a bit.

The way I envision it, these menu items should be thought of as being injected by a plugin that represents the potential providers of external areas, and therefore references to country borders or bodies of water should never appear here or anywhere else besides the classes that register them to the registry.

Part of what they register should be their menu handler functions, so that those can simply be set as the `handler` property on the menu items here.

src/os/ui/draw/basedrawcontrols.js
```diff
@@ -161,6 +164,7 @@ os.ui.draw.BaseDrawControlsCtrl.prototype.disposeInternal = function() {
   os.dispatcher.unlisten(os.ui.draw.DrawEventType.DRAWPOLYGON, this.onDrawType, false, this);
   os.dispatcher.unlisten(os.ui.draw.DrawEventType.DRAWLINE, this.onDrawType, false, this);
   os.dispatcher.unlisten(os.ui.draw.DrawEventType.DRAWCOUNTRY, this.onDrawType, false, this);
+  os.dispatcher.unlisten(os.ui.draw.DrawEventType.DRAWWATER, this.onDrawType, false, this);
```
jsalankey, line 167: As part of the refactor I mention below, things like these enumerated event types should not be referenced outside of the classes that create them.

src/os/ogc/query/ogcquery.js
```diff
@@ -0,0 +1,126 @@
...
+   * @private
+   */
+  onChoice_(callback, datum) {
+    if (datum && datum.length > 0) {
+      const service = registry.getInstance().get(this.type);
+      if (service) {
+        // inject the requested area into the callback so it has some additional context
+        service
+            .get(datum[0])
+            .then(goog.partial(callback, datum));
```
jsalankey, line 43: Nit: `goog.partial`'s functionality is replicated by calling `fn.bind(undefined, args)`.

src/os/ogc/ogcservice.js
```diff
@@ -0,0 +1,399 @@
...
+
+      var request = new Request(uri);
+      request.setHeader('Accept', 'application/xml');
+      request.setValidator(os.ogc.getException);
+      request.setDataFormatter(new WFSFormatter());
+
+      request.listenOnce(goog.net.EventType.SUCCESS, goog.partial(this.onAllLoaded_, resolve, reject), false, this);
+      request.listenOnce(goog.net.EventType.ERROR, function(event) {
+        this.onError(event, 'Failed to load data list.', false, reject);
+      }.bind(this), false, this);
```
jsalankey, line 241: Style thing: using `request.listen` is generally more verbose than using `request.getPromise()`. The promise version packs everything into a single call, doesn't require cleaning up listeners, and gets you directly to your result in the resolver function.

The only practical reason to use the `listen` version of a request anymore is if you need the request object again on the other side. Things like getting the response headers/response code are the cases where this is needed.

src/os/ogc/ogcservice.js
```diff
@@ -0,0 +1,399 @@
...
+      while (parser.hasNext()) {
+        var feature = parser.parseNext();
+        if (feature) {
+          list.push(feature);
+        }
+      }
+    }
+
+    if (list && list.length > 0) {
+      list.sort(goog.partial(osOlFeature.fieldSort, this.nameProperty_));
```
jsalankey, line 276: `goog.partial` calls can be replaced with `fn.bind(undefined, args)`. We'll replace them all some day,  but they're generally not preferred anymore.

src/os/ogc/query/ogcquery.js
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * Launch a confirmation dialog to pick area to import.
+   * @param {!string} title
+   * @param {!string} message
+   * @param {string=} opt_icon
+   * @param {function(!ol.Feature, !ol.Feature)=} opt_callback Callback to fire when the area is chosen and available
+   */
+  launchImport(title, message, opt_icon, opt_callback) {
+    if (!ogclistdirective) return; // touch the <ogclist> directive to get it to inject...
```
jsalankey, line 71: This line shouldn't be necessary (also `return` statements at the start of functions are make Greg sad). The directive is registered to Angular by that file being included. In order to get around the linter/compiler error for an unused variable, simply don't assign the return value from the `goog.require`

src/os/ui/ogc/ogclist.js
```diff
@@ -0,0 +1,256 @@
+goog.module('os.ui.ogc.List');
```
jsalankey, line 1: The convention decided on in the ES6 docs is for these files/modules to be named things along the lines of `ListUI`. It is also preferred that the path and filenames match these module names as well as they can. If the file is titled ogclistui.js, the module should be `OGCListUI`

src/os/ui/draw/drawpicker.js
glennkitchellcaci: rofl - I think the original writer of onCountryChosen() was frustrated by the difference between the "name, id" ONLY feature and the fully-loaded-from-the-FeatureLayer feature... 
Anyway, this is gone now and I updated onCountryChosen() to work generically instead.

src/os/data/registry.js
glennkitchellcaci: Ended up needing some of the extended abilities in Registry; but added functions and made it an EventTarget just like CollectionManager.

src/os/ogc/query/ogcquery.js
glennkitchellcaci: Replaced all the .partial()'s with .bind()'s in the OGC code.

src/os/ogc/query/ogcquery.js
glennkitchellcaci: Updated to have only the require.

src/os/ui/ogc/ogclist.js
glennkitchellcaci: Updated filenames and namespace names.

src/os/ogc/ogcservice.js
glennkitchellcaci: Woot! Rewritten to use that.

src/os/ogc/ogcservice.js
```diff
@@ -0,0 +1,397 @@
+goog.module('os.ogc.OGCService');
+goog.module.declareLegacyNamespace();
+
+const Promise = goog.require('goog.Promise');
```
jsalankey, line 4: This shadows the browser's native `Promise` API. We don't actually use it anywhere, but I think we're gonna prefer using `GoogPromise` for this in the future.

src/os/ogc/ogcservice.js
```diff
@@ -0,0 +1,397 @@
...
+class OGCService extends AbstractService {
+  /**
+   * Constructor.
+   */
+  constructor() {
+    super();
+
+    /**
+     * Promise used to request all countries.
+     * @type {goog.Promise}
```
jsalankey, line 44: Types declared like this should use the locally defined variable name instead of the global type declaration.

At a glance, there are quite a few places where this is happening in this PR, so I won't point them all out, but please keep an eye out for them.

src/plugin/areadata/areadataplugin.js
```diff
@@ -0,0 +1,120 @@
...
+  });
+};
+
+/**
+ *
+ */
+class AreaDataPlugin extends AbstractPlugin {
+  /**
+   *
+   */
```
jsalankey, line 60: Nit: our linter requires these stupid extra comment blocks as a holdover from the pre-ES6 days. This shouldn't be necessary, but if we're gonna have them, we might as well put something in them.

src/plugin/areadata/areadataplugin.js
```diff
@@ -0,0 +1,120 @@
...
+
+    this.id = ID;
+  }
+
+  /**
+   * @inheritDoc
+   */
+  init() {
+    // Read the settings.json and initialize the Area importer(s)...
+    // NOTE: the "arraylike" object returned by Settings is not iterable; pass it through an Array.from()
```
jsalankey, line 72: This... should not be the case. Settings returns standard strings, objects, and arrays that should all be iterable in the normal ways. They aren't like `NodeList`s or other weird `ArrayLike` structures.

src/os/ogc/ogcservice.js
glennkitchellcaci: I preferred the full-resolution pathname in the jsdocs/casts so that:
1) among multiple files we'll always have the same string to search for, and
2) there's no inconsistencies/swapping back and forth for "unused" requires (for JSDoc/Type-ing only) and ones where we actually "use" the Object/Utilities in the require
> e.g. 
> If I only had "osx.blah.Blah" in a Type-ing comment, the require at the top of the doc would be goog.require('osx.blah.Blah');
> const a = /** @type {osx.blah.Blah} */ ({...});
> 
> Buf If I actually create a Blah at some point, then I have to go through and replace a bunch...
> const Blah = goog.require('osx.blah.Blah');
> const a = /** @type {Blah} */ ({...});
> const b = new Blah();
Whatcha think?

src/plugin/areadata/areadataplugin.js
glennkitchellcaci: This one was weird -- it worked just fine in the full code, but returned an "almost an array" array in the compiled code. So the following .forEach(() => {}) function didn't work even though .length worked (a property that ONLY exists on Arrays (and Strings, which are Arrays).

src/os/ui/draw/basedrawcontrols.js
```diff
@@ -9,6 +9,8 @@ goog.require('os.config.Settings');
 goog.require('os.data.RecordField');
 goog.require('os.interaction.DragZoom');
 goog.require('os.metrics.Metrics');
+goog.require('os.ogc.OGCService');
```
jsalankey, line 12: Requires for this in these files don't look to be needed anymore. I think you can get rid of the `declareLegacyNamespace` on that module too if you take these out.

src/os/ogc/query/ogcquery.js
```diff
@@ -0,0 +1,143 @@
...
+            .get(data[0])
+            .then(callback.bind(undefined, data[0]));
+      }
+    }
+  }
+
+  /**
+   * Handle area being loaded from the server. Insert into the Area manager.
+   * @param {!Feature} datum The requested area
+   * @param {!Feature} feature The area border feature
```
jsalankey, line 59: Is the distinction between these two arguments still necessary? Seems confusing if we're just creating a feature.

glennkitchellcaci: Removed!

src/os/ogc/query/ogcquery.js
```diff
@@ -0,0 +1,141 @@
+goog.module('os.ogc.query.OGCQuery');
+
+goog.require('os.ui.ogc.OGCListUI'); // touch the <ogclist> directive to get it to inject...
+goog.require('plugin.areadata'); // touch the AreaData plugin constants to bring the code into core OS
```
jsalankey, line 4: Plugin references should never be added outside of plugins and `MainCtrl`.

src/os/data/registry.js
```diff
@@ -0,0 +1,186 @@
...
+     * @type {!Object<string, T>}
+     */
+    this.map_ = {};
+  }
+
+  /**
+   * @inheritDoc
+   */
+  dispose() {
+    this.removeAllListeners('propertychange');
```
jsalankey, line 41: This should use the constant `goog.events.EventType.PROPERTYCHANGE`.

externs/plugin.externs.js
```diff
@@ -0,0 +1,44 @@
...
+ *  icon: (string|undefined)
+ * }}
+ */
+pluginx.areadata.ListUIOptions;
+
+
+/**
+ * @typedef {{
+ *  id: (!string),
+ *  type: (plugin.areadata.AreaImportType|undefined),
```
jsalankey, line 35: I got to talking with @schmidtk and he makes the argument that nothing in an extern should ever reference an internally defined type. This will eventually break with the upcoming module shift as these will no longer be referenceable here.

This is a pattern we've used elsewhere, but we should avoid going forward.

src/os/data/registry.js
```diff
@@ -0,0 +1,186 @@
...
+   */
+  on(onAddChange, onRemove, opt_this) {
+    // call for all the existing entries...
+    this.keys().forEach((key) => {
+      onAddChange.call(opt_this, key);
+    });
+
+    // then listen for changes to and run listeners as needed
+    this.listen('propertychange', function(event) {
+      switch (event.property_) {
```
jsalankey, line 138: This is referencing a private property on a property change event.

src/os/data/registry.js
```diff
@@ -0,0 +1,186 @@
...
+   * @param {T} value
+   * @param {...} opt
+   * @return {boolean} true if prior value overwritten
+   * @template T
+   */
+  register(key, value, ...opt) {
+    const had = this.has(key);
+    this.map_[key] = [value, opt];
+    if (!had) {
+      this.dispatchEvent(new PropertyChangeEvent('add', key));
```
jsalankey, line 164: The non-enumerated event types here should use an enum.

src/os/ogc/query/ogcquery.js
glennkitchellcaci: Removed.

externs/plugin.externs.js
glennkitchellcaci: This and MenuItemOptions are switched over to Object<string, *> for now.

PR #1000, opened 2020-08-07
feat(settings): allow reset settings button to be disabled

ericadeckl:
UCEP-559


src/os/ui/config/abstractsettings.js
```diff
@@ -52,6 +52,16 @@ os.ui.config.AbstractSettingsCtrl = function($scope, $timeout) {
...
 
+  /**
+   * @type {boolean}
+   */
+  this.scope['isResetDisabled'] = false;
+
+  /**
+   * @type {string}
+   */
+  this.scope['resetDisabledText'] = 'You are currently not allowed to Reset Settings.';
```
wallw-bits, line 63: Is it possible to provide more information here? This should already be obvious from the disabled button.

ericadeckl: I kept this generic so that whomever extends it can set the disabled text appropriately based on why they disabled the button.  For example, in bits-internal I set it to more specific text: 'In order to Reset Settings, you may not have the same application opened multiple times.  Please close all but one, and retry.'  I can't think of how to provide more detailed text in this file.

schmidtk: You could disable the button when `resetDisabledText` is truthy and set the default value to an empty string. Then you wouldn't need `isResetDisabled` at all, and the override could set `resetDisabledText` to an appropriate message when needed.

ericadeckl: Great idea.  Thanks!

src/os/config/settings.js
```diff
@@ -702,6 +702,15 @@ os.config.Settings.prototype.getLastReset = function() {
 };
 
 
+/**
+ * Retrieve the settings' peer info objects
+ *
+ * @return {Array<os.xt.PeerInfo>}
+ */
+os.config.Settings.prototype.getPeerInfo = function() {
```
schmidtk, line 710: From an API perspective, do we want to expose the entire peer? If not, we should consider supporting the `opt_type` param for `getPeerInfo`.

src/os/ui/help/resetsettingsoption.js
```diff
@@ -0,0 +1,17 @@
+goog.provide('os.ui.help.resetSettingsOption');
```
schmidtk, line 1: For new files, please use the `goog.module` format to reduce future transforms.

src/os/ui/help/resetsettingsoption.js
```diff
@@ -0,0 +1,17 @@
...
+goog.require('os.ui.help');
+goog.require('os.ui.util.ResetSettings');
+
+
+/**
+ * Add the Reset Settings option
+ */
+os.ui.help.resetSettingsOption.addToNav = function() {
+  var menu = os.ui.help.MENU;
+  menu.getRoot().addChild(os.ui.util.resetSettingsOptions);
```
schmidtk, line 13: `os.ui.util.resetSettingsOptions` is only used to add this menu option, so I would suggest moving that object into this file.

PR #1001, opened 2020-08-10
refactor(search): moves property/function to module

jsalankey:
Cleans up an improper declaration to allow for proper module requires.

### DIFF
--- a/src/os/ui/search/place/place.js
+++ b/src/os/ui/search/place/place.js
@@ -1,0 +1,60 @@
+goog.module('os.ui.search.place');
+goog.module.declareLegacyNamespace();
+
+const Feature = goog.require('ol.Feature');
+const osObject = goog.require('os.object');
+const StyleField = goog.require('os.style.StyleField');
+const StyleType = goog.require('os.style.StyleType');
+const {GOOGLE_EARTH_URL, GoogleEarthIcons} = goog.require('os.ui.file.kml');
+const {getUid} = goog.require('ol');
+const {setFeatureStyle} = goog.require('os.style');
+
+
+/**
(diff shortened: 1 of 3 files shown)


src/os/ui/search/place/place.js
```diff
@@ -0,0 +1,59 @@
+goog.module('os.ui.search.place');
+goog.module.declareLegacyNamespace();
+
+const Feature = goog.require('ol.Feature');
+const OSObject = goog.require('os.object');
```
schmidtk, line 5: This could use destructuring for specific import(s), but I'm also wondering how we want to handle case for module imports like this. From the [Google style guide](https://google.github.io/styleguide/jsguide.html#file-goog-require), I think this should use camel case (`osObject` or just `object` though I know that doesn't work when both os/goog object references are used).

src/os/ui/search/place/place.js
```diff
@@ -0,0 +1,59 @@
...
+
+
+/**
+ * @type {Object}
+ */
+const FEATURE_CONFIG = {
+  'image': {
+    'type': 'icon',
+    'scale': 0.75,
+    'src': os.ui.file.kml.GOOGLE_EARTH_URL + os.ui.file.kml.GoogleEarthIcons.WHT_BLANK,
```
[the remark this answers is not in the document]
jsalankey: Yep.

src/os/ui/search/place/place.js
jsalankey: I'd agree about following the style guide, but we'll need to communicate that more widely (I hadn't seen it before today).

PR #1002, opened 2020-08-10
fix(kml): fixes KMLs/NetworkLinks not merging on refresh

jsalankey:


src/plugin/file/kml/kmlsource.js
```diff
@@ -193,6 +193,19 @@ plugin.file.kml.KMLSource.prototype.setRootNode = function(node) {
 };
 
 
+/**
+ * @inheritDoc
+ */
+plugin.file.kml.KMLSource.prototype.setEnabledInternal = function(value) {
+  if (!value) {
+    // when the layer is disabled, drop reference to the root node to save memory
+    this.setRootNode(null);
```
schmidtk, line 202: Ah yeah, looks like you found what I was after with the `getRootNode` change. I believe the memory issue was the only reason I made that change.

jsalankey: Gotcha, I will do some memory profiling to see if we're still retaining old nodes.

PR #1004, opened 2020-08-17
feat(statemanager): handle nulls, non-constructables

glennkitchellcaci:
Improved logging of error conditions. Created utility function for boolean switch on isPersistable(); Updated logic in getPersistable() to log and failover in the recoverable state and log for non-recoverable state

### DIFF
--- a/src/os/state/basestatemanager.js
+++ b/src/os/state/basestatemanager.js
@@ -135,7 +135,7 @@ os.state.BaseStateManager.prototype.registerPersistable = function(key, clazz) {
   if (!(key in this.persistableMap_)) {
     this.persistableMap_[key] = clazz;
   } else {
-    goog.log.error(this.log, 'Peristance class for ' + key + ' already registered!');
+    goog.log.error(this.log, 'Persistence class for ' + key + ' already registered!');
   }
 };
 
@@ -145,7 +145,6 @@ os.state.BaseStateManager.prototype.registerPersistable = function(key, clazz) {
  *
  * @param {string} key The object key
  * @return {os.IPersistable}
- * @suppress {checkTypes} To allow the catch() failover to exist
  */
 os.state.BaseStateManager.prototype.getPersistable = function(key) {
   var instance = null;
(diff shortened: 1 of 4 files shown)


src/os/state/basestatemanager.js
```diff
@@ -141,17 +141,77 @@ os.state.BaseStateManager.prototype.registerPersistable = function(key, clazz) {
...
 os.state.BaseStateManager.prototype.getPersistable = function(key) {
-  if (key in this.persistableMap_) {
-    return new this.persistableMap_[key]();
+  var instance = null;
+  if (this.isPersistable(key)) { // check for nulls, etc
+    var Clazz = this.persistableMap_[key];
+    try {
+      instance = new Clazz();
+    } catch (e) {
+      var args = [this.log, `Registered persistance class for "${key}"`];
```
jsalankey, line 157: Nit: persistence is misspelled in every log statement as "persist*a*nce"

src/os/state/basestatemanager.js
```diff
@@ -141,17 +141,77 @@ os.state.BaseStateManager.prototype.registerPersistable = function(key, clazz) {
 
 
 /**
- * Gets a registered {@link os.IPersistable} class.
+ * Gets an instance of a registered {@link os.IPersistable} class.
  *
  * @param {string} key The object key
  * @return {os.IPersistable}
+ * @suppress {checkTypes} To allow the catch() failover to exist
```
jsalankey, line 148: Whatever this is suppressing, we should stop doing it and get rid of the statement. Suppress calls should never be used to try to help find an error as they're more likely to cause even more errors.

src/os/state/basestatemanager.js
glennkitchellcaci: Oh good, I thought I was crazy and didn't even check the spelling of the other message in here that I was using as my archetype. Updated my new ones and the old one too.

src/os/state/basestatemanager.js
glennkitchellcaci: Removed. Also removed the corresponding branching and logging types in the code.

PR #1005, opened 2020-08-17
feat(slickgrid): update to react to change in 'dragEnabled' value

johnstacy:


src/os/ui/slick/slickgrid.js
```diff
@@ -452,6 +437,53 @@ os.ui.slick.SlickGridCtrl.prototype.disposeInternal = function() {
 };
 
 
+/**
+ * @param  {boolean} newValue
+ * @param  {boolean=} oldValue
```
jsalankey, line 442: The `oldValue` parameter here should have be `opt_oldValue` since it's an optional parameter. This should be enforced by the compiler or the linter as well...

src/os/ui/slick/slickgrid.js
```diff
@@ -452,6 +437,53 @@ os.ui.slick.SlickGridCtrl.prototype.disposeInternal = function() {
...
+      this.grid.onDragInit.subscribe(this.onDragInitFn_);
+      this.grid.onDragStart.subscribe(this.onDragStartFn_);
+      this.grid.onDrag.subscribe(this.onDragFn_);
+      this.grid.onDragEnd.subscribe(this.onDragEndFn_);
+
+      /**
+       * @type {goog.Timer}
+       * @protected
+       */
+      this.scrollTimer = new goog.Timer(20);
```
jsalankey, line 457: Declaring new member properties here is weird now, but won't be allowed once we go to ES6 because it's flat out disallowed to do this outside of the constructor.

johnstacy: good to know about that in ES6

PR #1006, opened 2020-08-17
feat: added new OSM buildings layer

ameletyan:


src/plugin/basemap/basemapdescriptor.js
```diff
@@ -45,6 +45,11 @@ goog.inherits(plugin.basemap.BaseMapDescriptor, os.data.LayerSyncDescriptor);
  */
 plugin.basemap.BaseMapDescriptor.prototype.setConfig = function(value) {
   this.origConf_ = value;
+
+  // Layer type override from the base map config
+  if (value['layerType']) {
+    this.setType(value['layerType']);
```
schmidtk, line 51: The compiler does not track types for properties using bracket notation, so you'll need to cast this as a string to match the param requirement of `setType`.

src/plugin/cesium/tiles/cesium3dtileslayer.js
```diff
@@ -151,19 +154,48 @@ plugin.cesium.tiles.Layer.prototype.onTileProgress = function(pendingRequests, t
...
+  const accessToken = os.settings.get(`plugin.cesium.SettingsKey.ACCESS_TOKEN.${layerId}`);
+
   if (typeof config['assetId'] == 'number') {
     this.assetId = /** @type {number} */ (config['assetId']);
   }
 
   if (config['accessToken']) {
     this.accessToken = /** @type {string} */ (config['accessToken']);
+  } else {
+    this.accessToken = /** @type {string} */ accessToken;
```
schmidtk, line 167: Missing parens here around `accessToken` (parens are required around something being type cast).

schmidtk:
I added a change to this PR to fix restoring 3D tile layers after app refresh. The required classes (provider/descriptor/layer config) were being added in the renderer init instead of the Cesium plugin init, so they are not available when the data manager is creating providers from config.

schmidtk:
Example config to add the Cesium Ion OSM Buildings layer:
```
"providers": {
  "3dtiles": {
    "layers": {
      "osm-buildings": {
         "assetId": 96188,
         "projection": "EPSG:4326",
         "provider": "Cesium Ion",
         "tileStyle": {
           "color": "color('${roof:colour}')"
         },
         "title": "OSM Buildings",
         "type": "3dtiles"
      }
    }
  }
}
```

You should be prompted to enter an Ion access token when the layer is activated.

schmidtk:
@csgibson I didn't realize you recently moved the Cesium init code I moved here. What was the reasoning for that? Registering the 3D tiles classes on renderer init prevents the data manager from creating the provider, descriptors, etc from config because that happens immediately after plugins have finished loading. Those classes won't be registered until Cesium is actually activated, which isn't very desirable for config-driven 3D tile layers.

src/plugin/cesium/tiles/cesium3dtileslayer.js
```diff
@@ -65,10 +70,25 @@ plugin.cesium.tiles.Layer.prototype.synchronize = function() {
...
+    if (!isNaN(this.assetId) && this.accessToken) {
       tilesetUrl = Cesium.IonResource.fromAssetId(this.assetId, {
         accessToken: this.accessToken
       });
+      tilesetUrl.then(() => {
+        console.log('Access token accepted');
+      }, () => {
+        console.log('Access token rejected');
+
+        this.error = "The provided access token was rejected. Turn the layer off and back on to provide another access token."
```
schmidtk, line 82: This class extends `plugin.cesium.Layer`, which has a `getErrorMessage` function you can extend to provide this error message. In `plugin.cesium.Layer.prototype.synchronize`, that function is called and if it returns an error message the layer will dispatch the property change event to update the layer node. That means you shouldn't need to duplicate the dispatch logic here, or elsewhere in the class. Instead, you should be able to extend `getErrorMessage` to provide an appropriate message when `assetId` is defined and `accessToken` is not (these go hand-in-hand to access a remote Cesium Ion resource).

src/plugin/cesium/tiles/cesium3dtileslayer.js
```diff
@@ -65,10 +70,25 @@ plugin.cesium.tiles.Layer.prototype.synchronize = function() {
 
   if (!this.hasError()) {
     var tilesetUrl = '';
-    if (!isNaN(this.assetId)) {
+    if (!isNaN(this.assetId) && this.accessToken) {
       tilesetUrl = Cesium.IonResource.fromAssetId(this.assetId, {
         accessToken: this.accessToken
       });
+      tilesetUrl.then(() => {
+        console.log('Access token accepted');
```
schmidtk, line 78: Two things:
- Avoid using `console` statements. All log messages should use `goog.log` at a level appropriate to the message.
- We don't really care if the promise is resolved, so the resolution parameter can be `undefined` here. We only care if it's rejected (invalid token).

src/plugin/cesium/tiles/cesium3dtileslayer.js
```diff
@@ -134,6 +164,14 @@ plugin.cesium.tiles.Layer.prototype.setOpacity = function(value) {
   }
 };
 
+/**
+ * @param {string} errorMsg The message of the error
+ * @protected
+ */
+plugin.cesium.tiles.Layer.prototype.setTokenError_ = function(errorMsg) {
+  this.tokenError_ = errorMsg;
```
schmidtk, line 172: You might want to call `this.updateError();` here, instead of after each `setTokenError_` call. Whenever `this.tokenError_` changes we'll likely want to update the error on the layer.

If you want to prevent potential for duplicate updates when the error didn't actually change, you could wrap this block in `if (this.tokenError_ !== errorMsg)`.

src/plugin/cesium/tiles/cesium3dtileslayer.js
```diff
@@ -65,10 +77,24 @@ plugin.cesium.tiles.Layer.prototype.synchronize = function() {
...
       tilesetUrl = Cesium.IonResource.fromAssetId(this.assetId, {
         accessToken: this.accessToken
       });
+
+      tilesetUrl.then(() => {
+        // We don't care to note that it resolves, we just want a response if it doesn't
+      }, () => {
+        // Clear the saved access token because it was rejected
+        const layerId = this.getId();
+        os.settings.set(`plugin.cesium.SettingsKey.ACCESS_TOKEN.${layerId}`, '');
```
schmidtk, line 90: I initially thought it might be nice to save an access token per layer, but I'm back tracking on that. The default token in Ion has read for all assets, and you can only select individual assets by creating a new token. A user would be most likely to use their default token (and we instruct them to do so), so I don't think we need to save multiple.

Side note, the template string here needs an expression (`${...}`) around `plugin.cesium.SettingsKey.ACCESS_TOKEN`. It's currently being used as a literal setting key instead of evaluating the property to get the correct key. With the above suggestion you'll just use `os.settings.set(plugin.cesium.SettingsKey.ACCESS_TOKEN, <value>)` though.

src/plugin/cesium/tiles/cesium3dtileslayer.js
```diff
@@ -65,10 +77,24 @@ plugin.cesium.tiles.Layer.prototype.synchronize = function() {
...
+
+        var errorMsg = 'The provided access token was rejected. ' +
+        'Turn the layer off and back on to provide another access token.';
+        this.setTokenError_(errorMsg);
+        this.updateError();
+        this.getIcons();
```
schmidtk, line 96: No need to call `this.getIcons();` after the error is updated. When the error changes, the layer node will respond to the dispatched event and update the icons.

src/plugin/cesium/tiles/cesium3dtileslayer.js
```diff
@@ -151,19 +189,66 @@ plugin.cesium.tiles.Layer.prototype.onTileProgress = function(pendingRequests, t
...
   if (config['url']) {
     this.url = /** @type {string} */ (config['url']);
   }
 
   this.synchronize();
+
+  // Save the current layer so that we can run synchronize() on it asynchronously
+  const layerRef = this;
+  if (!this.accessToken) {
+    os.ui.window.launchConfirmText(/** @type {!osx.window.ConfirmTextOptions} */ ({
```
schmidtk, line 218: This block might be useful extracted as a generic convenience function, like `plugin.cesium.getAccessToken`. The function should return a promise that resolves on confirm, rejects on cancel. There is an example of this in `os.track.promptForTitle`.

src/plugin/cesium/cesium.js
```diff
@@ -39,6 +39,7 @@ plugin.cesium.GeometryInstanceId = {
  * @enum {string}
  */
 plugin.cesium.SettingsKey = {
+  ACCESS_TOKEN: 'cesium.accessTokens',
```
schmidtk, line 42: Given the suggestion to use a single access token, please update this value to `cesium.accessToken`.

src/plugin/cesium/tiles/cesium3dtileslayer.js
schmidtk: One other note - if the error changed and is not null/empty, we may also want to fire an alert so the user sees the message prominently in the UI. The icon change is subtle and may be missed.

PR #1009, opened 2020-08-20
feat(kml): added mimetype detect for unknown file extensions

gregroyal:


src/plugin/file/kml/kmlparser.js
```diff
@@ -656,6 +659,23 @@ plugin.file.kml.KMLParser.prototype.imagesRemaining_ = function() {
...
+ * @private
+ */
+plugin.file.kml.KMLParser.prototype.processUnknownZipAsset_ = function(entry) {
+  entry.getData(new zip.ArrayBufferWriter(), (data) => {
+    const type = os.arraybuf.getMimeType(/** @type {ArrayBuffer} */ (data));
+    if (!type) {
+      // eh whatever just skip this
+      this.kmzImagesRemaining_--;
+    } else {
+      entry.getData(new zip.Data64URIWriter(type), this.processZipAsset_.bind(this, entry.filename));
```
schmidtk, line 674: Should this verify the type is an image? We might want to ignore results like ZIP/PDF.

PR #1013, opened 2020-09-02
refactor(import/export): changes the ui to show unsupported states

Gamerlios123:
refactor(import/export): added tip for unsupported features

### DIFF
--- a/src/os/state/state.js
+++ b/src/os/state/state.js
@@ -90,6 +90,24 @@ os.state.titleCompare = function(a, b) {
 
 
 /**
+ * Compares states by whether or not they are supported
+ *
+ * @param {os.state.IState} a A state
+ * @param {os.state.IState} b Another state
+ * @return {number} The comparison
+ */
+os.state.supportCompare = function(a, b) {
+  if (a.getSupported() && !b.getSupported()) {
+    return -1;
+  }
+  if (!a.getSupported() && b.getSupported()) {
+    return 1;
+  }
+  return 0;
+};
+
+
+/**
  * @param {?Array<!os.data.IDataDescriptor>} list The list of state descriptors
  */
 os.state.deleteStates = function(list) {
--- a/src/os/ui/state/stateexport.js
+++ b/src/os/ui/state/stateexport.js
@@ -76,16 +76,14 @@ os.ui.state.StateExportCtrl = function($scope, $element, $timeout) {
    * @type {boolean}
    */
   this['all'] = goog.array.every(this['states'], function(state) {
-    return state.getEnabled();
+    return state.getSupported() ? state.getEnabled() : true;
   }, this);
 
   /**
    * @type {boolean}
    */
   this['isSaving'] = true;
-  $timeout(function() {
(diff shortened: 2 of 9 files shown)


csgibson (approved):

src/os/ui/state/stateexport.js
```diff
@@ -1,5 +1,6 @@
 goog.provide('os.ui.state.StateExportCtrl');
 goog.provide('os.ui.state.stateExportDirective');
+goog.provide('os.ui.state.unsupportedState');
```
jsalankey, line 3: As we are progressing with the module transformation work, adding multiple provides to a file is just making someone have to ensure we're happy with it getting decomposed when we modularize the file in the future. I'd prefer adding this to a state utility class or another file.

schmidtk: Agree. Any new code should avoid adding multiple provides to files.

src/os/ui/state/stateimportui.js
```diff
@@ -140,7 +140,7 @@ os.ui.state.StateImportUI.prototype.showUI = function(file, config) {
     'label': 'Import State',
     'icon': 'fa fa-file-text',
     'x': 'center',
-    'y': 'center',
+    'y': '100',
```
bradh, line 143: Why change / hard code this?

gregroyal: Since the height is 'auto', I'm assuming the window position was firing too soon. Instead of hardcoding it, you can just emit the window ready event when the window will probably be ready to center. Its already being done in the stateimport.js constructor. You can always put `  $scope.$emit(os.ui.WindowEventType.READY);` in a timeout OR better yet in a resolve of a promise if its waiting for a network request or something

Gamerlios123: Ya, that was the issue. Stateexport.js looks to have solved with a timeout so I added that solution to state import as well.

src/os/state/state.js
```diff
@@ -89,6 +89,24 @@ os.state.titleCompare = function(a, b) {
...
 
+/**
+ * Compares states by whether or not they are supported
+ *
+ * @param {os.state.IState} a A state
+ * @param {os.state.IState} b Another state
+ * @return {number} The comparison
+ */
+os.state.supportCompare = function(a, b) {
+  if (a['supported'] && !b['supported']) {
```
schmidtk, line 100: This function takes two `os.state.IState` instances, which do not define a `supported` property. You should add an `isSupported` method to the interface, and implement in `AbstractState` to expose the new property.

schmidtk (changes requested):

jsalankey:
This looks good, but please update the commit message to include "BREAKING CHANGE: <details about the interface change>". Adding a method to an interface is a breaking change to the State API as any downstream project that implements it separately from `AbstractState` will have to add the new methods.

jsalankey (dismissed):

schmidtk (approved):

bradh (commented):
I obviously don't have any insight into LINX, but don't see any issues

PR #1014, opened 2020-09-03
Use webpack for JS compilation

schmidtk:
Building with webpack will allow use of all module types (ES6, CommonJS, goog, etc) within our code prior to converting everything to ES6. Webpack already handles common module types imported with `require`, and the [closure-webpack-plugin](https://webpack.js.org/plugins/closure-webpack-plugin/) allows webpack to identify modules/files using `goog.module` and `goog.provide` statements.

Using webpack will be helpful moving forward with modules because it will resolve modules using a more sensible path. Without a bundler native `import` statements must use a relative path to the resource. This gets messy with Yarn workspaces used commonly by OpenSphere, especially when `node_modules` may or may not be hoisted.

Without webpack, an import might look like:
```
import someExport from '../../../node_modules/some-package/path/to/file';
```

With webpack, the import looks like:
```
import someExport from 'some-package/path/to/file';
```

Relative paths may still be more convenient to reference files in the current project and can still be used, but cross-project references will be much cleaner with this approach.

**Related/Required PR's**

- ngageoint/opensphere-build-closure-helper/pull/22 (merged/released)
- ngageoint/opensphere-build-index/pull/22 (merged/released)
- ngageoint/opensphere-build-resolver/pull/69 (merged/released)

### Significant Changes

**Debug Build**

The debug app is now solely generated by running `yarn dev` (formerly `yarn run build:debug`). Running `yarn build` will no longer create both compiled & debug builds. This is the case because webpack only generates one output at a time, and while we could run it twice (production + development) with `build` that didn't seem necessary.

The debug build generates an application bundle at `.build/opensphere.js` (controlled in the webpack config). This will help with issues we've experienced loading thousands of source files in a browser. Debugging still works as it did before through the source map generated by the build. You can still open individual source files in Dev Tools, make local changes to them, etc. For local changes in Dev Tools, you do have to modify the `webpack-internal` instance of the file.

Running `yarn dev` will leave webpack in watch mode, which watches the file system for changes in dependent files. If changes are detected, it will automatically rebuild the application bundle and you can refresh the browser to pick up the changes. It will watch for changes in all dependencies, including plugins external to `opensphere`.

**ES6/CommonJS Modules**

Webpack will resolve external modules imported using `require(<module name>)`. For example:

```
const assign = require('object-assign');
assign({}, someObj);
```

This in theory allows us to bring in external libraries, though running the Closure Compiler in advanced mode will likely result in errors with these files. More investigation will be required to resolve that.

### Upgrading Apps

To upgrade an app using OpenSphere as a library, do the following:

1. Update all of the `opensphere-build-*` dependencies in the project and add webpack dependencies seen in this PR. Note `npm-run-all` is optional, but adding it will make it easier to copy build script changes from OpenSphere.
2. Update build scripts, using changes to `package.json` in this PR as a guideline.
3. Create a `webpack.config.js` file in the app root. This can be copied from `opensphere/webpack.config.js` with the output filenames updated appropriately for the app.
4. Add a `debugJs` property to the options exported by any index file. This should point to the development output file from webpack.
5. Run the build and resolve any new errors. Due to how the dependency list is produced by the webpack build, the compiler may find additional errors it previously missed. Typically these errors are due to implied dependencies that should be referenced with `goog.require` or `goog.requireType`.
6. If you're using a Yarn workspace containing projects that have not been updated, it's recommended to add `"**/@ngageoint/closure-webpack-plugin"` to your `nohoist` configuration to ensure this package picks up the correct version of the compiler.


gregroyal:
This seems great! is it in draft still because its not working yet (build error)? This means we can start using ES6 Modules imports in our code? Can we upgrade Openlayers and use the ES6 imports from them? This seems like it could change our upgrade route if so

schmidtk:
This does work locally with changes from ngageoint/opensphere-build-index/pull/22, but that will need to be merged/released and the major version updated here before the build will pass. I would like to get buy in from a few developers before merging that breaking change and continuing forward.

This will allow importing ES6 modules with `require`. Only ES6 modules can use `import/export` so the syntax will be different until files are converted. We will also have to update `eslint-config-opensphere` to work with ES6 modules in our own code, but I have those changes ready on a branch.

In theory we will be able to upgrade Openlayers with this change, but we will have to figure out type checking with Closure. I'm not certain that will work directly since their JSDoc style has changed. I'll have to investigate further on that.

wallw-bits:
> This in theory allows us to bring in external libraries, though running the Closure Compiler in advanced mode will likely result in errors with these files. More investigation will be required to resolve that.

Minimally, I think we can just add more entries for `hide_warnings_for` like we do for Closure, OpenLayers, ol-cesium, etc. Ideally there would be a way to provide typing for external files (like externs except not unminified). It'd be cool if the compiler could pick up `*.d.ts` files but I believe TypeScript's type system is more descriptive than the Closure Compiler's, so that might not be possible.

jsalankey (approved):

wallw-bits (approved):

PR #1015, opened 2020-09-03
More memory

bcleveland1022:
Added ability to run using increased memory.  Added a UI configuration space for plugins to add memory configuration UIs for user to change memory limits.


scripts/electron/preload.js
```diff
@@ -1,5 +1,5 @@
-const {contextBridge, ipcRenderer} = require('electron');
-
+const {contextBridge, ipcRenderer, remote} = require('electron');
+const {getMaximumMemory, setMaximumMemory} = require('../memconfig.js');
```
schmidtk, line 2: When referencing a module in another project, we prefer to use the full Node path:

```
require('opensphere-electron/app/src/memconfig.js')
```

views/plugin/electron/electronmemoryconfig.html
```diff
@@ -0,0 +1,14 @@
+<div>
+  <h5>Memory</h5>
+  <div class="form-group row flex-shrink-0">
+    <label class="col-form-label col-2">Max MB</label>
+    <input class="form-control col-2" id="maxMemory" type="number" min="512" step="1"
+        ng-model="ctrl.maxMemory" ng-change="ctrl.update()" required >
```
schmidtk, line 6: - I would wait to update the value until the user clicks the apply/restart button, and maybe flip that to "Apply and Restart". Currently it saves on input change, which may be bad if the user types in a value and closes Settings (not intending to change it).
- There should also be a `max` value set to the system's max available memory.
- If the value is invalid (too small or large), display an error message telling them what the valid range is and disable the apply button.

src/plugin/electron/electronconfig.js
```diff
@@ -0,0 +1,57 @@
+goog.provide('plugin.electron.ElectronConfigCtrl');
+goog.provide('plugin.electron.electronConfigDirective');
+
+goog.require('os.ui.Module');
+
+/**
+ * The electron configuration directive.
+ * @return {angular.Directive}
+ */
+plugin.electron.electronConfigDirective = function() {
```
schmidtk, line 10: We might want to be more consistent with the directive/template/file naming. Is this intended to be general Electron configuration or just memory config?

schmidtk (changes requested):

scripts/electron/preload.js
bcleveland1022: That didn't seem to work, is it because the preload.js is copied into `opensphere-electron` directory?

schmidtk: You're right, that won't work in the built application. The current solution works.

src/plugin/electron/electronconfig.js
bcleveland1022: changed the naming to be more consistent with memory config.

views/plugin/electron/electronmemoryconfig.html
bcleveland1022: Implemented everything suggested.

src/plugin/electron/electronmemoryconfig.js
```diff
@@ -0,0 +1,64 @@
+goog.provide('plugin.electron.ElectronMemoryConfigCtrl');
```
jsalankey, line 1: This file should be converted to a module. You should be able to just run the jscodeshift script on it and not have to do much else.

bcleveland1022: done.

jsalankey (dismissed):

jsalankey (approved):

schmidtk (approved):

PR #1016, opened 2020-09-03
Externs cleanup for moment, D3

welchyd:
The moment externs had a couple functions that needed optional params
I disabled some linter errors and cleaned up some urls while I was in there

### DIFF
--- a/externs/d3-3.4.1.js
+++ b/externs/d3-3.4.1.js
@@ -1,9 +1,10 @@
+/* eslint-disable jsdoc/require-returns-check */
 /**
  * @fileoverview Externs for D3 3.4.1.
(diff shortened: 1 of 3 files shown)


jsalankey (approved):

externs/d3-3.4.1.js
```diff
@@ -1,9 +1,10 @@
+/* eslint-disable jsdoc/require-returns-check */
```
wallw-bits, line 1: Why were these necessary? We do not lint externs.

welchyd: It's not really necessary, but some of the other externs have it

PR #1018, opened 2020-09-07
feat(es6): modify for use of new es6 transform for implicit require(s)

glennkitchellcaci:
Followed precedent to use a singleton to replace os.dispatcher; and will use it throughout the codebase. Mostly this refers to the new .jscodeshift.json file which is used in the providetomodule.js transformation to transform implicit require(s) to ES6.


src/os/os.js
```diff
@@ -5,15 +5,18 @@ goog.require('goog.events.EventTarget');
...
 
 
 /**
+ * TODO <AT>deprecated Please use goog.require('os.dispatch.dispatcher').getInstance() instead
+ *
  * The global event dispatcher.
  * @type {goog.events.EventTarget}
  */
-os.dispatcher = new goog.events.EventTarget();
+os.dispatcher = os.dispatch.dispatcher.getInstance();
```
schmidtk, line 19: If you use `get/set` for this and make those calls pass through to `getInstance` and `setInstance`, then any existing code that does `os.dispatcher = <new value>` will correctly update the instance.

src/os/dispatch/dispatcher.js
```diff
@@ -0,0 +1,33 @@
...
+const EventTarget = goog.require('goog.events.EventTarget');
+
+/**
+ * @type {EventTarget|null}
+ */
+let instance = null;
+
+/**
+ * singleton
+ * @return {EventTarget}
```
schmidtk, line 13: `!EventTarget` would be more accurate, given the return type is guaranteed to be non-null.

schmidtk (changes requested):

src/os/dispatch/dispatcher.js
```diff
@@ -0,0 +1,33 @@
+goog.module('os.dispatch.dispatcher');
```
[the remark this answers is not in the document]
glennkitchellcaci: Yeah, it's a little wonky for sure because of the namespace overlap and the lack of it being a class. At first I was thinking I'd just have it be a class that extends EventTarget, but that broke things down the line with the Analyze tool. So I figured it'd be likely to break things I couldn't see too. It looks like Kevin's idea will work great though if I can get it to not break everything attached to "os" -- we'll see soon.

src/os/os.js
glennkitchellcaci: Awesome idea -- but got blocked by how closure treats the "provide('blah')" namespace. So, as discussed, I put some tech debt into the code to use the new os.Dispatcher in the transforms but keep it pointed at this os.dispatcher for now for backward compatibility. Will re-update os.Dispatcher to use its own "instance" variable again later.

schmidtk (approved):

PR #1021, opened 2020-09-11
feat(tile): add tile layer sharpness control

schmidtk:
Adds a sharpness control to Tile and Base Map layers. Access the control from Layers > (Tile/Base Map Layer) > Style.

Sharpness is adjusted using the convolution filter here: https://www.html5rocks.com/en/tutorials/canvas/imagefilters/

Tests will fail until ngageoint/opensphere-state-schema/pull/18 is merged to add the property to the state schema.

### DIFF
--- a/src/os/color.js
+++ b/src/os/color.js
@@ -820,28 +820,17 @@ const sharpnessMatrix = [
  * @param {!Array<number>} data The image data.
  * @param {number} width The image width.
  * @param {number} height The image height.
- * @param {number} sharpness The target sharpness, between 0 and 1.
- */
-const adjustSharpness = function(data, width, height, sharpness) {
-  // Multiply sharpness by a factor of 10 to enhance the effect of the convolution algorithm.
-  convolute(data, width, height, sharpnessMatrix, sharpness * 10);
-};
-
-
-/**
- * Applies a convolution transform to an array of image data.
- *
- * @param {!Array<number>} data The image data.
- * @param {number} width The image width.
- * @param {number} height The image height.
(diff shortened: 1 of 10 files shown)


schmidtk:
When sharpening an image, the sum of the weights must equal 1 to avoid impacting the brightness of the center pixel for each pass. This was not the case when adjusting pixels at the edge of an image, because there isn't an adjacent pixel to consider outside the edge. The result was a white border around the image where the brightness was being blown out. I fixed this, but the solution is specific to sharpening so I replaced the `convolute` function with `adjustSharpness`.

src/os/color.js
```diff
@@ -800,6 +800,91 @@ const adjustColor = function(data, brightness, contrast, saturation) {
...
+ * Applies a sharpness transform to an array of image data.
+ *
+ * @param {!Array<number>} data The image data.
+ * @param {number} width The image width.
+ * @param {number} height The image height.
+ * @param {number} value The sharpness value.
+ *
+ * @see https://www.html5rocks.com/en/tutorials/canvas/imagefilters/
+ */
+const adjustSharpness = function(data, width, height, value) {
```
jsalankey, line 827: I know we took the implementation here from html5rocks, but it might still good to add tests for this if we ever adjust it.

schmidtk: I started making tests, but any adjustment changes the output which will break any useful test. So you're not really testing for a correct answer, you're just testing for the answer that's currently produced by the function. Didn't seem like a good exercise.

jsalankey (approved):

PR #1024, opened 2020-09-15
chore(query): replace os.query.isWorldQuery into os.query.utils

jkae:
- Consolidate `os.query` and `os.query.utils` by migrating `isWorldQuery` and related functions / constants
- Update `os.query.utils` to be a module
- Update callers

Breaking change for plugins that utilize `os.query.isWorldQuery`.

### DIFF
--- a/src/os/query/queryutils.js
+++ b/src/os/query/queryutils.js
@@ -109,3 +109,5 @@ exports.WORLD_ZOOM_FEATURE = new olFeature(new Polygon([[
 
 
 exports.WORLD_EXTENT = WORLD_EXTENT;
+exports.WORLD_COORDS = WORLD_COORDS;
+exports.WORLD_GEOM = WORLD_GEOM;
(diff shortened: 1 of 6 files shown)


jsalankey:
Can you update the commit message here to include a breaking change line? This is moving stuff from one namespace to another, so we expect a line giving a basic explanation what was moved and how to update to to fix it (i.e. "BREAKING CHANGE: <explanation>").

src/os/query/queryutils.js
```diff
@@ -1,56 +1,113 @@
...
+  [181, 90],
+  [181, -90],
+  [179, -90],
+  [179, 90]
+]]));
+
+
+exports.WORLD_EXTENT = WORLD_EXTENT;
+exports.WORLD_COORDS = WORLD_COORDS;
+exports.WORLD_GEOM = WORLD_GEOM;
```
jsalankey, line 113: Not worth holding this up over, but a cleaner way to define the exports is to just declare everything as `const/let` in the file and then define the `exports = { ... }` object down here instead of a mix.

jsalankey (approved):

PR #1026, opened 2020-09-16
feat(service): add cache capability to abstract service

csgibson:
Adds cache capability to abstract service but leaves it up to child services to trigger.

Responses events are added to a map. Then when a new request comes through it checks against the map and if found it returns the event.

Another map tracks an expiration time for that request. After that time the response is deleted from the map and any further request will go through the cycle again.

### DIFF
--- a/src/os/ui/abstractservice.js
+++ b/src/os/ui/abstractservice.js
@@ -25,6 +25,12 @@ os.ui.AbstractService = function() {
    * @type {goog.log.Logger}
    */
   this.log = os.ui.AbstractService.LOGGER_;
+
+  /**
+   * in milliseconds: maximum allowed time between requests before responses will be considered stale
+   * Recommended that child classes set their own
+   */
+  this.maxTimeBeforeStale = 5 * 60 * 1000;
 };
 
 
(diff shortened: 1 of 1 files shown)


csgibson:
@glennkitchellcaci would `os.ogc.OGCService` also benefit from this? I can migrate it over

glennkitchellcaci:
> @glennkitchellcaci would `os.ogc.OGCService` also benefit from this? I can migrate it over

That'd most definitely help with the datasets that we're using right now (that are almost completely static) -- I haven't read into the code too far, but if the caching is disable-able on a per-service basis, then that'd future-proof OGCService too.

csgibson:
> 
> 
> > @glennkitchellcaci would `os.ogc.OGCService` also benefit from this? I can migrate it over
> 
> That'd most definitely help with the datasets that we're using right now (that are almost completely static) -- I haven't read into the code too far, but if the caching is disable-able on a per-service basis, then that'd future-proof OGCService too.

It's really an opt-in for each service. Instead of calling `request.load()` you'd call `this.loadRequest(request)`.

Currently looking into other options as well. i.e. server configuration for more effective browser caching.

wallw-bits:
Instead of implementing a cache here, it'd be preferable for any service which wants to hang on to its response to simply do so. In other words, just keep the promise from `request.getPromise()` around and anything which wants to attach a `.then()` to it will get the value. Anything other than that should probably be handled by browser cache directives, since those are much more robust.

csgibson:
Going to address our immediate need in the one service. Then come back to this when we have more time to devote to it.

PR #1027, opened 2020-09-16
fix(position): fix whitespace and case before mgrs parse

wallw-bits:
`CoordinateSearch` strips whitespace, uppercases, and checks a regexp before sending the value off to `osasm`. Now the `position` field does that too.

### DIFF
--- a/src/os/ui/geo/position.js
+++ b/src/os/ui/geo/position.js
@@ -5,6 +5,7 @@ goog.provide('os.ui.geo.positionDirective');
 goog.require('os.geo');
 goog.require('os.ui.Module');
 goog.require('os.ui.geo.geoDirective');
+goog.require('os.ui.geo.mgrs');
 goog.require('os.ui.popover.popoverDirective');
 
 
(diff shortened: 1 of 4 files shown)


jsalankey (dismissed):

src/os/ui/geo/position.js
```diff
@@ -189,19 +189,23 @@ os.ui.geo.PositionCtrl.prototype.destroy_ = function() {
...
       result = null;
     }
     if (result == null) {
       try {
-        var coord = osasm.toLonLat(this.scope_['posText']);
-        result = /** @type {!osx.geo.Location} */ ({
-          lon: coord[0],
-          lat: coord[1]
-        });
+        const mgrs = text.replace(/\s+/g, '').toUpperCase();
```
schmidtk, line 201: `os.ui.geo.geoLinkFn` will need similar logic, or this form needs to validate differently. Currently the position is parsed, but the field is still marked as invalid.

schmidtk (changes requested):

src/os/ui/geo/position.js
wallw-bits: Gotcha. I'll extract that out.

schmidtk (approved):

PR #1030, opened 2020-09-18
feat(featureactions): Promise-based Eventing

glennkitchellcaci:
FeatureActions' eventing now rolls-up to the ImportManager level to run the os.styles events once. For 150K features with multiple feature actions, multiple style changes, this is 59.4% (2.5x) faster applying the styles and 68.3% (3.2x) faster unapplying them.

### DIFF
--- a/src/os/im/action/importactionmanager.js
+++ b/src/os/im/action/importactionmanager.js
@@ -103,6 +103,13 @@ os.im.action.ImportActionManager.STORAGE_KEY = 'os.importActions';
 
 
 /**
+ * @type {number}
+ * @const
+ */
+os.im.action.ImportActionManager.MIN_ITEMS_MERGE_NOTIFY_COLOR = 10000;
+
+
+/**
  * @inheritDoc
  */
 os.im.action.ImportActionManager.prototype.disposeInternal = function() {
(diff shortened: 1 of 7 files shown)


glennkitchellcaci:
We need orders of magnitude speed improvements not just 3x or 4x -- But this is a cohesive change to just one area of the code, so after a little review and polish, I think this will be good.

The much more dangerous change is to os.style -- and that'll be in a wholly separate endeavor.

src/os/im/action/importactionmanager.js
```diff
@@ -351,24 +358,74 @@ os.im.action.ImportActionManager.prototype.getEntryItems = function(type) {
...
+          ps.forEach((p) => {
+            promises.push(p);
+          });
         }
       }
     }
+
+    // once all processItems() are done, do a big notify of style and color changes
+    if (promises.length > 0) {
+      Promise.allSettled(promises).then((results) => {
```
glennkitchellcaci, line 393: Use Promise.all() instead for old-browser compatibility; https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

glennkitchellcaci:
FYI: I tried to take out the Promises and just return the Config objects as suggested; but for some reason (maybe how the Chrome and Firefox JS engines manage memory in Promises?) the run-times for Feature Action Groups (multiple actions) on 150K Features were consistently 10+ seconds longer without them.

I have a git-stash of that change if we want it anyway.

glennkitchellcaci:
Using No-Promises branch instead. https://github.com/ngageoint/opensphere/pull/1041

PR #1031, opened 2020-09-18
Added method for combining separated Date and Time

Gamerlios123:


src/os/time/time.js
```diff
@@ -882,3 +882,22 @@ os.time.step = function(date, duration, offset, opt_local) {
...
+/**
+ * Combines the days and time for seperate mapped fields
+ * @param {string} days The year, month, day
+ * @param {string} time The hours, minutes, seconds, milliseconds
+ * @param {string} daysFormat The format for days
+ * @param {string} timeFormat The format for time
+ * @return {moment}
+ */
+os.time.combineDateTime = function(days, time, daysFormat, timeFormat) {
+  var daysMoment = os.time.parseMoment(days, daysFormat, true);
```
schmidtk, line 896: You should be able to avoid parsing into two separate `moment` objects by combining the format strings and parsing once. For example:

```
os.time.parseMoment(`${days} ${time}`, `${daysFormat} ${timeFormat}`, true)
```

This would be preferred in cases where this function may be called on a lot of data and performance is a concern.

src/os/time/time.js
```diff
@@ -882,3 +882,22 @@ os.time.step = function(date, duration, offset, opt_local) {
   var next = os.time.offset(date, duration, offset, opt_local);
   return (next.getTime() - date.getTime());
 };
+
+
+/**
+ * Combines the days and time for seperate mapped fields
+ * @param {string} days The year, month, day
+ * @param {string} time The hours, minutes, seconds, milliseconds
+ * @param {string} daysFormat The format for days
```
schmidtk, line 891: It might be useful to assume ISO 8601 format, so the `daysFormat` and `timeFormat` fields can be optional.

schmidtk (changes requested):
`time.js` has fairly good test coverage. Could you add tests for the new function? Tests should cover both happy path and error cases (null/invalid date/time strings).

Gamerlios123:
Ok, I added some tests and defaulted the formats to what I could find of ISO 8601 online. The date format is 'YYYY-MM-DD' and for the time, I couldn't find an example of a ISO 8601 time by itself so I used 'HH:mm:ss'.

schmidtk (approved):

schmidtk (approved):

PR #1035, opened 2020-09-22
Add support for Vector Tile layers

schmidtk:
This PR was built off work started by @wallw-bits to add Vector Tile support to OpenSphere.

The initial work here supports:

- Adding [Vector Tile](https://docs.mapbox.com/vector-tiles/reference/) layers via config or state files.
- Styling layers using the [MapBox Style Specification](https://docs.mapbox.com/mapbox-gl-js/style-spec/) by providing a `styleUrl` in the config.
- Displaying Vector Tiles in both 2D and 3D (Cesium).
- Double click interaction to display VT metadata in 2D/3D.

Future work will add an import UI to allow adding these layers in-app.

Sample configuration to add Vector Tile layers is attached below. Note the [MapTiler](https://www.maptiler.com/) config will require setting up a free account and adding your own access token in the `key` URI params.

[Vector Tile Config](https://github.com/ngageoint/opensphere/files/5777990/vectortiles.txt)


src/plugin/vectortile/vectortilelayerconfig.js
```diff
@@ -0,0 +1,294 @@
...
+  /**
+   * @param {Object<string, *>} options The options
+   * @return {VectorTileSource}
+   */
+  getSource(options) {
+    // Copy the original options to avoid adding a non-serializable object to it.
+    const sourceOptions = Object.assign({}, options);
+
+    const vtFormat = /** @type {VectorTileFormat|undefined} */ (sourceOptions['format']);
+    sourceOptions['format'] = getVectorTileFormat(vtFormat || VectorTileFormat.MVT);
```
jsalankey, line 148: Is this format property used by us or in the superclass for the `VectorTileSource`? Do we need it in quotes if its an externed options type?

src/os/layer/vectortile.js
```diff
@@ -0,0 +1,727 @@
...
+const renamelayer = goog.require('os.ui.renamelayer');
+
+
+/**
+ * OpenSphere vector tile layer.
+ *
+ * @implements {ILayer}
+ * @implements {IGroupable}
+ */
+class VectorTile extends VectorTileLayer {
```
jsalankey, line 48: This isn't something we can fix here, but the amount of boilerplate code we've copied around for our tile layer classes makes me sad. This is 8 or 9 of them now that have identical interface code.

src/os/layer/vectortile.js
```diff
@@ -0,0 +1,727 @@
...
+        var projArea = olExtent.getArea(projExtent);
+        var layerArea = olExtent.getArea(layerExtent);
+        return !olExtent.isEmpty(layerExtent) && layerArea / projArea < 0.8;
+      case ActionEventType.IDENTIFY:
+      case ActionEventType.REFRESH:
+      case ActionEventType.SHOW_DESCRIPTION:
+        return true;
+      case ActionEventType.RENAME:
+        return Array.isArray(opt_actionArgs) && opt_actionArgs.length === 1;
+      case ActionEventType.MOST_RECENT:
```
jsalankey, line 577: Do VectorTiles support time? Are any of these other actions inapplicable?

src/plugin/vectortile/vectortileplugin.js
```diff
@@ -0,0 +1,72 @@
...
+    let i = arr.length;
+    while (i--) {
+      if (arr[i] instanceof OLDoubleClick) {
+        break;
+      }
+    }
+
+    // ... and add ours just before it
+    if (i > -1) {
+      interactions.insertAt(i, new DoubleClick());
```
jsalankey, line 49: Any concern with this mucking with other interactions? Performance concerns? Drill pick is pretty pretty pretty slow already with a large number of stacked features.

src/plugin/vectortile/vectortilelayerconfig.js
schmidtk: Good point. It's used by the OL class so `sourceOptions` will need a type cast, but `format` is defined on `olx.source.VectorTileOptions`.

src/plugin/vectortile/vectortileplugin.js
schmidtk: Interactions are run in reverse order, so this is running the VT double click after the normal one. If features are hit, this won't run. If features aren't hit, it will run the VT double click.

src/os/layer/vectortile.js
schmidtk: Yeah I'm not a fan either. We would have to add interface methods to the lower level OL classes to fix it, which would mean more mixins. Or the code would have to be moved off of the layer (not sure where). Will had originally reused methods from `os.layer.Tile` by assigning them off the prototype, but that felt a little wonky too and I had to change some of the implementations to suit vector tiles.

src/os/layer/vectortile.js
schmidtk: No they don't, at least not in the request. I removed that action and the others are all supported.

src/plugin/vectortile/vectortileplugin.js
schmidtk: I did notice the hover interaction was causing performance issues with a vector tile layer, because the hit detection done by `forEachFeatureAtPixel` is replaying the tile render. Vector tile layers aren't supported by hover/select/etc so I omitted them in `os.interaction.defaultLayerFilter`.

src/os/layer/vectortile.js
wallw-bits: There's a fairly nice way to do this in JS but the Closure Compiler doesn't like it.

schmidtk:
I rebased this against master and also fixed an issue where labels were given a default stroke when one wasn't specified in the MVT style. Our text reader would previously set label stroke to black/2px if one wasn't specified, and this was being used by our general feature labels. I changed text reader to leave the stroke undefined if the `stroke` property was explicitly set to `null`. This should avoid breaking changes, but support the "no stroke" case for all labels.

wallw-bits (approved):

PR #1038, opened 2020-09-25
Support multiple terrain providers

schmidtk:
OpenSphere is currently limited to a single terrain provider, which is determined by the last config loaded from settings. As a user, it would be useful to switch between available providers and also configure custom terrain. This PR addresses the first of those concerns, exposing all configured (and supported) terrain providers through Settings > Map > Display.

**Testing Notes**

[OpenSphere](https://merciful-yarn.surge.sh/)

- In Settings > Map > Display, there should now be a list of supported terrain providers under the Enable Terrain checkbox. This is only displayed if multiple unique (by `id`) terrain layers have been configured.
- Changing the terrain provider should update the active provider in settings. If terrain is enabled, it should update the provider currently used by the renderer.
- When the OSM Buildings is activated, you should be prompted to enter a Cesium Ion access token and given instructions to do so. Add your token and click OK.
- If Cesium World Terrain isn't already the active terrain provider and enabled, you should be prompted to enable it. Clicking Yes should switch CWT to the active terrain and turn it on. If you have Settings open while this happens, you should see these changes in the UI. Clicking No should do nothing, and you shouldn't be prompted again (this choice is saved in settings).
- **Note:** The "Fake Terrain" provider will black out the globe if enabled because it's not a real terrain provider. This is an existing issue and not addressed in this PR. The provider is simply there to test the above.

**Changes for WebGL Renderers**

- `this.supportedTerrainTypes` was added to `AbstractWebGLRenderer`. Renderers should add their supported types to this list, so Display Settings can populate available options.
- `AbstractWebGLRenderer#getActiveTerrainProvider` should be used to get the options for the active terrain provider. For backward compatibility, these options are still saved to `os.config.DisplaySetting.TERRAIN_OPTIONS` which has been deprecated in favor of `os.map.terrain.TerrainSetting.ACTIVE_TERRAIN`.

**Changes for App Settings**

- Configured terrain should now define a `title` property, which is the user-facing name in settings. If a title is not provided, a generic default title will be used. The new `id` property comes from the existing object key in config.
- Cesium 3D Tile layers can now be configured to enable Cesium World Terrain when activated. In config, set `"useWorldTerrain": true` to enable this. When the layer is turned on, the user will be prompted to activate CWT if not already active, and supported. This is useful for layers like [OSM Buildings](https://cesium.com/content/cesium-osm-buildings/) that use similar elevation data as CWT.

**Additional Changes**

- The `confirmtext` directive now supports rendering HTML in the `prompt` text. This was added to support advanced markup in messages displayed to users. Specifically, the prompt to create/provide a Cesium Ion access token now links out to Ion.

### DIFF
--- a/src/os/map/terrain.js
+++ b/src/os/map/terrain.js
@@ -3,8 +3,6 @@ goog.module.declareLegacyNamespace();
 
 const log = goog.require('goog.log');
 
-const Settings = goog.require('os.config.Settings');
-
 
 /**
  * Terrain event types.
@@ -50,13 +48,6 @@ const providers = [];
 
 
 /**
- * The active terrain provider.
- * @type {osx.map.TerrainProviderOptions|undefined}
- */
-let activeProvider = undefined;
-
-
-/**
  * If a terrain provider is available.
  * @return {boolean}
  */
@@ -84,12 +75,6 @@ const addTerrainProvider = (options) => {
     if (!providers.some((p) => p.id === options.id)) {
       providers.push(options);
       os.dispatcher.dispatchEvent(TerrainEventType.PROVIDERS);
-
-      // if the terrain provider was previously active, activate it now
-      const activeId = Settings.getInstance().get(TerrainSetting.ACTIVE_TERRAIN);
-      if (activeId && activeId === options.id) {
-        setActiveTerrainProvider(options);
-      }
     } else {
       log.warning(logger, `Ignoring duplicate terrain provider ${options.name} with id '${options.id}'`);
     }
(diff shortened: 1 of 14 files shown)


jsalankey:
My token on the Cesium site was originally too long for the 200 character limit on the token input. Regenerated it and the new one is shorter and works, but we shouldn't need a length limit on that input.

src/os/map/terrain.js
```diff
@@ -0,0 +1,101 @@
...
+  if (isValidProvider(options)) {
+    if (!options.name) {
+      options.name = 'Unknown Terrain Provider';
+    }
+
+    if (!providers.some((p) => p.id === options.id)) {
+      providers.push(options);
+      os.dispatcher.dispatchEvent(TerrainEventType.PROVIDERS);
+    } else {
+      log.warning(logger, `Ignoring duplicate terrain provider ${options.name} with id '${options.id}'`);
```
jsalankey, line 79: I could possibly see a use case for allowing us to override an existing configured provider. We could log it (like we do with layer config overrides), but still allow the override to happen.

src/plugin/cesium/cesium.js
```diff
@@ -403,37 +417,136 @@ plugin.cesium.updateCesiumLayerProperties = function(olLayer, csLayer) {
...
+ * @param {string} prompt The message to display.
+ */
+plugin.cesium.promptForWorldTerrain = function(prompt) {
+  const showPrompt = os.settings.get(plugin.cesium.SettingsKey.SHOW_TERRAIN_PROMPT, true);
+  if (showPrompt && !plugin.cesium.isWorldTerrainActive() && plugin.cesium.hasWorldTerrain()) {
+    os.ui.window.ConfirmUI.launchConfirm(/** @type {!osx.window.ConfirmTextOptions} */ ({
+      confirm: plugin.cesium.enableWorldTerrain,
+      cancel: () => {
+        // Stop asking if the user says no
+        os.settings.set(plugin.cesium.SettingsKey.SHOW_TERRAIN_PROMPT, false);
```
jsalankey, line 477: Will this cause the prompt to never pop up again?

schmidtk: Unless they reset settings, yes. I wasn't sure how annoying I wanted this to be. Hopefully they click Yes because OSM Buildings in particular are awful if the terrain doesn't line up, but if they click No should we continue telling them whenever the layer is activated (including app refresh)? I could go either way.

src/os/map/terrain.js
schmidtk: If you configure a provider with the same ID, our settings merge will already handle merging in the overrides. If this case happens, it's because `addTerrainProvider` was called after the settings load with a duplicate ID. That seems to be an error case. I could see modifying an existing set of options, but you can do that directly on the object without this call.

src/plugin/cesium/cesium.js
schmidtk: I decided to remove this. If we want something like this, we should either implement a "remember this decision" or "don't ask me again" type checkbox on `confirmtext`.

schmidtk:
- Access token input expanded to 2000 chars.
- Added a method to create an Ion asset URL. This verifies the access token, and if the URL fails to load will clear the token in settings and alert the user that the token is invalid. This is used by 3D tile layers and Cesium World Terrain.
- If the 3D tile layer fails to load (invalid access token, etc), you will no longer be prompted to enable CWT.
- You will now be prompted is CWT is active but not enabled.
- Removed the setting to track if the user chose not to enable CWT. This would be better suited to a "Remember this decision" or "Don't ask me again" sort of checkbox, which we might consider adding to the confirm prompt.

jsalankey (approved):

PR #1041, opened 2020-09-28
feat(featureactions): Event Roll-Up

glennkitchellcaci:
FeatureActions' eventing now rolls-up to the ImportManager level to run the os.styles events once. For 150K features with multiple feature actions, multiple style changes, this is 59.4% (2.5x) faster applying the styles and 68.3% (3.2x) faster unapplying them.

### DIFF
--- a/src/os/im/action/importaction.js
+++ b/src/os/im/action/importaction.js
@@ -1,5 +1,4 @@
 goog.provide('os.im.action');
-goog.provide('os.im.action.ImportActionCallbackConfig');
 goog.provide('os.im.action.TagName');
 
 
@@ -42,18 +41,6 @@ os.im.action.TagName = {
   IMPORT_ACTIONS: 'importActions',
   IMPORT_ACTION: 'importAction'
 };
-
-
-/**
- * @typedef {{
- *  color: (Array<Iterable>|undefined),
- *  labelUpdateShown: boolean,
- *  notifyStyleChange: boolean,
- *  setColor: boolean,
- *  setFeaturesStyle: boolean
- * }}
- */
-os.im.action.ImportActionCallbackConfig;
 
 
 /**
--- a/src/os/im/action/importaction.js
+++ b/src/os/im/action/importaction.js
@@ -1,4 +1,5 @@
 goog.provide('os.im.action');
+goog.provide('os.im.action.ImportActionCallbackConfig');
 goog.provide('os.im.action.TagName');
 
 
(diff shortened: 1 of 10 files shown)


src/os/im/action/importactionmanager.js
```diff
@@ -351,24 +358,72 @@ os.im.action.ImportActionManager.prototype.getEntryItems = function(type) {
...
+        }
+        if (cfgs) {
+          cfgs.forEach((cfg) => {
+            configs.push(cfg);
+          });
         }
       }
     }
+
+    if (configs.length > 0) {
```
schmidtk, line 391: The following seems to be geared around items being `ol.Feature` instances. Should this be in `plugin.im.action.feature.Manager`?

src/os/im/action/importactionmanager.js
```diff
@@ -381,13 +436,132 @@ os.im.action.ImportActionManager.prototype.processItems = function(entryType, op
...
+
+      // loop backwards through colors... and remove items overlap in previous entries (last one wins)
+      for (var i = (len - 1); i >= 0; i--) {
+        var ids = {};
+        var [items] = config.color[i] || [];
+
+        // map the array by ids
+        if (i > 0) { // skip when no more loops to do
+          (items || []).reduce((map, item) => {
+            map[item.id_] = true;
```
schmidtk, line 499: This is also specific to `ol.Feature`. The generic item handled by this manager may not have an `id_` field. I'm curious why the compiler didn't complain about this, given it's also a private field.

schmidtk (changes requested):

src/os/im/action/importaction.js
```diff
@@ -1,4 +1,5 @@
 goog.provide('os.im.action');
+goog.provide('os.im.action.ImportActionCallbackConfig');
```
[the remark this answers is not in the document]
glennkitchellcaci: Moved into its own file. Figured it didn't need the externs treatment since it's OK to minify; but let me know putting it into os.externs is better for another reason.

src/plugin/featureaction/featureactionmanager.js
```diff
@@ -118,6 +126,176 @@ plugin.im.action.feature.Manager.prototype.initialize = function() {
...
+
+
+/**
+ * Consolidate results of desired notification(s) from multiple FeatureActions
+ *
+ * @param {os.im.action.ImportActionCallbackConfig} config
+ * @param {os.im.action.ImportActionCallbackConfig} cfg
+ * @private
+ */
+plugin.im.action.feature.Manager.mergeNotify_ = function(config, cfg) {
```
jsalankey, line 178: Nit: variable naming on merge functions should make it clear which object is the `source` and which is the `target` (or the `to` and `from` object).

test/plugin/featureaction/featureactionmanager.test.js
```diff
@@ -0,0 +1,78 @@
+goog.require('os.im.action.ImportActionCallbackConfig');
+goog.require('plugin.im.action.feature.Manager');
+
+describe('plugin.im.action.feature.Manager', function() {
+  const FeatureActionManager = goog.module.get('plugin.im.action.feature.Manager');
```
jsalankey, line 5: `FeatureActionManager` isn't a module yet, so this isn't strictly necessary. However, if it works, then it does future proof us once we convert it to a module.

jsalankey (approved):

PR #1043, opened 2020-09-28
Revert "feat(vector): speed up processing of feature removal"

jkae:
Reverts ngageoint/opensphere#1025


jkae:
Unfortunately errors out if features decide they want to stick around and send a change event after they've been removed. Unlisten should be handled by OpenLayers because  while it doesn't actually work seem to work correctly -- via testing and some debug capture -- it's the expected behavior.

wallw-bits (approved):

PR #1044, opened 2020-09-29
feat: added dont show again checkbox to confirmation window

erica-manzer:


src/os/ui/window/confirm.js
```diff
@@ -53,6 +53,8 @@ const launchConfirm = function(opt_options, opt_scopeOptions) {
   scopeOptions['noButtonTitle'] = options.noButtonTitle || '';
   scopeOptions['noButtonClass'] = options.noButtonClass || 'btn-secondary';
   scopeOptions['formClass'] = options.formClass || 'form-horizontal';
+  scopeOptions['dontShowAgainCheckbox'] = options['dontShowAgainCheckbox'] || false;
```
schmidtk, line 56: It might be useful to make this a little more generic. For example, a confirmation window might want the text to read "Remember this decision" instead. What we're adding is a checkbox with a callback to inform the window creator of its state, the actual purpose is irrelevant to the confirm directive.

I would propose these options:
- `checkboxCallback`: Function to call when checkbox is toggled. Defaults to `goog.nullFunction` as you've already done.
- `checkboxText`: Text to display next to the checkbox. Defaults to an empty string.

The UI would show the checkbox if `checkboxText` is defined, hide it otherwise.

schmidtk (changes requested):

src/os/ui/window/confirm.js
```diff
@@ -53,6 +53,8 @@ const launchConfirm = function(opt_options, opt_scopeOptions) {
   scopeOptions['noButtonTitle'] = options.noButtonTitle || '';
   scopeOptions['noButtonClass'] = options.noButtonClass || 'btn-secondary';
   scopeOptions['formClass'] = options.formClass || 'form-horizontal';
+  scopeOptions['checkboxText'] = options['checkboxText'] || '';
```
schmidtk, line 56: Could you add both `checkbox` and `checkboxText` to `osx.window.ConfirmOptions` in our externs? Then `checkboxText` should use dot notation here, instead of bracket. That will enforce better typing in the compiler.

schmidtk (dismissed):

schmidtk (approved):

PR #1047, opened 2020-09-30
refactor(wfs): adds plugin support for different WFS types

jsalankey:
Adds the ability to register parser config/type detection for requesting and parsing different WFS formats.

### DIFF
--- a/src/plugin/ogc/wfs/wfslayerconfig.js
+++ b/src/plugin/ogc/wfs/wfslayerconfig.js
@@ -183,6 +183,11 @@ plugin.ogc.wfs.WFSLayerConfig.prototype.getRequest = function(options) {
     request.setDataFormatter(new os.ogc.wfs.WFSFormatter());
   }
 
+  const type = this.getBestType(options);
+  if (type && type.responseType != null) {
+    request.setResponseType(/** @type {goog.net.XhrIo.ResponseType} */ (type.responseType));
+  }
+
   return request;
 };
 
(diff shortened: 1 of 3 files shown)


src/plugin/ogc/wfs/wfslayerconfig.js
```diff
@@ -352,94 +357,113 @@ plugin.ogc.wfs.WFSLayerConfig.prototype.addMappings = function(layer, options) {
...
-  GML2: 2
+plugin.ogc.wfs.WFSLayerConfig.registerType = function(config) {
+  const configs = plugin.ogc.wfs.WFSLayerConfig.TYPE_CONFIGS;
+
+  if (!configs.find((c) => c.type === config.type)) {
+    configs.push(config);
+  }
+
+  // sort them by priority, highest first
+  configs.sort((a, b) => goog.array.defaultCompare(b.priority, a.priority));
```
schmidtk, line 398: You can simplify numeric sorts (and avoid using `defaultCompare`) with:
`(a, b) => b.priority - a.priority`

src/plugin/ogc/wfs/wfslayerconfig.js
```diff
@@ -352,94 +357,113 @@ plugin.ogc.wfs.WFSLayerConfig.prototype.addMappings = function(layer, options) {
...
-          return i;
+          return pref;
         }
       }
     }
   }
 
   this.params.remove('outputformat');
-  return 1;
+  return preferred[1];
```
schmidtk, line 440: This was intended to make GML3 the default format, which it will do unless `registerType` is called and index 1 changes. We may want to assign the predefined configs as their own objects, reference those objects in the `TYPE_CONFIGS` definition, and return the GML3 type here.

test/plugin/ogc/wfs/wfslayerconfig.test.js
```diff
@@ -85,7 +94,31 @@ describe('plugin.ogc.wfs.WFSLayerConfig', function() {
...
+        'outputformat': 'avro/binary'
+      },
+      'formats': ['application/json', 'avro/binary', 'gml3', 'GML2']
+    };
+
+    wfs.initializeConfig(config);
+    const result = wfs.getBestType(config);
+    expect(result).toBe(avroConfig);
+    expect(wfs.params.get('outputformat')).toBe('avro/binary');
+  });
```
schmidtk, line 123: It might be good to add a test for the `getRequest` change to verify the response type.

schmidtk (changes requested):

src/plugin/ogc/wfs/wfslayerconfig.js
jsalankey: Agreed. This was from my testing where I was pushing a higher priority one in. For the sake of my direct object comparison tests, having those as constants would be good.

schmidtk (approved):

PR #1052, opened 2020-10-07
chore(es6): added more globals to be used when running jscodeshift

smithtb:
Added some more globals to the jscodeshift file based on issues I ran into while converting plugin code.

### DIFF
--- a/.jscodeshift.json
+++ b/.jscodeshift.json
@@ -1,11 +1,18 @@
 {
   "globals": {
     "os.alertManager": {"require": "os.alert.AlertManager", "singleton": true},
+    "os.areaManager": {"require": "os.query.AreaManager", "singleton": true},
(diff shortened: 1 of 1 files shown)


.jscodeshift.json
```diff
@@ -1,11 +1,18 @@
 {
   "globals": {
     "os.alertManager": {"require": "os.alert.AlertManager", "singleton": true},
+    "os.areaManager": {"require": "os.AreaManager", "singleton": true},
+    "os.ui.areaManager": {"require": "os.AreaManager", "singleton": true},
```
jsalankey, line 5: Are these requires correct? I would think they should be `os.query.AreaManager`. Similar question with `FilterManager` below.

smithtb: Huh - not sure why I did that, but those are wrong. Fixed!

jsalankey (approved):

PR #1053, opened 2020-10-08
feat(presets): Presets from Services

glennkitchellcaci:
Loads Presets from Services; the first of which re-implements the Settings-based Presets load from file(s).
Services _optionally_ provide CRUD abilities for os.layer.Preset objects. At a minimum, find() should be implemented and the corresponding PresetServiceAction.FIND should be flagged as supported.


src/os/layer/preset/ipresetservice.js
```diff
@@ -0,0 +1,71 @@
+goog.module('os.layer.preset.IPresetService');
+goog.module.declareLegacyNamespace();
```
jsalankey, line 2: This file is only required by other modules, so it shouldn't need to call this.

src/os/layer/preset/abstractpresetservice.js
```diff
@@ -0,0 +1,192 @@
...
+   */
+  find(search) {
+    return new Promise((resolve) => {
+      resolve(null);
+    });
+  }
+
+  /**
+   * The compiler is wrong; preset, opt_boolean, and the return are exactly as they should be
+   * @suppress {checkTypes}
```
jsalankey, line 51: `checkTypes` suppressions are VERY strongly discouraged, to the point where we basically won't allow them anymore. Whatever compiler error is happening should be fixed correctly.

src/os/layer/preset/abstractpresetservice.js
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * Handle setPublished
+   * @param {!osx.layer.Preset} preset
+   * @param {!boolean} isPublished
+   * @param {!Function<?>} resolve
+   * @param {!Function<?>} reject
+   * @protected
+   */
+  setPublishedInternal(preset, isPublished, resolve, reject) {
+    preset['published'] = isPublished;
```
jsalankey, line 174: Properties on externed types can be safely referenced with . notation. Bracket notation actually prevents the compiler from properly type checking these sorts of calls.

views/layer/vector.html
```diff
@@ -12,7 +12,7 @@
         </div>
         <div class="d-flex col">
           <select class="custom-select" ng-model="vector.preset"
-              ng-options="preset.label for preset in vector.presets track by preset.id">
+              ng-options="(preset.label + (preset.default === true ? ' (default)' : '') + (preset.published === true ? '' : ' (unpublished)')) for preset in vector.presets track by preset.id">
```
jsalankey, line 15: This text seems like it could get a bit long. For the "Unpublished" state, I feel like displaying an asterisk and maybe altering the tooltip could maybe work? Similar to how a text editor indicates on its tab that your file has unsaved changes.

src/os/layer/preset/abstractpresetservice.js
```diff
@@ -0,0 +1,192 @@
...
+   * @param {!Function<?>} reject
+   * @protected
+   */
+  setDefaultInternal(preset, isDefault, resolve, reject) {
+    const id = preset['id'];
+    const layerFilterKey = preset['layerFilterKey'];
+
+    // loop through all presets in the layer; set the one with ID to isDefault
+    this.find(/** @type {osx.layer.PresetSearch} */ ({
+      'layerFilterKey': [layerFilterKey]
```
jsalankey, line 93: Properties on externed types don't need to be referenced in quotes.

src/os/layer/preset/abstractpresetservice.js
```diff
@@ -0,0 +1,192 @@
...
+    })).then(
+        (results) => {
```
jsalankey, line 95: The deep nesting of complex closures here concerns me a bit. These functions can be defined with typing for their variables like member functions on a class. I think that would help make it clearer what's happening here.

Also there's a bunch of bracket notation references to objects that might be externed types, where . notation will work.

src/os/layer/preset/layerpresetmanager.js
```diff
@@ -1,223 +1,240 @@
...
 
+  /**
+   * @param {string} key
+   * @param {IPresetService} service
+   * @param {...} opt
+   * @return {boolean} true if prior value overwritten
+   */
+  registerService(key, service, ...opt) {
+    this.presets_ = {}; // set presets to reload every layer
+    return this.services_.register.apply(this.services_, [key, service].concat(opt));
```
jsalankey, line 105: Does this need to be called with an `apply`?

src/os/layer/preset/layerpresetmanager.js
```diff
@@ -1,223 +1,240 @@
...
-  goog.log.error(os.layer.preset.LayerPresetManager.LOGGER_, 'Failed to load presets. Reason: ' + msg);
-};
+  /**
+   * Gets a promise that resolves to the presets for a given layer type.
+   *
+   * @param {string} type
+   * @param {string} url
+   * @deprecated Please create a os.layer.preset.PresetService instead
+   */
+  registerPreset(type, url) {}
```
jsalankey, line 124: Are these here to avoid a breaking change? If so, I feel like they aren't really deprecated... it's still a breaking change because registering this way will still not work even if the code will still build. I'd rather we just remove them.

src/os/layer/preset/layerpresetmanager.js
```diff
@@ -1,223 +1,240 @@
...
+   * Initializes the layer presets for a layer.
+   *
+   * @param {string} id The layer ID.
+   * @param {boolean=} opt_applyDefault Whether to apply the default styles on load.
+   * @protected
+   */
+  initPreset(id, opt_applyDefault) {
+    // use the filter key to pull the value from settings
+    var filterKey;
+    var layer = os.map.mapContainer.getLayer(id);
```
jsalankey, line 159: Sine you ES6-ified this class, references like this should be cleaned up as well.

views/layer/vector.html
glennkitchellcaci: I'll switch over to the asterisk for now. If we want to do an <option ng-repeat="..."> instead for more flexibility than <select ng-options="..."> gives us (e.g. tooltips, custom classes, etc) then we can cross that bridge later.

src/os/layer/preset/presetserviceaction.js
```diff
@@ -0,0 +1,16 @@
+goog.module('os.layer.preset.PresetServiceAction');
```
jsalankey, line 1: Minor code style consolidation thing: both this file and settingkey.js can be consolidated into the browser preset.js utility file.

src/os/layer/preset/layerpresetmanager.js
glennkitchellcaci: I need it to be called in this format `this.services_.register(key, service, opt[0], opt[1], opt[2], ...)` so I picked .apply() since it accepts all the arguments as a single array. There might be a way to do something like... `this.services.register(...arguments)` but I'm not sure -- and I seem to recall that gcc blocks developers from using the "arguments" pseudo-variable.

src/os/layer/preset/presetserviceaction.js
glennkitchellcaci: preset.js has a circular dependency issue from staticlayerconfig, since `os/layer/config/staticlayerconfig.js` needs `os/layer/vector.js` which loops back around to layers. I'll leave the "global" in there for now and just add it to the mental list of things that'll break in layers before we can go to ES6. Wheee. <sarcasm>

src/os/layer/preset/layerpresetmanager.js
glennkitchellcaci: Doing goog.require('os.MapContainer') properly creates a circular dependency somewhere in the os.layer chain. Added a comment here; but pushing forward in the interest of completing this feature on time.

src/os/layer/preset/abstractpresetservice.js
glennkitchellcaci: Even with doing the JSDoc cast on every parameter and the return, GCC still thought that the params and return were "?".  So, instead of @inheritDoc, just added the same JSDoc as ipresetservice.js.

src/os/layer/preset/abstractpresetservice.js
```diff
@@ -0,0 +1,213 @@
...
+    });
+  }
+
+  /**
+   * Re-set the params and return instead of inheritDoc.
+   * The compiler is wrong; preset, opt_boolean, and the return are exactly as they should be.
+   *
+   * @param {osx.layer.Preset|string} preset
+   * @param {boolean=} opt_boolean
+   * @return {!Promise<osx.layer.Preset|null,?>}
```
jsalankey, line 55: If you want to make one last effort at using `@inheritDoc` here, the only other weird thing is the `,?` part of the type annotation.  I don't really know what that's supposed to mean in the context of a solution of a promise. `?osx.layer.Preset` makes sense as "null or an osx.layer.Preset", but the comma is weird.

jsalankey (approved):

PR #1054, opened 2020-10-08
Vidoverlay

bcleveland1022:
Prevent the video overlay from showing up on cesium.

### DIFF
--- a/src/plugin/cesium/cesiumrenderer.js
+++ b/src/plugin/cesium/cesiumrenderer.js
@@ -127,6 +127,10 @@ plugin.cesium.CesiumRenderer.prototype.isInitialized = function() {
  */
 plugin.cesium.CesiumRenderer.prototype.initialize = function() {
   if (!this.olCesium_ && this.map) {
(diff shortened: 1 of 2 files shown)


src/plugin/cesium/cesiumrenderer.js
```diff
@@ -127,6 +127,10 @@ plugin.cesium.CesiumRenderer.prototype.isInitialized = function() {
  */
 plugin.cesium.CesiumRenderer.prototype.initialize = function() {
   if (!this.olCesium_ && this.map) {
+    // Do not allow user to turn on video tile overlay
+    if (plugin.fmv && plugin.fmv.TILE_OVERLAY_ENABLED) {
```
bradh, line 131: Since this isn't (yet?) part of OpenSphere, is there a more general way to do this that doesn't involve coupling OS to the plugin? An interface on the plugin perhaps?

jsalankey: Yep, need to find a cleaner way to couple this to an external plugin.

PR #1065, opened 2020-11-04
fix: fix url modifier with multiple configs

wallw-bits:
The config here is called multiple times because the `baseUrl` stuff runs a separate config from `urlReplace`. Do not reset the replacement config so that all of those configs can be added.


schmidtk:
What was the original goal of clearing the replace array? Was that making a different assumption about when this would be called?

wallw-bits:
It just was. When we started using it for the base server stuff is when it got used twice, but since `urlReplace` is less likely to be configured in dev, and `baseUrl` is unlikely to be configured in production, this bug was not noticed now.

schmidtk (approved):

PR #1068, opened 2020-11-10
feat(presets): UI for admins

glennkitchellcaci:
These are the UI stubs for the management of Presets. It is disabled if there is not a PresetService in which the user is an Admin.


src/os/layer/preset/presetmenubutton_shim.js
```diff
@@ -0,0 +1,16 @@
+goog.provide('os.layer.preset.PresetMenuButtonCtrl');
```
jsalankey, line 1: We shouldn't need a shim file for this as its a new directive and only required in one place.

src/os/layer/preset/presetmenubutton.js
```diff
@@ -0,0 +1,220 @@
...
+    this['isAdmin'] = false;
+
+    this.init_();
+  }
+
+  /**
+   * @private
+   */
+  init_() {
+    this['isAdmin'] = LayerPresetManager.getInstance().isAdmin(); // requires ng-if
```
jsalankey, line 57: Is this guaranteed to be resolved at the time this UI first renders?

src/os/layer/preset/presetmenubutton.js
```diff
@@ -0,0 +1,220 @@
...
+        }, {
+          label: 'Set as Default',
+          eventType: EventType.TOGGLE_DEFAULT_TRUE,
+          tooltip: 'Make Preset the default style when layer added',
+          icons: ['<i class="fa fa-fw fa-star"></i>'],
+          handler: this.toggleDefault.bind(this),
+          metricKey: Metrics.TOGGLE_DEFAULT,
+          sort: 20
+        }, {
+          label: 'Revoke Default',
```
jsalankey, line 100: Revoke and Set as Default being separate menu options seems confusing. It doesn't look like they have a visibility toggle, so both will show at all times regardless of whether the selected option is already the default. It would be better for this to be a single option that changes its text and functionality based on the state of the underlying preset. The `beforeRender` function can change things like the `label` property as well I believe.

src/os/layer/preset/presetmenubutton.js
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * The preset directive.
+ * @return {angular.Directive}
+ */
+const directive = () => ({
+  restrict: 'E',
+  replace: true,
+  scope: {
+    'parentCtrl': '='
```
jsalankey, line 193: Passing the whole `parentCtrl` instance down to here creates an uncomfortably close coupling. I would rather see just the preset getting passed in and then potentially using Angular eventing for other communication between the UIs.

src/os/layer/preset/presetmenubutton.js
glennkitchellcaci: Yep, because it's under an ng-if, not an ng-show. The other option is to rewrite the LayerPresetManager to give access to its PresetServiceManager (which is an EventTarget) so others like this can listen to when events are thrown when Services are completed loading. This was a lot less code and eventing in the interest of finishing the other tickets; but the latter is more robust.

src/os/layer/preset/presetmenubutton.js
glennkitchellcaci: There's a beforeRender on the group which shows/hides the MenuItems as appropriate for the selected Preset -- it does so in one handler rather than multiple handlers. I can edit that handler to swap around the icons, labels, and events if we have to have less MenuItems; but this is simple and effective.

src/os/layer/preset/presetmenubutton.js
glennkitchellcaci: Ah yup -- I had just the preset but swapped it around while I was testing things out. I'll put that back.

glennkitchellcaci: Ty ty, @jsalankey -- I updated it to use the $emit from the child and $on in the parent.

jsalankey (approved):

PR #1069, opened 2020-11-11
feat(places): extract code to an abstract manager class

jkae:
In the event one might wish to create a managed KML layer similar to Saved Places, I've gone ahead and extracted most of the generic code out to an abstract class.

### DIFF
--- a/src/plugin/places/abstractkmlmanager.js
+++ b/src/plugin/places/abstractkmlmanager.js
@@ -1,0 +1,479 @@
+goog.module('plugin.places.AbstractKMLManager');
+
+const Delay = goog.require('goog.async.Delay');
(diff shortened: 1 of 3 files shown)


src/plugin/places/abstractkmlmanager.js
```diff
@@ -0,0 +1,472 @@
+goog.module('plugin.places.AbstractKMLManager');
```
jsalankey, line 1: This class should live under the `plugin.file.kml` namespace since it no longer has anything Places-specific in it. Within the plugin hierarchy, we prefer to keep classes in the more generic namespaces if possible.

jsalankey (approved):

PR #1070, opened 2020-11-18
feat(presets): Save layer style presets

glennkitchellcaci:
If a PresetService implements UPDATE and the User has permissions to do them, a UI is added to save Layer Styles and ImportActions as a Preset.


src/os/layer/preset/presetmenubutton.js
```diff
@@ -127,6 +137,98 @@ class Controller extends MenuButtonCtrl {
...
+   *
+   * @param {!osx.layer.Preset} source
+   * @return {!osx.layer.Preset}
+   */
+  clone(source) {
+    const clone = /** @type {osx.layer.Preset} */ (Object.assign({}, source));
+
+    // get the current layerConfig
+    if (clone.layerId) {
+      const layer = os.MapContainer.getInstance().getLayer(clone.layerId); // TODO fix global reference
```
glennkitchellcaci, line 151: MapContainer must require something in Layers that eventually winds around back to Presets... can't fix this require() until we unwind that; which is both potentially huge and better suited for the ES6 work.

src/os/layer/preset/presetmenubutton.js
```diff
@@ -145,12 +247,75 @@ class Controller extends MenuButtonCtrl {
...
+    const preset = this.clone(source);
 
-    console.log(`preset.save(${preset.label})`, preset);
+    // get PresetService(s) that support save()
+    const services = LayerPresetManager.getInstance().supporting(OsLayerPreset.PresetServiceAction.UPDATE);
+    if (services && services.length > 0) {
+      // TODO if more than one, open a dropdown modal to select one
+      services[0].update(preset).then(this.saveSuccess.bind(this), this.saveFailure.bind(this));
+    } else {
+      AlertManager.getInstance().sendAlert('No services found that support SAVE action', AlertEventSeverity.ERROR);
```
jsalankey, line 263: This is probably more of a developer error that would be better logged or asserted than as an alert to the user. The user wouldn't understand what this means or what to do about it.

glennkitchellcaci: Updated!

jsalankey (approved):

PR #1071, opened 2020-11-18
chore(package): use npm-run-all for copy

jkae:
I was personally having an issue with Windows/npm/git-bash where running the copy commands in parallel just crashed out of xargs. There seems to be a common, albeit random, problem where high process/service counts can cause this to happen. Using `npm-run-all` in sequential mode fixes it, though it might slow the build a little.

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     "copy:views": "xargs -n 1 -I @ sh -c 'cp -rf @ $(cat .build/version)/views/.' < .build/copy-views-args",
     "copy:config": "cp .build/settings.json dist/opensphere/config",
     "copy:resources": "xargs -n 2 cp -r < .build/resources-copy-files",
-    "copy": "npm run copy:files && npm run copy:images && npm run copy:onboarding && npm run copy:views && npm run copy:config && npm run copy:resources",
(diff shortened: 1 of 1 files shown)


schmidtk (approved):
Looks good. You can use the `run-s` shorthand for `npm-run-all -s` if you don't need to chain `-s` and `-p` sequences.

schmidtk:
The existing `&&` chain is running those commands sequentially. Shells will not start a command until the previous completes successfully. I would guess using `npm-run-all` introduces a small amount of overhead between commands, which might be why it's fixing your issue?

jkae:
> 
> 
> The existing `&&` chain is running those commands sequentially. Shells will not start a command until the previous completes successfully. I would guess using `npm-run-all` introduces a small amount of overhead between commands, which might be why it's fixing your issue?

I had read that `&&` spawns the processes all at the same time so I assumed they were in parallel, but maybe it's that instead. Or `npm-run-all` has a longer wait between executions or something, yeah.

schmidtk:
`A && B` will run B if and only if A succeeds. `A & B` would background A and run B in parallel.

PR #1072, opened 2020-11-18
OGC server URL handling fixes

estadtherr:
This request includes three changes:
1. Adds a boolean "parseOperationURLs" configuration parameter to OGC layers/providers. When set to true (the default), operation URLs parsed from Capabilities documents will be used for layer queries (GetMap and GetFeature). When false, the "wms" and "wfs" URLs set in the provider configuration will be used.
1. Fixes an error with configuring WFS layers with GML output format. Previously, the parser was not being set correctly because the wrong parameter was used to lookup the parser from the import manager.
1. Adds a unit test for the WMTS server module.


src/os/ui/ogc/ogcserver.js
```diff
@@ -857,17 +910,19 @@ os.ui.ogc.OGCServer.prototype.parseWfsCapabilities = function(response, uri) {
...
-          var attr = getFeatureEl.attributes[0];
-          // Attr.value is the DOM4 property, while Attr.nodeValue inherited from Node should work on older browsers
-          this.setWfsUrl(attr.value || attr.nodeValue);
+        if (this.parseOperationURLs_) {
+          if (getFeatureEl.hasAttributeNS(ol.format.XLink.NAMESPACE_URI, 'href')) {
+            this.setWfsUrl(getFeatureEl.getAttributeNS(ol.format.XLink.NAMESPACE_URI, 'href'));
+          } else {
+            var attr = getFeatureEl.attributes[0];
+            // Attr.value is the DOM4 property, while Attr.nodeValue inherited from Node should work on older browsers
+            this.setWfsUrl(attr.value || attr.nodeValue);
```
schmidtk, line 919: Should this be `this.setWFSOperationURL`?

src/os/ui/ogc/ogcserver.js
```diff
@@ -857,17 +910,19 @@ os.ui.ogc.OGCServer.prototype.parseWfsCapabilities = function(response, uri) {
...
+          }
         }
         this.setWfsPost(true);
       } else {
         getFeatureEl = op.querySelector('Get');
-        if (getFeatureEl != null) {
+        if (getFeatureEl != null && this.parseOperationURLs_) {
           if (getFeatureEl.hasAttributeNS(ol.format.XLink.NAMESPACE_URI, 'href')) {
             this.setWfsUrl(getFeatureEl.getAttributeNS(ol.format.XLink.NAMESPACE_URI, 'href'));
```
schmidtk, line 927: Here as well - `setWFSOperationURL`?

schmidtk (changes requested):

src/os/ui/ogc/ogcserver.js
estadtherr: I ended up removing the separate "operation URL" field and just set the `this.wfsUrl_` and `this.wmsUrl_` directly as it did before. I think the separate operation URL is a step towards possibly keeping track of the different URLs for the different operations in the Capabilities documents, but as just a single URL it seems redundant, since the`this.wmsUrl_` and `this.wfsUrl_` fields are already essentially used as the operation URL in the layer descriptors.

schmidtk (approved):

PR #1073, opened 2020-11-20
feat(presets): Implement Publish/Make Default/Delete

glennkitchellcaci:
Improved UX. Integrated calls to PresetServices to Publish/Make Default/Delete.

### DIFF
--- a/src/os/layer/preset/presetmenubutton.js
+++ b/src/os/layer/preset/presetmenubutton.js
@@ -15,9 +15,13 @@ const {Presets: OsMetrics} = goog.require('os.metrics.keys');
 const AlertManager = goog.require('os.alert.AlertManager');
 const AlertEventSeverity = goog.require('os.alert.AlertEventSeverity');
 const ConfirmUI = goog.require('os.ui.window.ConfirmUI');
+const OsUi = goog.require('os.ui');
+const {DEFAULT_PRESET_ID} = goog.require('os.layer.preset');
 
 const FilterActionEntry = goog.requireType('os.im.action.FilterActionEntry');
 const ILayer = goog.requireType('os.layer.ILayer');
+const IPresetService = goog.requireType('os.layer.preset.IPresetService');
+
 
 const GoogLog = goog.require('goog.log');
 
@@ -62,6 +66,11 @@ class Controller extends MenuButtonCtrl {
      * @type {boolean}
      */
     this['isAdmin'] = false;
+
+    /**
+     * @type {boolean}
+     */
+    this['thinking'] = false;
 
     this.init_();
   }
@@ -235,10 +244,15 @@ class Controller extends MenuButtonCtrl {
   }
 
   /**
+   * Get a service that will be able to do the requested action
+   *
    * @param {!OsLayerPreset.PresetServiceAction} action
+   * @param {osx.layer.Preset=} opt_preset
    * @return {Promise<IPresetService>}
    */
-  getService(action) {
+  getService(action, opt_preset) {
+    // TODO if the user passes in a Preset, get the service from which the preset came
+
     // get PresetService(s) that support save()
     const services = LayerPresetManager.getInstance().supporting(action);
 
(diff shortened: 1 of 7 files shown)


src/os/layer/preset/presetmenubutton.js
```diff
@@ -35,12 +38,16 @@ const EventType = {
...
   SAVE: 'save',
+  TOGGLE_PRESET: 'toggle-preset',
   TOGGLE_DEFAULT_TRUE: 'toggle-default-true',
   TOGGLE_DEFAULT_FALSE: 'toggle-default-false',
   TOGGLE_PUBLISHED_TRUE: 'toggle-published-true',
   TOGGLE_PUBLISHED_FALSE: 'toggle-published-false'
 };
 
+/**
+ * @const {GoogLog.Logger}
```
jsalankey, line 49: I think this is mixing up `@type {...}` with `@const` which should be a standalone

src/os/layer/preset/presetmenubutton.js
```diff
@@ -245,9 +282,60 @@ class Controller extends MenuButtonCtrl {
...
+    // Refresh the application
+    ConfirmUI.launchConfirm(/** @type {osx.window.ConfirmOptions} */ ({
+      yesText: 'Delete',
+      yesIcon: 'fa fa-trash-o',
+      yesButtonClass: 'btn-danger',
+      confirm: (() => {
+        // get PresetService(s) that support remove()
+        this.getService(OsLayerPreset.PresetServiceAction.REMOVE).then(
+            (service) => {
+              this['thinking'] = true;
```
jsalankey, line 301: Is the `this` context on this function correct? It doesn't look like the context argument was provided to the `.then()` call.

src/os/layer/preset/presetmenubutton.js
```diff
@@ -258,14 +346,17 @@ class Controller extends MenuButtonCtrl {
...
-    }
+    // get PresetService(s) that support uodate()
+    this.getService(OsLayerPreset.PresetServiceAction.UPDATE).then(
+        (service) => {
+          service.update(preset).then(
+              this.saveSuccess.bind(this),
+              this.onServiceFailure.bind(this, 'Could not save Preset'));
+        },
+        (msg) => {
+          GoogLog.error(LOGGER, 'No services found that support save action');
```
jsalankey, line 357: This should include the `msg` argument value in its logged message.

src/os/layer/preset/presetmenubutton.js
glennkitchellcaci: derp. ty ty

src/os/layer/preset/presetmenubutton.js
```diff
@@ -382,15 +540,18 @@ const directive = () => ({
   controllerAs: 'ctrl',
   template: `
 <div class="btn-group o-add-data-button" ng-right-click="ctrl.openMenu()">
-  <button type="button" class="btn btn-primary col-auto" ng-click="ctrl.notifyApplyPreset()"
-    title="Apply the layer style preset">
-    <i class="fa fa-check"></i>
+  <button type="button" class="btn btn-primary col-auto" 
+      ng-click="ctrl.notifyApplyPreset()"
+      title="Apply the layer style preset">
+    <i class='fa fa-check'></i>
```
jsalankey, line 546: Nit: single quotes in an HTML template. I thought the HTML linter disallowed this...

src/os/layer/preset/presetmenubutton.js
glennkitchellcaci: :point_up: Magic. I had moved one of these then's into a function at one point during these changes and discovered those ones needed the .bind(this), but these arrow functions are OK.

src/os/layer/preset/presetmenubutton.js
glennkitchellcaci: Updated. Ironically, msg was basically the same thing, so I just used it instead, lol.

jsalankey (approved):

PR #1076, opened 2020-11-24
fix(mapcontainer): add features returns correct features

arwatts16:
`addFeatures` returns the features added to the map rather than the ones passed into the function the same way `addFeature` works.


src/os/mapcontainer.js
```diff
@@ -1723,8 +1723,9 @@ os.MapContainer.prototype.addFeatures = function(features, opt_style) {
   var added = [];
   os.metrics.Metrics.getInstance().updateMetric(os.metrics.keys.Map.ADD_FEATURES, 1);
   for (var i = 0, n = features.length; i < n; i++) {
-    if (this.addFeature(features[i], opt_style)) {
-      added.push(features[i]);
+    var addedFeature = this.addFeature(features[i], opt_style);
```
bcleveland1022, line 1726: do we need to use const or let here?

bcleveland1022 (approved):

jsalankey (approved):

PR #1077, opened 2020-11-24
feat(feature-actions): adds checkbox for hiding default actions

jsalankey:


views/im/action/importactions.html
```diff
@@ -28,6 +28,10 @@
             Create
           </button>
         </div>
+        <div class="form-check form-check-inline" ng-if="ctrl.hasDefaultActions" title="Show default feature actions in the list. Uncheck to declutter the view.">
+          <input class="form-check-input" type="checkbox" id="showDefaultActions" ng-model="ctrl.showDefaultActions" ng-change="ctrl.toggleDefaultActions()">
+          <label class="form-check-label" for="showDefaultActions">Show Default Feature Actions</label>
```
schmidtk, line 33: Multiple feature action windows can be opened at the same time, which breaks the `for/id` interaction. You can either ensure `id` is unique for each instance of this directive, or you can nest the `input` inside `label` and remove `form-check-inline` to fix the styling.

schmidtk (changes requested):
Only issue I noticed was the checkbox behavior with multiple FA windows open.

views/im/action/importactions.html
jsalankey: Good catch, and good to have a way around using the `for/id` attributes because they always feel flaky to me.

Also noticed that some of our checkboxes on the layers window are styled differently (blue vs. white background in Chrome). Might need to make a pass to make these all consistent and generally avoid the `for/id` attributes.

schmidtk (approved):

PR #1078, opened 2020-11-25
feat(search): update feature result card behavior and extension points

schmidtk:
The feature result card did not handle the `SelectionType.CHANGED` event, which is fired when `setSelectedItems` is called on a vector source. This PR adds support for that event, as well as simplifying the source event handler behavior.

As part of this update, I added functions to set the select/highlight state of the card so extending classes may contribute their own behavior. I also added a `highlightConfig` property to the class, so extending classes may change how their result is highlighted on the map.


schmidtk:
Updated with a full test suite for `FeatureResultCardCtrl`.

jsalankey (approved):

PR #1079, opened 2020-11-26
feat(track): prompt to include metadata in tracks

schmidtk:
This PR addresses a few things with tracks:

- The Create Track option in the layer menu did not support including original metadata in the resulting track. This is now possible via checkbox in the UI prompt (defaults to false per existing behavior).
- The Add to Track now automatically includes metadata if previously included on the target track. This is intended to avoid gaps in metadata when features are added to the track.
- When including metadata in tracks, the additional fields are not added as columns to the track source (Saved Places currently). This prevented the fields from showing up in UI's that are based on source columns (Feature Info, Show Features, etc).
- The Create/Add To track options in the layer menu have confused some because they use selected data if available, all data if there isn't a selection. This is communicated in the tooltip, but we often get feedback that tooltips aren't read or folks don't know/care to read them. To resolve this, I have create separate events for creating/adding using all and selected data. This expanded the menu a bit, so I created a Tracks submenu to hold all track actions.

The `confirmtext` dialog did not previously support the checkbox control from `confirm`, so this has been added. I also changed the checkbox callback behavior so it _only_ fires when `confirm` is called (pressing OK). This is consistent with how the value callback is fired, and prevents changing the value backing the checkbox when the user cancels the dialog.

The new `promptForTitleAndMetadata` could replace `promptForTitle`, if we want to deprecate the latter. The resolve value is different so I couldn't update `promptForTitle` without a breaking change.

### DIFF
--- a/src/os/track.js
+++ b/src/os/track.js
@@ -1159,6 +1159,41 @@ os.track.promptForTitle = function(opt_default) {
 
 
 /**
+ * Prompt the user to choose a track title, and if original metadata should be included.
+ *
+ * @param {string=} opt_default The default title. Defaults to an empty string.
+ * @param {boolean=} opt_includeMetadata If metadata should be included. Defaults to false.
+ * @return {!goog.Promise}
+ */
+os.track.promptForTitleAndMetadata = function(opt_default = '', opt_includeMetadata = false) {
+  let includeMetadata = opt_includeMetadata;
+  const setIncludeMetadata = (value) => {
+    includeMetadata = value;
+  };
+
+  return new goog.Promise(function(resolve, reject) {
+    os.ui.window.launchConfirmText(/** @type {!osx.window.ConfirmTextOptions} */ ({
+      confirm: (title) => {
+        resolve({title, includeMetadata});
+      },
+      cancel: reject,
+      defaultValue: opt_default,
+      checkboxText: 'Include original metadata in track',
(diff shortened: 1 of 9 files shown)


src/os/track.js
```diff
@@ -1158,6 +1158,41 @@ os.track.promptForTitle = function(opt_default) {
...
 
 
+/**
+ * Prompt the user to choose a track title, and if original metadata should be included.
+ *
+ * @param {string=} opt_default The default title. Defaults to an empty string.
+ * @param {boolean=} opt_includeMetadata If metadata should be included. Defaults to false.
+ * @return {!goog.Promise}
+ */
+os.track.promptForTitleAndMetadata = function(opt_default = '', opt_includeMetadata = false) {
```
jsalankey, line 1168: Could this get merged into the existing function? I don't think anything else is calling the old one, and that function is already being used for the deprecated `plugin.track.promptForTitle` from when tracks were a plugin, could be confusing to have all these functions doing very close to the same thing.

schmidtk: I tried doing that, but that would require passing in the checkbox callback in addition to the default value in order to avoid the promise resolution API change. I didn't really like that solution, but I can go back to that if we want. Or I can deprecate the original function in favor of this one. I was trying to avoid passing around callbacks because the caller only cares about the end result.

jsalankey (approved):

PR #1083, opened 2020-12-01
feat: support Show Features in Places tab

bradh:
We previously supported this on the Layer view, but not in Places.

Implementation is closely modeled off Layer approach, but uses a different metric key.

image

### DIFF
--- a/src/os/metrics/metricskeys.js
+++ b/src/os/metrics/metricskeys.js
@@ -363,6 +363,7 @@ os.metrics.Places = {
   SAVE_TO: 'places.contextMenu.saveTo',
   EXPORT_CONTEXT: 'places.contextMenu.export',
   QUICK_ADD_PLACES: 'places.quickAdd',
+  FEATURE_LIST: 'places.featureList',
   REMOVE_PLACE: 'places.removePlace',
   REMOVE_ALL: 'places.removeAll'
 };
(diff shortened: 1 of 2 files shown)


schmidtk (approved):
Looks good, thanks!

One question - do we want to show the option if a place is selected, or just folders as implemented? I don't have a preference either way, just curious on your thoughts.

bradh:
> One question - do we want to show the option if a place is selected, or just folders as implemented? I don't have a preference either way, just curious on your thoughts.

I thought about whether it was an "any node" thing, and decided that if you have a place selected, then its probably more appropriate to just show Feature Info for that place. The alternative view is that the functionality might be more discoverable if it happened on every feature (although the context menu is already pretty full, so maybe not). 

I don't feel very strongly about it either way though, so happy to adapt as required.

schmidtk:
That makes sense, thanks! My inclination was that right-clicking a folder would only show features under that folder, but I don't think we have the plumbing to do that currently. We do this for the Export Places menu option, to allow exporting only part of the tree. Could be a useful future addition.

bradh:
Does the 'have the plumbing to do that' relate to there being no equivalent to `os.ui.launchFeatureList(source)` for a folder node?

I can add a ticket with that, if useful. Not sure when I'll have time, but that functionality could be useful to me in the future.

schmidtk:
Yes, that UI is based on the `sourcegrid` directive which interacts with a vector source. We would either need to filter the data, or change the backing model to work with a node. Or something smarter than those two suggestions.

PR #1085, opened 2020-12-01
feat(areas): save the id on the area so they can be ordered consistently

ericadeckl:
UCEP-777

### DIFF
--- a/src/os/ogc/filter/ogcspatialformatter.js
+++ b/src/os/ogc/filter/ogcspatialformatter.js
@@ -45,13 +45,14 @@ os.ogc.filter.OGCSpatialFormatter.prototype.format = function(feature) {
     // name/desc does not matter.
     var name = os.xml.escape(/** @type {string} */ (feature.get('title') || 'New Area'));
     var description = os.xml.escape(/** @type {string} */ (feature.get('description')));
+    var id = feature.getId() != null ? os.xml.escape(feature.getId().toString()) : undefined;
 
     switch (type) {
       case ol.geom.GeometryType.CIRCLE:
         geometry = /** @type {ol.geom.Circle} */ (geometry);
 
         var polyCircle = new ol.geom.Polygon([os.geo.interpolateCircle(geometry.getCenter(), geometry.getRadius())]);
-        result = os.ogc.spatial.formatGMLIntersection(polyCircle, this.column_, name, description) || '';
+        result = os.ogc.spatial.formatGMLIntersection(polyCircle, this.column_, name, description, id) || '';
         break;
       case ol.geom.GeometryType.MULTI_LINE_STRING:
       case ol.geom.GeometryType.POLYGON:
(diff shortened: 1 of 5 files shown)


src/os/ui/query/basicqueryreader.js
```diff
@@ -84,6 +84,9 @@ os.ui.query.BasicQueryReader.prototype.parseEntries_ = function(ele) {
       var title = os.xml.unescape(ele.getAttribute('areanamehint') || ele.getAttribute('namehint') || 'New Area');
       area.set('title', title);
       area.set('description', os.xml.unescape(ele.getAttribute('description')));
+      if (ele.getAttribute('id')) {
```
schmidtk, line 87: Could you verify this behavior in the existing test for this class?

src/os/ogc/filter/ogcspatialformatter.js
```diff
@@ -45,31 +45,32 @@ os.ogc.filter.OGCSpatialFormatter.prototype.format = function(feature) {
     // name/desc does not matter.
     var name = os.xml.escape(/** @type {string} */ (feature.get('title') || 'New Area'));
     var description = os.xml.escape(/** @type {string} */ (feature.get('description')));
+    var id = feature.getId();
```
schmidtk, line 48: I would recommend calling `os.xml.escape` on the id, in case it contains characters that are invalid in an XML attribute.

schmidtk (changes requested):

src/os/ui/query/basicqueryreader.js
```diff
@@ -108,6 +111,9 @@ os.ui.query.BasicQueryReader.prototype.parseEntries_ = function(ele) {
       var title = os.xml.unescape(ele.getAttribute('areanamehint') || ele.getAttribute('namehint') || 'New Area');
       excArea.set('title', title);
       excArea.set('description', os.xml.unescape(ele.getAttribute('description')));
+      if (ele.getAttribute('id')) {
+        excArea.setId(ele.getAttribute('id'));
```
jsalankey, line 115: The earlier `os.xml.escape` needs to be undone with an `os.xml.unescape` here.

test/os/ui/query/basicquery.xml
```diff
@@ -79,7 +79,7 @@
     <And>
       <And>
         <Or>
-          <Intersects description="Some Description" namehint="Some Area" xmlns="http://www.opengis.net/ogc">
+          <Intersects description="Some Description" namehint="Some Area" id="ID" xmlns="http://www.opengis.net/ogc">
```
jsalankey, line 82: It might be useful to include an escaped character here like `&amp;` to verify in the test that it gets brought back to just an `&`

ericadeckl: This is a great idea!

schmidtk (approved):

schmidtk:
All of my feedback is addressed and looks like @jsalankey's as well. I'll defer to him to verify.

PR #1088, opened 2020-12-02
feat(presets): support sharing presets among matching layerFilterKeys

glennkitchellcaci:
Fixed minor issues with Feature Action searches and filters in edge cases.
Used more robust layerpresetmanager.getPresets() to load all feature actions when adding a source instead of only the embedded loadDefaults().

### DIFF
--- a/src/plugin/featureaction/featureactionmanager.js
+++ b/src/plugin/featureaction/featureactionmanager.js
@@ -9,10 +9,10 @@ goog.require('os.data.event.DataEventType');
 goog.require('os.im.action.ImportActionCallbackConfig');
 goog.require('os.im.action.ImportActionManager');
 goog.require('os.implements');
+goog.require('os.layer.preset.LayerPresetManager');
 goog.require('os.source.IImportSource');
 goog.require('plugin.im.action.feature');
 goog.require('plugin.im.action.feature.Entry');
-
 
 
 /**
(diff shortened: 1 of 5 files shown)


src/plugin/featureaction/featureactionmanager.js
```diff
@@ -309,9 +309,15 @@ plugin.im.action.feature.Manager.prototype.addSource_ = function(source) {
       this.sourceListeners_[id] = ol.events.listen(/** @type {ol.events.EventTarget} */ (source),
           goog.events.EventType.PROPERTYCHANGE, this.onSourcePropertyChange_, this);
 
-      this.loadDefaults(id).thenAlways(function() {
+      var promise = os.layer.preset.LayerPresetManager.getInstance().getPresets(id, true);
```
jsalankey, line 312: I'm not sure I'm following this change... when is the `loadDefaults` call happening now? Is the preset manager now completely responsible for that?

glennkitchellcaci: Yeah, that "true" param makes it so when a new source is added, it'll look for and apply whatever "default=true" preset it detects - if it hasn't done so before (i.e. it'll rely on the User's current settings if it the layer/source has been added before).

jsalankey: The reason for my trepidation here is that the original intent of default feature actions in this manager was for them to be an independent piece of functionality that could hypothetically be loaded from anywhere (not just a preset)... Not sure if such a thing could still be supported just by looking at this PR.

glennkitchellcaci: ImportActionManager.loadDefaults(id) is called by SettingsPresetService through...
`lpm.getPresets(id)` >> `for presetservices...` >> `sps.find(id)` >> `iam.loadDefaults(id)`
There aren't settings.json configs such that the sps.find() wouldn't end up calling loadDefaults(). It's incredibly circuitous compared to "this.loadDefaults()"; but opens the door for Default FA's that aren't deployed with the code and/or settings.json like how loadDefaults() requires.

jsalankey (approved):

jsalankey (approved):

PR #1089, opened 2020-12-02
fix(noscroll): remove unused noscroll window option

swtyler:
Before Opensphere moved to bootstrap 4, we had a scroll "windowOption". That option was removed, or rather it no longer does anything, but the code was never cleaned up


michael-roschenwimmer (approved):
Looks good to me!

bradh:
This is a breaking (source incompatible) change for external plugins. It looks like you only need to remove the relevant line though.

PR #1090, opened 2020-12-02
Normalize 'NoBreakSpace' for File Input

Gamerlios123:
fix(nobreakspace): normalized white-space for file input

### DIFF
--- a/src/os/file/file.js
+++ b/src/os/file/file.js
@@ -118,7 +118,11 @@ os.file.File.prototype.getContent = function() {
  * @param {?(ArrayBuffer|Object|string)} value
  */
 os.file.File.prototype.setContent = function(value) {
-  this.content_ = value;
+  if (value instanceof ArrayBuffer) {
+    this.content_ = os.file.normalizeWhitespace(new TextDecoder('utf-8').decode(value));
+  } else {
+    this.content_ = value;
+  }
 };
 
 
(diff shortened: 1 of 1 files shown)


src/os/file/file.js
```diff
@@ -118,7 +118,11 @@ os.file.File.prototype.getContent = function() {
  * @param {?(ArrayBuffer|Object|string)} value
  */
 os.file.File.prototype.setContent = function(value) {
-  this.content_ = value;
+  if (value instanceof ArrayBuffer) {
+    this.content_ = os.file.normalizeWhitespace(new TextDecoder('utf-8').decode(value));
```
wallw-bits, line 122: You cannot assume UTF8 here. A better place for this call would be in `File.prototype.convertContentToString()`.

src/os/file/file.js
```diff
@@ -440,3 +444,14 @@ os.file.deserializeFile = function(data) {
...
+
+
+/**
+ * Normalizes all whitespace excluding newlines to std space
+ * @param {?(ArrayBuffer|string)} string
+ * @return {string}
+ */
+os.file.normalizeWhitespace = function(string) {
+  // Convert whitespace (-newlines) to standard space
+  return string.replace(/[^\S\n]/g, '\u0020');
```
wallw-bits, line 456: I do not think this is a good idea. Some files are tab-delimited. Let's discuss what you were trying to fix.

wallw-bits (changes requested):

wallw-bits:
I think @Gamerlios123 was seeing two separate issues, the first being caused by an encoding detection failure in `jschardet` (`"\u00a0" is not detecting as UTF-8), and the second being exacerbated by the first because it makes it more difficult to replace that character in whatever plugin if it gets decoded incorrectly. I will write that up the decoding issue separately and then any replacements can be done by plugin or application code.

PR #1091, opened 2020-12-03
feat(video): indicate if webgl renderer can render video in overlay

bcleveland1022:


jsalankey (approved):

src/plugin/cesium/cesiumrenderer.js
```diff
@@ -426,6 +426,12 @@ plugin.cesium.CesiumRenderer.prototype.renderSync = function() {
   }
 };
 
+/**
+ * @inheritDoc
+ */
+plugin.cesium.CesiumRenderer.prototype.supportsVideoOverlay = function() {
+  return false;
+};
```
jsalankey, line 434: This might be better if moved to abstractwebglrenderer.js. That would cover any class that extends it as a base implementer of the interface.

PR #1094, opened 2020-12-04
feat(map): allow rotation in 2d and don't reset when switching to 2d

schmidtk:
Rotation was disabled in 2D and reset when switching from 3D to 2D for performance reasons. This was largely due to inefficient tile rendering in Chrome, which has since been fixed. The canvas renderer still goes down a less efficient path, but works perfectly fine in many situations. It seems the preferred behavior is to maintain rotation when switching between 3D/2D, so the current view is maintained between the two.

This PR makes the following changes:

- Rotation is no longer reset when switching from 3D to 2D.
- The above can be toggled in Settings > Display if the old behavior is preferred.
- 3D rotation controls now work in 2D. This includes the keyboard (Shift + Left/Right Arrow) and mouse (Right/Middle Click + Drag) interactions.
- The above controls are documented in the Controls UI (Support > Controls).
- WebGL heading and OpenLayers rotation both use 0 for north, but the directions are opposite. This caused a few bugs we hadn't noticed. The rotate map control (top/right) will now rotate towards north (3D was rotating the wrong direction), and view persist/restore will no longer flip rotation direction in situations where 2D saved the view and 3D restored it.

[Test Deployment](https://noisy-arithmetic.surge.sh/)

### DIFF
--- a/src/os/control/rotatecontrol.js
+++ b/src/os/control/rotatecontrol.js
@@ -53,7 +53,8 @@ os.control.Rotate.render = function(mapEvent) {
       return;
     }
 
-    rotation = camera.getHeading();
+    // OpenLayers rotation is in the opposite direction of WebGL camera heading.
+    rotation = -camera.getHeading();
   } else {
     var frameState = mapEvent.frameState;
     if (!frameState) {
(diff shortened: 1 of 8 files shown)


src/os/control/rotatecontrol.js
```diff
@@ -53,7 +53,8 @@ os.control.Rotate.render = function(mapEvent) {
       return;
     }
 
-    rotation = camera.getHeading();
+    // OpenLayers rotation is in the opposite direction of WebGL camera heading.
+    rotation = -camera.getHeading();
```
glennkitchellcaci, line 57: Can/should this negative multiplier be renderer-based, or is this pretty much standard for a "heading" versus the corresponding projection transformation?

schmidtk: Internally the renderer can do what it wants, but the API should be standardized so switching renderers doesn't change inputs/outputs.

jsalankey (approved):

PR #1095, opened 2020-12-04
feat(filters): improve help text

glennkitchellcaci:
Added popovers and placeholders to text-based Filters to improve understanding which which Filter types have case-sensitivity and wildcard support.

### DIFF
--- a/src/os/ui/filter/op/islikeop.js
+++ b/src/os/ui/filter/op/islikeop.js
@@ -16,7 +16,7 @@ goog.require('os.xsd.DataType');
 os.ui.filter.op.IsLike = function() {
   os.ui.filter.op.IsLike.base(this, 'constructor',
       'PropertyIsLike', 'is like', 'like',
-      [os.xsd.DataType.STRING], 'matchCase="false" wildCard="*" singleChar="." escape="\\"',
+      [os.xsd.DataType.STRING], 'wildCard="*" singleChar="." escape="\\"',
       'e.g. abc*' + os.ui.filter.op.Op.TEXT.CASE_INSENSITIVE,
       undefined,
       undefined,
--- a/src/os/ui/filter/op/islikeop.js
+++ b/src/os/ui/filter/op/islikeop.js
@@ -15,8 +15,14 @@ goog.require('os.xsd.DataType');
  */
 os.ui.filter.op.IsLike = function() {
   os.ui.filter.op.IsLike.base(this, 'constructor',
-      'PropertyIsLike', 'is like', 'like', [os.xsd.DataType.STRING],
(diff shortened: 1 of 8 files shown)


bradh:
The travis failure looks related. Probably broke some assumption in the tests.

src/os/ui/filter/op/islikeop.js
```diff
@@ -15,8 +15,14 @@ goog.require('os.xsd.DataType');
  */
 os.ui.filter.op.IsLike = function() {
   os.ui.filter.op.IsLike.base(this, 'constructor',
-      'PropertyIsLike', 'is like', 'like', [os.xsd.DataType.STRING],
-      'wildCard="*" singleChar="." escape="\\"', 'e.g. abc*');
+      'PropertyIsLike', 'is like', 'like',
+      [os.xsd.DataType.STRING], 'matchCase="false" wildCard="*" singleChar="." escape="\\"',
```
jkae, line 19: https://github.com/ngageoint/opensphere/blob/044e0e8bea9a1e68b09b9f10b62ad4feb4314b87/test/os/ui/filter/op/islikeop.test.js#L15

glennkitchellcaci: > The travis failure looks related. Probably broke some assumption in the tests.

Ty ty.

And... for now I'm actually going to revert this. Looks like the "matchCase" attribute isn't for PropertyIsLike (a value comparison); only the binary comparisons. https://docs.geoserver.org/latest/en/user/styling/sld/reference/filters.html

src/os/ui/filter/op/op.js
```diff
@@ -283,3 +295,21 @@ os.ui.filter.op.Op.prototype.validate = function(value, key) {
...
+/**
+ * Text reused in several ops; to save on total the KB of the min file
+ * @enum {string}
+ * @const
+ */
+os.ui.filter.op.Op.TEXT = {
+  CASE_INSENSITIVE: '  (INFO: not case-sensitive)',
+  CASE_INSENSITIVE_DETAIL: '<strong>Case-insensitive</strong><sup>1</sup>, i.e. "A" == "a" and<br />' +
+  '<strong>Wildcard</strong> support, i.e. "a*" == "ab"<br /><br /><sup>1</sup>&nbsp;Typically, case' +
+  '-<i>in</i>sensitive text search is available for "like" filters. Some OGC Servers may not implement it.',
```
jsalankey, line 309: The reference to OGC servers here is not totally generic. This filter builder can be used for other server types. "Data Providers may not support it" would be a more generic description.

glennkitchellcaci: +1

jsalankey (approved):

PR #1102, opened 2020-12-07
Upgrade Toast UI

kwilson-bits:
Upgrade Toast UI from v1.4.6 to v2.5.X and make necessary changes to how we alter the settings and extract values out of the Toast Editor.

### DIFF
--- a/externs/tui-editor.externs.js
+++ b/externs/tui-editor.externs.js
@@ -3,42 +3,42 @@
  *
  * @externs
  */
-var tui = {};
+var toastui = {};
 
 
 /**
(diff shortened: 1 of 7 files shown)


csgibson (changes requested):
A preview of the built code is promising. Everything appears to work as well as before but will give us the ability to fix some lingering bugs in other plugins and add more toolbar options to do things like text color and highlight. 

A couple of notes on the PR:

1. Please squash the commits to a single one
2. Also add some basic info to the description

csgibson (approved):

PR #1107, opened 2020-12-08
feat(featureinfo): Added a new SimpleProperties Directive

glennkitchellcaci:
For data-dense Features, you can now configure the `<simple-properties>` Directive to additionally show the desired field(s) above the alphabetically-sorted Fields table. Configs allow for a per-field exact match and/or regex match(es). "Default" text is optional per-field as well.

### DIFF
--- a/src/os/ui/feature/simpleproperties.js
+++ b/src/os/ui/feature/simpleproperties.js
@@ -10,7 +10,6 @@ const OsSettings = goog.require('os.config.Settings');
 const Feature = goog.requireType('ol.Feature');
 
 
-
 /**
  * @type {!string}
  * @const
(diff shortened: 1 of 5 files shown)


src/os/ui/feature/simpleproperties.js
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * Add the directive to the module.
+ */
+OsModule.directive('simpleProperties', [directive]);
+
+exports = {
+  Controller,
+  directive,
+  SIMPLE_PROPERTIES_COLUMNS_KEY
```
jsalankey, line 243: Does this need to be exported? Doesn't look like anything here uses it.

src/os/ui/feature/simpleproperties.js
```diff
@@ -0,0 +1,244 @@
+goog.module('os.ui.feature.simpleproperties');
```
jsalankey, line 1: This module is just a standard UI/controller directive, and it should follow the naming conventions (i.e. `os.ui.feature.SimplePropertiesUI`).

src/os/ui/feature/simplepropertiesui.js
```diff
@@ -0,0 +1,243 @@
+goog.module('os.ui.feature.simplepropertiesUI');
```
jsalankey, line 1: Nittest of nits: the capitalization here is still not correct.

jsalankey (approved):

PR #1110, opened 2020-12-09
feat(chardetng-wasm): pulling chardetng-wasm package into Opensphere

Gamerlios123:
Test Steps:
- OS should build
- Test should run
- Binary files with `chardetng` are detected as plain/text.


wallw-bits:
Note, `chardetng-wasm` is not built with Emscripten and as such, it doesn't have the same fallback for non-streaming wasm compilation. You need your dev server to return `application/wasm` for `.wasm` files.

Here's the config for nginx:

```
http {
  types {
    application/wasm wasm;
  }
}
```

src/os/file/mime/text.js
```diff
@@ -65,22 +65,12 @@ os.file.mime.text.getText = function(buffer, opt_file) {
...
-      }
-
-      var encoding = jschardet.detect(binaryString).encoding;
+      var encoding = chardetng.detect(buffer);
       if (encoding) {
         // strip any UTF BOMs before decoding
         var boms = os.file.mime.text.BOMS_;
-        var n;
-        for (i = 0, n = boms.length; i < n; i++) {
+        for (var i = 0, n = boms.length; i < n; i++) {
```
csgibson, line 73: is `n` uninstantiated here?

Gamerlios123: No that's valid JS.
`var a = 5, b = 10, c = 15;`
`console.log('${a} ${b} ${c}');`
`Output: "5 10 15"`

csgibson: Ha, what an idiot; my bad, gotta stop doing these reviews before I eat.

test/os/file/mime/text.test.js
```diff
@@ -3,16 +3,6 @@ goog.require('os.file.mime.mock');
 goog.require('os.file.mime.text');
 
 describe('os.file.mime.text', function() {
```
glennkitchellcaci, line 5: Is this no longer a valid testing concern? If it doesn't work on "first N characters" anymore, can't you instead feed the file to this test?

package.json
```diff
@@ -315,7 +316,6 @@
     "google-closure-library": "^20200112.0.0",
     "html2canvas": "=1.0.0-rc.4",
     "jquery": "3.4.1",
```
glennkitchellcaci, line 318: Since it's wasm, I'm sure it's fine, but please resolve this merge and double-check that jQuery 3.5 doesn't break anything for the chardetng-wasm.

wallw-bits: It does not.

test/os/file/mime/text.test.js
Gamerlios123: This test was removed since chardetng won't return a false value to satisfy the test. Also I'm not sure I follow what your suggesting here..

glennkitchellcaci: My thought here was that there could be another way to test the same thing as the old test. If not, no big deal.

package.json
glennkitchellcaci: I fixed the merge conflict instead. And yup, I'd be more than a little surprised if _any_ wasm library had a dependency on jquery.

glennkitchellcaci (approved):

PR #1111, opened 2020-12-10
WIP: feat(adddata): show additional properties in layer description

glennkitchellcaci:
Reads from the AccessConstraints (an OGCServer standard in getCapabilities response) and shows additional data to User in the Add Data modal; as configured in settings.json.

### DIFF
--- a/src/os/data/basedescriptor.js
+++ b/src/os/data/basedescriptor.js
@@ -747,7 +747,7 @@ os.data.BaseDescriptor.prototype.restore = function(from) {
  * @inheritDoc
  */
 os.data.BaseDescriptor.prototype.getHtmlDescription = function() {
-  var linebreak = '<br />';
+  var linebreak = '<br>';
 
   var propertiesHTML = null;
   var properties = this.getProperties();
@@ -792,7 +792,7 @@ os.data.BaseDescriptor.prototype.getHtmlDescription = function() {
 
   var desc = this.getDescription();
   text.push((desc ? desc : 'No description provided') + linebreak + linebreak);
-  text.push('Tags: ' + (this.getTags() ? this.getTags().join(', ') : '(none)') + linebreak);
+  text.push('Tags: ' + (this.getTags() ? this.getTags().join(', ') : '(none)') + linebreak + linebreak);
 
   return text.join('');
 };
--- a/src/os/data/basedescriptor.js
+++ b/src/os/data/basedescriptor.js
@@ -136,6 +136,12 @@ os.data.BaseDescriptor = function() {
   this.tags = null;
 
   /**
+   * @type {?Array<string>}
+   * @protected
+   */
+  this.properties = null;
+
+  /**
    * @type {!string}
    * @protected
    */
@@ -607,6 +613,24 @@ os.data.BaseDescriptor.prototype.setTags = function(value) {
 
 
 /**
+ * Gets the Properties HTML
+ * @return {?Array<string>}
+ */
+os.data.BaseDescriptor.prototype.getProperties = function() {
+  return this.properties;
+};
+
+
+/**
+ * Sets the Properties XML
+ * @param {?Array<string>} value
+ */
+os.data.BaseDescriptor.prototype.setProperties = function(value) {
+  this.properties = value;
+};
+
+
+/**
  * @inheritDoc
  */
 os.data.BaseDescriptor.prototype.getSearchText = function() {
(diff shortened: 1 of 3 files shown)


src/os/data/basedescriptor.js
```diff
@@ -721,18 +747,35 @@ os.data.BaseDescriptor.prototype.restore = function(from) {
...
+
+  var text = [];
+
+  if (propertiesHTML) {
+    text.push(propertiesHTML);
+    text.push(linebreak);
+    text.push(linebreak);
+  }
+
+  text.push('Layer Name: ' + this.getTitle() + linebreak);
```
jsalankey, line 766: Putting the name anywhere but at the top seems weird to me.

bcleveland1022 (approved):

src/os/ui/ogc/ogcserver.js
```diff
@@ -821,6 +843,64 @@ os.ui.ogc.OGCServer.prototype.displayConsentAlerts = function(result) {
...
+    try {
+      // clean up the HTML-encoded strings, e.g. &gt;
+      xmlDocument = (new DOMParser()).parseFromString(goog.string.unescapeEntities(constraints), 'text/xml');
+    } catch (e) {
+      goog.log.error(os.ui.ogc.OGCServer.LOGGER_, 'Could not parse XML for OGCServer additional information', e);
+    }
+  }
+
+  try {
+    var setting = /** @type {Object<?, osx.ogc.ServerProperty>} */ (os.config.Settings.getInstance()
```
jsalankey, line 868: `?` should never be a valid key type for an object. Object keys are always strings (even if you try to make them numbers, they are strings).

src/os/ui/ogc/ogcserver.js
```diff
@@ -821,6 +843,64 @@ os.ui.ogc.OGCServer.prototype.displayConsentAlerts = function(result) {
...
+        .get(os.ui.ogc.OGCServer.DESCRIPTION_KEY, {}));
+    if (setting) {
+      var configs = Object.entries(setting);
+
+      // Display info
+      configs.forEach(([key, config]) => {
+        // help out the compiler
+        key = /** @type {string} */ (key);
+        config = /** @type {osx.ogc.ServerProperty} */ (config);
```
jsalankey, line 877: The compiler shouldn't need this help. Breaking out the object into an an array with the above `Object.entries` call also seems unnecessary.

src/os/ui/ogc/ogcserver.js
glennkitchellcaci: The compiler didn't like it when it was a string though... I'll see what other things I can do to trick it.

src/os/ui/ogc/ogcserver.js
```diff
@@ -821,6 +843,64 @@ os.ui.ogc.OGCServer.prototype.displayConsentAlerts = function(result) {
...
+/**
+ * Extract relevant information from the GetCapabilities result for display in Descriptors, HoverText, etc
+ * @param {string} constraints
+ * @return {?Array<string>}
+ */
+os.ui.ogc.OGCServer.prototype.getProperties = function(constraints) {
+  var properties = [];
+  var xmlDocument = null;
+
+  if (constraints && constraints.indexOf('xml version') && window.DOMParser) {
```
jsalankey, line 855: is there a specific reason for calling to `DOMParser` here? `os.xml.loadXml` is our typical call to parse a document from XML. the `os.xml` namespace also has a bunch of other utility functions for XML (like escaping).

glennkitchellcaci: In this doc it was all `goog.dom.xml`; so I just was using the "vanilla JS" way for this new code. I'll update to use the `os.xml` in the final version.

src/os/ui/ogc/ogcserver.js
```diff
@@ -983,6 +1061,11 @@ os.ui.ogc.OGCServer.prototype.parseWfsCapabilities = function(response, uri) {
         descriptor.setKeywords(tags);
       }
 
+      if (descriptor instanceof os.data.BaseDescriptor) {
```
schmidtk, line 1064: Should we add the new functions to `IDataDescriptor` so we don't need this `instanceof` check? It's a breaking change, but seems cleaner than this.

schmidtk, line 1064: This feels like it's working around API design a bit. Should the new functions be added to an interface instead of using `instanceof`? I'm not sure if they're intended to be specific to OGC, making `IOGCDescriptor` an appropriate place for them, or if other descriptors may also benefit which would make `IDataDescriptor` a better choice.

schmidtk:
This is OBE / no longer needed. Closing.

PR #1118, opened 2020-12-14
feat(toastui): upgrade toast ui and integration

kwilson-bits:
Upgrade Toast UI from v1.4.6 to v2.5.X and make necessary changes to how we alter the settings and extract values out of the Toast Editor.

### DIFF
--- a/scss/os/_overrides_tuieditor.scss
+++ b/scss/os/_overrides_tuieditor.scss
@@ -94,6 +94,8 @@
   //////////// WYSIWYG / DISPLAY STYLES //////////
   // Instead of adding the contents and overriding them, just use the ones we want
   .tui-editor-contents {
+    color: $body-color;
+
     *:not(table) {
       line-height: $line-height-base;
     }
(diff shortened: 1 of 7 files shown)


scss/os/_overrides_tuieditor.scss
```diff
@@ -102,18 +102,121 @@
...
+    }
+
+    h6 {
+      font-size: .875rem;
+      font-weight: normal;
+    }
+
+    @media (max-width: 1200px) {
+      h1 {
+        font-size: calc(1.34375rem + 1.125vw);
```
gregroyal, line 166: These are some very specific values. 1.34375 ?

scss/os/_overrides_tuieditor.scss
```diff
@@ -499,10 +602,41 @@
...
 .tui-tooltip {
   z-index: $zindex-tooltip;
 }
+
+.tui-md-code {
+  background: none;
+}
+
+.tui-md-code.tui-md-marked-text {
+  color: #f89406;
```
gregroyal, line 625: prefer variables over strings for colors

scss/os/_overrides_tuieditor.scss
kwilson-bits: I couldn't find a variable or calculation for any of the header sizes, and actually ended up grepping them out of the `.build/themes` files to find what was used. They match what was compiled before the Toast UI upgrade. The problem I ran into is that the updated Toast UI has a newer stylesheet that mucks with a lot of our theming.

src/os/ui/text/tuieditorlang.js
```diff
@@ -8,24 +8,22 @@ goog.provide('os.ui.text.TuiEditorLang');
...
-    // options['Insert CodeBlock'] = options['Insert CodeBlock'] + ' (shift + ctrl + p)';
-    options['Italic'] = options['Italic'] + ' (ctrl + i)';
-    options['Line'] = options['Line'] + ' (ctrl + l)';
-    options['Ordered list'] = options['Ordered list'] + ' (ctrl + o)';
-    options['Strike'] = options['Strike'] + ' (ctrl + s)';
-    options['Task'] = options['Task'] + ' (alt + t)';
-    options['Unordered list'] = options['Unordered list'] + ' (ctrl + u)';
+    languageOverrides['Bold'] = 'Bold (ctrl + b)';
+    languageOverrides['Blockquote'] = 'Blockquote (alt + q)';
+    languageOverrides['Code'] = 'Inline code (shift + ctrl + c)';
```
csgibson, line 19: Do we need an override for the code block too?
I know it was commented out in the old version but may be fixed in the new version? idk

scss/os/_overrides_tuieditor.scss
kwilson-bits: Color codes have been replaced with variables.

src/os/ui/text/tuieditorlang.js
kwilson-bits: It hasn't changed for the new version, so I would leave the language override out. While the shortcut still works in Chrome it still does _not_ work in Firefox.

csgibson (approved):

PR #1121, opened 2020-12-18
fix(export): add options for area and places

arwatts16:
Creates an optional dialog for additional export options for both export dialog and kml tree export. Allows the user to export all, selected, or active data.

### DIFF
--- a/src/plugin/places/ui/placesui.js
+++ b/src/plugin/places/ui/placesui.js
@@ -3,6 +3,7 @@ goog.provide('plugin.places.ui.placesDirective');
 
 goog.require('goog.Disposable');
 goog.require('os.defines');
+goog.require('os.ex.ExportOptions');
 goog.require('os.metrics.Metrics');
 goog.require('os.metrics.keys');
 goog.require('os.ui.Module');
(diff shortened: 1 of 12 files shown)


src/os/ui/ex/addexportoptions.js
```diff
@@ -0,0 +1,86 @@
+goog.provide('os.ui.addExportCtrl');
+goog.provide('os.ui.addExportDirective');
```
glennkitchellcaci, line 2: New angular directives should be made in the new goog.module() format. You can just...
1. Run the es6-transform on this file,
2. Delete the shim that it makes, 
3. Rename the file to "addexportoptionsui.js" and the module to "AddExportOptionsUI", (to follow the new-ish naming convention), and
4. Make a small tweak to the goog.require's elsewhere that pull in the directive

src/plugin/places/ui/placesui.js
```diff
@@ -139,14 +139,50 @@ plugin.places.ui.PlacesCtrl.prototype.hasRoot = function() {
  */
 plugin.places.ui.PlacesCtrl.prototype.export = function() {
   if (this.placesRoot_) {
-    plugin.file.kml.ui.launchTreeExport(this.placesRoot_, 'Export Places');
+    var activePlaces = this.getActivePlaces(this.placesRoot_);
+
+    var options = /** @type {os.ex.ExportOptions} */ ({
```
glennkitchellcaci, line 144: Since these options are not used directly by the angular directive (it looks like you're copying them from `scope.options['blah']` to `controller['blah']` in all cases), you can:
1. Modify the /src/os/ex/export.js file to have these properties in the typedef and
2. Remove the quotes around all uses, e.g. `controller['blah'] = scope.options.blah;` instead of `controller['blah'] = scope.options['blah'];` so that the code can minify.

As a side note, "fields" and "items" are properties currently required by os.ex.ExportOptions, so I'm surprised that this builds... interesting. Make sure to look at existing uses of os.ex.ExportOptions to make sure that you're reusing the typedef in a good way.

glennkitchellcaci (commented):
Looks good so far. I suggest a couple points of cleanup to make use of the code minify-er and the new module format for new files.

glennkitchellcaci (approved):

glennkitchellcaci (approved):

schmidtk:
@arwatts16 & @glennkitchellcaci It looks like this change is causing errors when exporting data, and in some cases prevents export from working entirely.

PR #1122, opened 2021-01-04
fix(ui): use <button> rather than <span> with button role

bradh:
Fixes linter warning on build


schmidtk:
I addressed this in #1137 using the pattern recommended by Bootstrap. It looked like changing the tabs to use `button` resulted in some unwanted highlight borders when focused.

bradh:
Thanks

PR #1125, opened 2021-01-05
fix(offscreenpopup): change popup position to absolute

Gamerlios123:


lilysikes (approved):

csgibson (approved):
Tested the fix in opensphere too, looks fine.

PR #1128, opened 2021-01-11
WIP: feat(toastui): add support for optional plugins

kwilson-bits:


bradh (commented):
Is there an example for how to use this extension point?

A nice fully worked tutorial would be glorious, but even some inline code snippets might be very useful.

csgibson:
reworked a little and moved to https://github.com/ngageoint/opensphere/pull/1133

PR #1129, opened 2021-01-12
Modify interaction

jsalankey:
Adds an adapted version of the Modify Interaction provided by Openlayers. This interaction allows for dynamically updating geometries on the map with clicking and dragging. The 3D sync of the interaction is accomplished through the Drawing Layer and new dynamic synchronizers rather than the traditional method used for other pointer interactions (which set update/cleanup functions). This pattern should be cleaner and generally more easily adaptable to other use cases.

There are a few little quirks I still need to iron out:
- The vertex feature gets stuck sometimes in 2D.
- The alt-click interaction is a little funky and doesn't seem to work sometimes when it should. Might be an OL issue.
- Interpolated geometries don't modify as expected due to the intermediate coordinates. This is an underlying issue that might not be fixable in this issue.

### DIFF
--- a/src/os/interaction/modifyinteraction.js
+++ b/src/os/interaction/modifyinteraction.js
@@ -5,10 +5,11 @@ const I3DSupport = goog.require('os.I3DSupport');
 const KeyCodes = goog.require('goog.events.KeyCodes');
 const KeyHandler = goog.require('goog.events.KeyHandler');
 const OLModify = goog.require('ol.interaction.Modify');
+const PayloadEvent = goog.require('os.events.PayloadEvent');
 const osImplements = goog.require('os.implements');
 const {MODAL_SELECTOR} = goog.require('os.ui');
+const {ModifyEventType} = goog.require('os.interaction');
 const {notifyStyleChange} = goog.require('os.style');
-const PayloadEvent = goog.require('os.events.PayloadEvent');
 
 const KeyEvent = goog.requireType('goog.events.KeyEvent');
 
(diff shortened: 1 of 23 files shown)


src/os/interaction/modifyinteraction.js
```diff
@@ -0,0 +1,258 @@
...
+
+    this.keyHandler.listen(KeyHandler.EventType.KEY, this.handleKeyEvent, true, this);
+
+    // jank alert: the functions that are called when the interaction starts and ends are hard to override, so instead
+    // listen to our own events and toggle the map movement on and off
+    ol.events.listen(this, ol.interaction.ModifyEventType.MODIFYSTART, this.handleStart, this);
+    ol.events.listen(this, ol.interaction.ModifyEventType.MODIFYEND, this.handleEnd, this);
+
+    // use the drawing layer as the overlay layer so that geometries are synced to WebGL automatically
+    this.overlay_ = os.map.mapContainer.getDrawingLayer();
```
glennkitchellcaci, line 58: Nit: missing require() and .getInstance()

src/os/ui/help/controlblock.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  /**
+   * Get the key text
+   * @param {string} other
+   * @return {?string}
+   * @export
+   */
+  getFont(other) {
+    var font = os.ui.help.Controls.FONT_CLASS[other];
```
glennkitchellcaci, line 103: Nit: require()

test/plugin/cesium/sync/dynamicmultipolygonconverter.test.js
```diff
@@ -0,0 +1,87 @@
...
+        [[0, 0], [0, 5], [5, 5], [5, 0], [0, 0]]
+      ],
+      [
+        [[10, 10], [10, 20], [20, 20], [20, 10], [10, 10]]
+      ]
+    ];
+    geometry = new ol.geom.MultiPolygon(coords);
+    feature = new os.feature.DynamicFeature(geometry);
+    style = new ol.style.Style();
+    layer = new os.layer.Vector();
```
glennkitchellcaci, line 35: Nit: A couple more require()'s os.layer, os.feature

glennkitchellcaci (changes requested):
Found a few os namespace things to require() in. Hopefully it doesn't create any circular dependencies.

src/os/interaction/modifyinteraction.js
schmidtk: `ol.events` and `ol.interaction.ModifyEventType` should also be referenced by require.

jsalankey: I'll add a comment here. Requiring the `MapContainer` was creating a circular dependency somewhere, which was why I went to the sucky reference. I take a quick look to see if I can untangle it, but that's a big mess for later.

test/plugin/cesium/sync/dynamicmultipolygonconverter.test.js
jsalankey: These are all required correctly. The base OL classes and `DynamicFeature` are not modules yet and so I don't think they can be included with `goog.module.get`.

schmidtk (approved):

schmidtk (approved):

PR #1131, opened 2021-01-13
feat(docked-window): Add basic support for a docked window

glennkitchellcaci:
Basic changes to os.ui.window for supporting a docked window.


views/main.html
```diff
@@ -5,7 +5,8 @@
       <div list list-id="main-content" list-prefix="<div>" list-suffix="</div>"></div>
       <map></map>
     </div>
-    <timeline-panel ng-if="mainCtrl.timeline"></timeline-panel>
+    <timeline-panel ng-if="mainCtrl.timeline"></timeline-panel><!-- TODO move into dock-bottom? -->
```
bcleveland1022, line 8: is there a ticket for that todo?

glennkitchellcaci: I started looking at putting the timeline into the new "docked window" specifically in the dock-bottom area, but it looks like it'd take some enhancements (resize and some of the settings pass-through's) to be able to support it. I can either remove the comment or leave it around as a reminder for when/if those features get added.

bcleveland1022: you can leave it.

bcleveland1022 (dismissed):

bcleveland1022 (dismissed):

bcleveland1022 (approved):

PR #1133, opened 2021-01-13
feat(toastui): add support for optional plugins

kwilson-bits:
Using this as a starting point:
https://nhn.github.io/tui.editor/latest/tutorial-example10-editor-with-color-syntax-plugin

### DIFF
--- a/externs/tui-editor.externs.js
+++ b/externs/tui-editor.externs.js
@@ -14,16 +14,15 @@ toastui.EventManager = function() {};
 
 /**
  * @param {string} eventName
- * @param {function()} cb
+ * @param {function(string)} cb
  */
 toastui.EventManager.prototype.listen = function(eventName, cb) {};
 
 
 /**
  * @param {string} eventName
- * @param {function()} cb
  */
-toastui.EventManager.prototype.addEventType = function(eventName, cb) {};
+toastui.EventManager.prototype.addEventType = function(eventName) {};
 
 
 /**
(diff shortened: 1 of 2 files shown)


src/os/ui/text/tuieditor.js
```diff
@@ -287,6 +287,12 @@ os.ui.text.TuiEditorCtrl.prototype.getOptions = function() {
     options['toolbarItems'] = toolbarItems;
   }
 
+  const plugins = this.getPlugins();
+
+  if (plugins instanceof Array && plugins.length > 0) {
```
csgibson, line 292: This might be better as `Array.isArray(plugins)`

csgibson (commented):
nitpick feedback

externs/tui-editor.externs.js
```diff
@@ -14,16 +14,15 @@ toastui.EventManager = function() {};
 
 /**
  * @param {string} eventName
- * @param {function()} cb
```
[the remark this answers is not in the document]
kwilson-bits: It would not build with the parameter as just `function()` so I simplified it to just `*`. Looking at it again, I changed it to `function(string)` to better define what is being passed in, and it builds again.

src/os/ui/text/tuieditor.js
kwilson-bits: The first check has been updated to use `Array.isArray()`

kwilson-bits: Actually, it looks to work just fine with empty arrays. I changed it to just be the one check.

csgibson (approved):

PR #1134, opened 2021-01-14
Feat datepicker shortcuts

ameletyan:


src/os/ui/datetime/datecontrol.js
```diff
@@ -165,6 +174,14 @@ os.ui.datetime.DateControlCtrl.prototype.onStartDateChanged_ = function(newVal,
 
     goog.log.fine(os.ui.datetime.DateControlCtrl.LOGGER_,
         'start changed: ' + this['startDate'].toUTCString() + ' to ' + this['endDate'].toUTCString());
+
+    // open the datepicker for the end date
+    const endDatePicker = $('#timelineEndDate');
```
schmidtk, line 179: Find elements relative to the directive's root element instead of a global jQuery search. It's both faster and more specific to avoid finding a matching element elsewhere on the page. To do this, add `$element` to the constructor params and assign to `this.element_`, then use `this.element_.find(<selector>)` instead of `$(<selector>)`. The pattern for this can be found in other directives, `os.ui.FeatureListCtrl` is one example.

src/os/ui/datetime/datecontrol.js
```diff
@@ -165,6 +174,14 @@ os.ui.datetime.DateControlCtrl.prototype.onStartDateChanged_ = function(newVal,
...
+    if (endDatePicker && this['showEndDateCalendar']) {
```
schmidtk, line 180: This flag is a bit fragile because the start date may be changed outside of this directive. For example, if you open the timeline and change the yellow load range the end date picker will open.

Instead of using a flag to determine when the end date should open, find a way to notify this controller when the user specifically selects a date in the start date picker.

views/datetime/datecontrol.html
```diff
@@ -3,17 +3,18 @@
...
       <button class="btn btn-secondary flex-shrink-0" title="Previous {{dateControl.duration}}"
           ng-click="dateControl.shiftDate(-1)"
-          ng-disabled="dateControl.disabled"
-          ng-if="dateControl.duration !== 'custom'">
+          ng-disabled="dateControl.disabled">
         <i class="fa fa-chevron-left"></i>
       </button>
     </div>
-    <wheel-date title="Start date"
+    <wheel-date id="timelineStartDate"
```
schmidtk, line 10: Prefer using a specific `class` selector instead of an `id`. Element ID's are expected to be unique within a page, but this can't be enforced. Instead we use a class that both implies it's used as a selector in JavaScript, and is descriptive of the directive it belongs to. For example, this might use `class="js-datecontrol__startDate"` and `class="js-datecontrol__endDate".

This combined with finding the selector from the directive's root element are a more reliable way to protect against finding the wrong element on the page.

src/os/time/time.js
```diff
@@ -606,6 +623,18 @@ os.time.round = function(date, duration, opt_roundDown) {
         testDate.setUTCDate(testDate.getUTCDate() + 1);
       }
       break;
+    case 'last 7 days':
```
schmidtk, line 626: I'm not certain this is the best place to handle setting up these relative ranges. `os.time.round` should support rounding up and down, and I also don't know that rounding makes sense here because we want dates to be relative to "now" versus "today at 00:00.000".

It might be better for `onDurationChanged` to have specific handling for relative dates, where the `endDate` is set to now and `startDate` is set to the end date minus the appropriate duration. 

One issue with this is it would set the range at the time the user changes the duration, but would remain fixed to that specific range. If someone is using the application for several hours and later makes a query, they will be querying data relative to when they changed the duration not the current time. We may need to handle this within the timeline controller instead of setting a static range once within the date control.

src/os/time/time.js
```diff
@@ -25,6 +25,9 @@ os.time.Duration = {
   WEEK: 'week',
   MONTH: 'month',
   YEAR: 'year',
+  LAST7DAYS: 'last 7 days',
```
schmidtk, line 28: I believe we want last 24/48 hours as options. With dense vector data these would be the most common selections.

schmidtk (changes requested):

src/os/ui/datetime/datecontrol.js
```diff
@@ -220,6 +254,20 @@ os.ui.datetime.DateControlCtrl.prototype.onDurationChanged = function() {
...
 
+/**
+ * Set the duration to a date range relative to now.
+ *
+ * @param {number} days
+ * @private
+ */
+os.ui.datetime.DateControlCtrl.prototype.setRelativeDateRange = function(days) {
+  this['relativeDuration'] = true;
+  os.time.resetDate(this['startDate']);
```
schmidtk, line 265: I think we want to initialize both dates to `new Date()` so it's actually using "now" versus the start of the current day, then adjust `startDate` by the number of days. After making that change, you should be able to remove `os.time.resetDate` as it will no longer be used.

src/os/ui/datetime/datecontrol.js
```diff
@@ -101,13 +105,23 @@ os.ui.datetime.DateControlCtrl = function($scope) {
...
+  this['relativeDuration'] = false;
+
   // take over updating the timeline controller
   this.assumeControl();
 
   // watch for changes to start/end dates
   $scope.$watch('dateControl.startDate', this.onStartDateChanged_.bind(this));
   $scope.$watch('dateControl.endDate', this.onEndDateChanged_.bind(this));
 
+  $scope.$on('startDate.userSelected', function(e) {
```
schmidtk, line 121: Could you create a more formal handler for this? Something like `this.onStartDateSelected_`? The JSDoc type for Angular events is `angular.Scope.Event`, for the function's docs.

src/os/ui/datetime/datecontrol.js
```diff
@@ -205,12 +219,32 @@ os.ui.datetime.DateControlCtrl.prototype.onDurationChanged = function() {
...
+        this.setRelativeDateRange(7);
+        break;
+      case os.time.Duration.LAST14DAYS:
+        this.setRelativeDateRange(14);
+        break;
+      case os.time.Duration.LAST30DAYS:
+        this.setRelativeDateRange(30);
+        break;
+      case os.time.Duration.CUSTOM:
+        this['relativeDuration'] = false;
```
schmidtk, line 239: I would consider setting `this['relativeDuration'] = false;` prior to the `switch` statement and letting `setRelativeDateRange` change it back. That would avoid the need to add that line for any non-relative duration.

schmidtk (changes requested):

src/os/ui/datetime/datecontrol.js
ameletyan: Ah right, we even talked about that this morning! Thanks for being vigilant.

schmidtk (changes requested):
There are some issues related to how `os.layer.AnimatedTile` handles durations. There are a few situations where it has special handling for `os.time.Duration.CUSTOM` versus non-custom durations, but it now also needs to consider how to handle relative durations. This is causing tile requests with a time dimension to create an incorrect time parameter.

To reproduce, add the following GeoServer:
https://demo.geo-solutions.it/geoserver/ows

Then load the `black_marble` layer and select a relative time and look at the tile requests in the Network tab. You'll see requests like this for "Last 24 Hours":

```
TIME: 2021-01-26T00:42:12Z/2021-01-26T12:42:12Z
```

This is using the non-custom path in `os.layer.AnimatedTile.getTimeParameter` to create the parameter, but for relative dates we want the same behavior as custom. There are a few other places in that class where we should ensure the correct logic is applied for relative dates. We might want a `os.time.isRelativeDuration` convenience function to assist in determining this case.

src/os/time/time.js
```diff
@@ -896,3 +901,36 @@ os.time.combineDateTime = function(days, time, opt_daysFormat = 'YYYY-MM-DD', op
...
+
+
+/**
+ * Determines whether a duration is relative or not
+ *
+ * @param {string} duration The duration that will be analyzed
+ * @return {boolean} Whether the duration is relative or not
+ */
+os.time.isRelativeDuration = function(duration) {
+  var isRelativeDuration = false;
```
schmidtk, line 913: I'm wondering if we should make an assumption about the duration string to simplify this function and future maintenance. If we assume relative durations will always be "last xx", we can make this function dead simple:

```
/**
 * Regular expression to test for a relative duration string.
 * @type {RegExp}
 */
os.time.RELATIVE_DURATION_REGEXP = /^last /i;

os.time.isRelativeDuration = function(duration) {
  return os.time.RELATIVE_DURATION_REGEXP.test(duration);
};
```

Then if the relative options change in the future, we won't have to update this function assuming they follow the same naming scheme.

schmidtk (changes requested):

schmidtk:
I found/fixed an issue concerning how we translate dates to/from UTC when passing values between jQueryUI's date picker and the timeline controller. This is due to the application considering dates in UTC and jQueryUI using local dates. For relative dates, we're creating the date objects ourselves and they do not need to be translated.

The only remaining issue is how we want to handle reading/writing these relative durations in state files. I think we may need to treat them as custom durations, so if a state is saved and loaded at a later time the same data will be displayed.

src/os/ui/datetime/datecontrol.js
```diff
@@ -262,9 +331,15 @@ os.ui.datetime.DateControlCtrl.prototype.updateController_ = function() {
...
  */
 os.ui.datetime.DateControlCtrl.prototype.shiftDate = function(direction) {
+  var modifier = 1;
   if (!this['disabled']) {
-    this['startDate'] = os.time.offset(this['startDate'], this['duration'], direction, true);
-    this['endDate'] = os.time.offset(this['endDate'], this['duration'], direction, true);
+    if (this['duration'] === os.time.Duration.CUSTOM) {
+      // For custom durations, multiply the offset by the difference in days between the start and end dates
+      const millisecondsPerDay = 1000 * 60 * 60 * 24;
+      modifier = (this['endDate'] - this['startDate']) / millisecondsPerDay;
```
schmidtk, line 339: The end date is inclusive (end of the UI's date) while start is exclusive (beginning of the UI's date), so I believe we want to add a day to the end date for this computation. I'm also not sure on the browser support for directly subtracting `Date` objects, so we may want to use `<date>.getTime()` instead to make sure this works as expected.

The result would be something like:
```
const startTime = this['startDate'].getTime();
const endTime = this['endDate'].getTime() + millisecondsPerDay;
modifier = (endTime - startTime) / millisecondsPerDay;
```
This should fix the current issue with the forward/back buttons not quite working as expected.

schmidtk (changes requested):
This is looking good! One last comment and we should be good.

src/os/ui/datetime/datecontrol.js
```diff
@@ -262,9 +331,17 @@ os.ui.datetime.DateControlCtrl.prototype.updateController_ = function() {
  * @export
  */
 os.ui.datetime.DateControlCtrl.prototype.shiftDate = function(direction) {
+  var modifier = 1;
   if (!this['disabled']) {
-    this['startDate'] = os.time.offset(this['startDate'], this['duration'], direction, true);
-    this['endDate'] = os.time.offset(this['endDate'], this['duration'], direction, true);
+    if (this['duration'] === os.time.Duration.CUSTOM) {
+      // For custom durations, multiply the offset by the difference in days between the start and end dates
+      const millisecondsPerDay = 1000 * 60 * 60 * 24;
```
jsalankey, line 338: Nit: we have a constant for this (and many similar static durations) in time.js: `os.time.millisecondsInDay`.

schmidtk (approved):

PR #1139, opened 2021-01-19
refactor(karma): use webpack to build karma test dependencies

schmidtk:
**Draft TODO:**

- [x] Release [`@ngageoint/opensphere-coverage-loader`](https://www.npmjs.com/package/@ngageoint/opensphere-coverage-loader)
- [x] Merge/release ngageoint/eslint-plugin-opensphere/pull/13
- [x] Merge/release ngageoint/opensphere-build-resolver/pull/74

By bundling all test dependencies with webpack, both `goog.module` and ES6 modules can be supported without specific loaders or Karma config. The test webpack config uses `@ngageoint/opensphere-coverage-loader` to instrument the webpack output with coverage code, which is supplemented by `karma-coverage-istanbul-reporter` to generate reports.

`@ngageoint/opensphere-coverage-loader` is a fork of [istanbul-instrumenter-loader](https://github.com/webpack-contrib/istanbul-instrumenter-loader) that updates the `istanbul-lib-instrument` dependency to fix a bug with how source output is written after adding coverage. With the old version, `class Child extends Parent` would be rewritten as `class Child extends (Parent)`, which causes problems with Angular's argument injection code.


jsalankey:
Couple questions:
1. Would we need a similar change to how the tests load the `src/**.js` files if we wanted to include ES6 modules into those?
2. Is this generating and loading the webpack bundle with source maps back to the original test files so that debugging is easier (similar to the webpack debug app build)?

package.json
```diff
@@ -248,6 +249,7 @@
     "@commitlint/config-conventional": "^8.1.0",
     "@commitlint/travis-cli": "^8.1.0",
     "@ngageoint/closure-webpack-plugin": "^1.0.0",
+    "@ngageoint/istanbul-instrumenter-loader": "^1.0.0",
```
jsalankey, line 252: Is this a new repository/npm package for us to manage?

schmidtk: Yes. The original is linked in the description and is archived/no longer managed. Code-wise it's extremely small and shouldn't change, we just need to update a dependency (`istanbul-lib-instrument`) to fix an issue with classes and Angular.

schmidtk:
> 1. Would we need a similar change to how the tests load the `src/**.js` files if we wanted to include ES6 modules into those?

We can already use ES6 modules in `src` files, but they would fail to load in tests because the loader does not work with modules. This fixes that, and removes the need for the loader entirely. We can now create ES6 modules in `src` and test mocks. The `*.test.js` files will not be modules. They don't export anything anyway, and this would require additional changes with Karma that aren't necessary.

> 2. Is this generating and loading the webpack bundle with source maps back to the original test files so that debugging is easier (similar to the webpack debug app build)?

Yes it is. It also seems to result in more accurate coverage than the old plugin, I believe due to conflicts between the `coverage` and `googmodule` preprocessors. Using webpack removes the need for the `googmodule` preprocessor because we're producing an ES5 output file for Karma to load (instead of original source files), and we no longer need the `coverage` preprocessor because that is handled in webpack via `istanbul-instrumenter-loader`.

wallw-bits (approved):

PR #1146, opened 2021-01-21
feat(slicktreestyle): improve interface for styling slickTree

Gamerlios123:
Needed for a client ticket adding Object Type Counts to all Object Type Trees.

- Improve interface for styling SlickTree for typesTree in bits-internal.

- Adds extension to slicktreenode for Object Type Tree nodes. Most of the style interface changes are made in slicktreenode for accessibility to other instances but can be moved to the new objecttypenode.js.

### DIFF
--- a/src/os/ui/nodetoggle.js
+++ b/src/os/ui/nodetoggle.js
@@ -14,7 +14,7 @@ os.ui.nodeToggleDirective = function() {
     restrict: 'AE',
     replace: true,
     template: '<i class="js-node-toggle c-node-toggle fa fa-fw" ' +
-        'ng-class="{\'fa-caret-right\': item.collapsed, \'fa-caret-down\': !item.collapsed}"></i>',
+        'ng-class="{\'{{collapsedIcon}}\': item.collapsed, \'{{expandedIcon}}\': !item.collapsed}"></i>',
     controller: os.ui.NodeToggleCtrl,
     controllerAs: 'toggle'
   };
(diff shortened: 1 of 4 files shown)


src/os/ui/nodetoggle.js
```diff
@@ -99,3 +125,16 @@ os.ui.NodeToggleCtrl.prototype.updateOpacity = function() {
...
 };
+
+
+/**
+ * Updates the icons displayed by the directive.
+ *
+ * @private
+ */
+os.ui.NodeToggleCtrl.prototype.updateIcons_ = function() {
+  var toggleIcons = this.scope_['item']['getToggleIcons']();
```
schmidtk, line 136: I would recommend using a JSDoc `@type` assignment so you can call `.getToggleIcons()` instead of using bracket notation. That will allow the compiler to type check the call. This is done elsewhere in this file with:

```
var item = /** @type {os.ui.slick.SlickTreeNode} */ (this.scope_['item']);
```

src/os/ui/nodetoggle.js
```diff
@@ -49,17 +49,40 @@ os.ui.NodeToggleCtrl = function($scope, $element) {
    */
   this.element_ = $element;
 
+  /**
+   * @type {string}
+   */
+  this.scope_['expandedIcon'] = '';
```
schmidtk, line 55: We should initialize these to the defaults to ensure icons are set if `item` isn't a `SlickTreeNode`.

schmidtk (changes requested):

src/os/ui/nodetoggle.js
Gamerlios123: Thanks for the review.

schmidtk (approved):

schmidtk (approved):

PR #1152, opened 2021-01-26
refactor(electron): rework certificate selection ui

schmidtk:
The existing Electron certificate selection UI lacks in a few areas:

- If the default cert isn't the one you want, it takes three clicks to change it (open dropdown, select cert, OK).
- The dropdown doesn't show enough information about the cert until opened, and questionably not then either.

To address these issues, each certificate is now displayed as a card with clearly displayed cert details.

Old UI:
Old UI

New UI:
New UI

### DIFF
--- a/src/plugin/electron/electronconfirmcert.js
+++ b/src/plugin/electron/electronconfirmcert.js
@@ -68,7 +68,6 @@ class Controller {
     if (this.scope) {
       const certs = /** @type {Array<!Electron.Certificate>} */ (this.scope['certs']);
       if (certs) {
-        certs.sort(sortCerts);
         this['cert'] = certs[0];
       }
 
@@ -77,29 +76,20 @@ class Controller {
   }
 
   /**
-   * Get the user-facing name for a certificate.
+   * Format the certificate expiry to a date string.
    * @param {Electron.Certificate} cert The certificate.
-   * @return {string} The name.
+   * @return {string} The formatted expiry.
    * @export
    */
-  getName(cert) {
-    return cert ? `${cert.subjectName} (${cert.issuerName})` : 'Unknown Certificate';
+  formatExpiry(cert) {
+    let expiry = 'Unknown';
+    if (cert && cert.validExpiry > 0) {
+      expiry = moment(cert.validExpiry * 1000).format('dddd, MMMM D, YYYY');
+    }
+
+    return expiry;
   }
 }
-
-
-/**
- * Sort client certificates by subject/issuer name.
- * @param {Electron.Certificate} a First certificate.
- * @param {Electron.Certificate} b Second certificate.
- * @return {number} The sort value.
- */
-const sortCerts = (a, b) => {
-  // Sort by subject name, then issuer name.
-  return a.subjectName === b.subjectName ?
-      (a.issuerName > b.issuerName ? 1 : a.issuerName === b.issuerName ? 0 : -1) :
-      (a.subjectName > b.subjectName ? 1 : -1);
-};
 
 
 /**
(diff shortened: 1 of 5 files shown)


src/plugin/electron/electronconfirmcert.js
```diff
@@ -77,31 +76,22 @@ class Controller {
...
-   * @return {string} The name.
+   * @return {string} The formatted expiry.
    * @export
    */
-  getName(cert) {
-    return cert ? `${cert.subjectName} (${cert.issuerName})` : 'Unknown Certificate';
+  formatExpiry(cert) {
+    let expiry = 'Unknown';
+    if (cert && cert.validExpiry > 0) {
+      expiry = moment(cert.validExpiry * 1000).format('dddd, MMMM D, YYYY');
```
bradh, line 87: Year then Month, then Day? (and skip day of week)?

src/plugin/electron/electronconfirmcert.js
```diff
@@ -68,7 +68,6 @@ class Controller {
     if (this.scope) {
       const certs = /** @type {Array<!Electron.Certificate>} */ (this.scope['certs']);
       if (certs) {
-        certs.sort(sortCerts);
         this['cert'] = certs[0];
```
bradh, line 71: It still might be nice to sort certificates. Default, then by expiry date from latest to earliest, so you get ones that might work before ones that probably won't.

Possibly this is already done somewhere else, if so, disregard.

src/plugin/electron/electronconfirmcert.js
schmidtk: I was following the format Chrome uses when you inspect certificates. Open to suggestion, but this seems equivalent to how one would verbalize a date ("Tuesday, May 10, 2022") versus the ISO format ("2022 May 10") which is maybe more natural to read.

src/plugin/electron/electronconfirmcert.js
schmidtk: I was conflicted on this because it seems Chromium is already doing some internal sorting in the list provided to Electron. It seemed better to leave the list as-is.

src/plugin/electron/electronconfirmcert.js
bradh: My Chrome 86 doesn't show dates in the chooser. I see Subject, Issuer and Serial Number. In the inspection, I see Valid from and Valid to as numerical dates (which appear localised to DD/MM/YYYY for me).

Localised date format would be good, but that seemed like more effort that the ISO thing.

schmidtk: Ahh I was looking at the certificate inspection dialog, if you look at the cert for the current page.
Cert Dialog

bradh: I see a different format again (DD-Mmm-YYYY) on a different machine. A little experimentation suggests that (at least on Windows) its whatever the "Short date" format is from the "Date & Time Settings".

schmidtk: Ahh, that makes sense. Should we use `Date#toLocaleDateString` and let the browser format it? I like the full month, but I'm not sure how to format that appropriately by locale.

wallw-bits: For this particular ticket, OpenSphere does not support internationalization. Since `Date#toLocalDateString` is probably not what you want, I'd use a format consistent with the rest of the tool and add a ticket for potentially addressing internationalization elsewhere.

schmidtk: Sounds good. I changed the format to match the timeline panel, so it will read something like `2021 Feb 1`.

jsalankey (approved):

PR #1155, opened 2021-01-27
refactor(jquery): update jquery version

kwilson-bits:


schmidtk (approved):

schmidtk:
Do we want to use a caret version for jQuery given it can have breaking changes in minor releases? We could use `~3.5.0` to support 3.5.x, and still get patch updates.

PR #1156, opened 2021-01-27
Add support for Arc ImageServer and better FeatureServer support

jsalankey:
This PR includes updates to allow for loading an Arc FeatureServer that has no corresponding MapServer as well as adding support for the Arc ImageServer service. This works very easily in 2D mode through Openlayers' included `ImageArcGISRest` source, but had some hitches in 3D, some of which still need addressing. I have been unable to solve these two issues:

1. The images sometimes get clipped by the globe.
2. There is occasional flickering when swapping out the primitive with a new one on a fresh image loading.

### DIFF
--- a/externs/Cesium.externs.js
+++ b/externs/Cesium.externs.js
@@ -1393,6 +1393,15 @@ Cesium.Cartesian3.fromDegrees = function(lon, lat, opt_alt, opt_ellipsoid, opt_r
 
 
 /**
+ * @param {Array<number>} coordinates The coordinates in degrees
+ * @param {Cesium.Ellipsoid=} opt_ellipsoid The ellipsoid.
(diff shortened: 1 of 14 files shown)


schmidtk:
I'm trying to load the [City of Aurora ArcGIS server](https://ags.auroragov.org/aurora/rest/services/OpenData/MapServer), and it doesn't load any of the layers. When the service node loads, the service type is `null` and it doesn't process any of the layers.

jsalankey:
> I'm trying to load the [City of Aurora ArcGIS server](https://ags.auroragov.org/aurora/rest/services/OpenData/MapServer), and it doesn't load any of the layers. When the service node loads, the service type is `null` and it doesn't process any of the layers.

Good catch, should be fixed. Server I was testing with didn't have a `layers` property, so I forgot to update that part of the loader.

schmidtk (approved):

PR #1157, opened 2021-01-28
fix: avoid overwriting feature label config

wallw-bits:
Only update those defaults if the config does not have a specific value set.

Tested against regular file layers and the following snippet:
```javascript
let feature = new ol.Feature(new ol.geom.Point([0, 0]));
let config = {
  'text': {
    'text': 'Test',
    'textAlign': 'left',
    'textBaseline': 'middle',
    'font': '10px/10px Arial',
    'fillColor': 'rgba(255, 255, 0, 1)',
    'offsetX': 0,
    'offsetY': 15,
    'placement': 'point'
  },
};
config[os.style.StyleField.LABEL_SIZE] = 10;
feature.set(os.style.StyleType.FEATURE, config);
os.map.mapContainer.addFeature(feature);
```

Should probably also test against layers with more styling, such as KML or Simple Style GeoJSON.


schmidtk:
I tested with regular files and remote (Arc) layers and labels are rendered the same in 2D/3D as on `master`.

With KML, we were previously picking up a label color from other color fields. In this PR, KML labels will always use the default layer color. The old behavior was not correct, but I think we can fix this by implementing `toConfig` in `TextReader` and calling that in `StyleReader#toConfig`.

wallw-bits:
The `toConfig` bit did not work (or perhaps I did not understand what you were saying). However, the new update seems to get the same behavior as `master` for that file. You mentioned that the previous behavior was not correct, so maybe we should talk about what that should be.

src/os/style/label.js
```diff
@@ -505,41 +493,134 @@ os.style.label.createAdditionalLabels = function(feature, config, opt_layerConfi
...
+ * @param {ol.style.Style} labelStyle
+ * @param {Object} config
+ */
+os.style.label.updateDefaultOffsetX = function(labelStyle, config) {
+  // labels need to be offset a little more when next to an icon. this helps, but isn't nearly complete.
+  // TODO: determine the size of the rendered feature and use that for the x offset
+  if (!config || !config['text'] || config['text']['offsetX'] === undefined) {
+    const text = labelStyle.getText();
+    const fontSize = os.style.label.DEFAULT_SIZE;
+    const offsetx = os.style.isIconConfig(config) ? fontSize : (fontSize / 2);
```
schmidtk, line 606: The `baseLabelConfig` passed into this function does not have information from the original config that `isIconConfig` depends on. This check always returns false, which is changing the default x-offset for icons.

schmidtk (changes requested):
The default x-offset for icons changed from master, more specific comment on the related line.

src/os/style/label.js
wallw-bits: Should be fixed in my recent update.

schmidtk (approved):

PR #1158, opened 2021-01-29
fix: calls to updatePreview continue to subclasses

jkae:
Experienced an issue where the `goog.base` call from `PlacemarkEditCtrl` was failing out after completing the associated code in `FeatureEditCtrl`.

Change events for Orientation in ellipse geometries -- which were handled via `ng-change` hooks in the html -- were unable to propagate through to where they needed.


schmidtk (changes requested):
The `ng-change` wasn't working properly because `PlacemarkEditCtrl#updatePreview` does not have `@export` on it. The `@export` annotation makes an unminified reference to the minified function, so having it on `FeatureEditCtrl` but not `PlacemarkEditCtrl` meant the function Angular was calling was from the wrong class.

Could you add `@export` to `PlacemarkEditCtrl#updatePreview` and see if that resolves your issue? The rest of the changes in this PR are fine, but I personally would prefer keeping `ng-change` instead of adding more `$watch` statements.

schmidtk (approved):

PR #1159, opened 2021-01-29
Iterate on modify interaction

schmidtk:
This resolves a few outstanding issues with the modify interaction caused by modifying the interpolated copy of the geometry. The interaction will now render the interpolated geometry, but modify the original. This makes it easier to modify large geometries with many intermediate interpolated points between original vertices.

This also removes interpolation from the polygon draw interaction, which was modifying the geometry directly and discarding the original drawn coordinates. The drawn polygon is expected to be interpolated at the feature level, where the original will be saved for future reference.


src/os/interaction/modifyinteraction.js
```diff
@@ -36,19 +137,51 @@ const OSMap = goog.requireType('os.Map');
...
   /**
    * Constructor.
-   * @param {olx.interaction.ModifyOptions=} opt_options Options.
+   * @param {!Feature} feature The feature to modify.
    */
-  constructor(opt_options) {
-    const options = opt_options || {};
+  constructor(feature) {
+    const interpolationMethod = /** @type {interpolate.Method} */ (feature.get(interpolate.METHOD_FIELD)) ||
+        interpolate.Method.RHUMB;
```
jsalankey, line 144: Should this default to the current application interpolation method instead of Rhumb?

src/os/interaction/modifyinteraction.js
```diff
@@ -247,62 +482,54 @@ class Modify extends OLModify {
...
-      }),
-      stroke: new Stroke({
-        color: [255, 255, 255, 1],
-        width: 2
-      })
-    }),
-    zIndex: Infinity
-  })
-];
+OLModify.handleDownEvent_ = function(evt) {
```
jsalankey, line 505: Will this kind of mixin/override work once we take `declareLegacyNamespace` out of this file? We can worry about that when we get to it, just might add a note if it's a concern.

src/os/interaction/modifyinteraction.js
schmidtk: Yeah I'm not sure why I defaulted to rhumb. It's now using `interpolate.getMethod()` instead.

src/os/interaction/modifyinteraction.js
schmidtk: Yeah it will work until we update OpenLayers, which will probably break all of our mixins. All `declareLegacyNamespace` is doing is adding the module's exports to `window` using the module's namespace, so legacy `goog.provide` files can reference it. Removing it won't change any interaction with `goog.provide` files from OL/Closure/etc.

jsalankey (approved):

PR #1165, opened 2021-02-04
fix(ol): fix errors with gcc's compilation of ol namespaces

schmidtk:
GCC assumes `ol` already exists due to including `typedefs.js` as an externs file. It also never creates the `ol.ext` namespace, presumably due to how those files are included in the compilation which results in Closure not replacing their
`goog.provide` statements properly.


jsalankey (approved):

schmidtk:
I added `@suppress {checkTypes}` to the file so GCC will ignore the actual OL types referenced in the JSDocs. That removes the need to `goog.require` anything, which either needed to require every referenced namespace in the file or rely on those to be implicitly required elsewhere.

jsalankey (approved):

PR #1168, opened 2021-02-10
feat(kml): utilizes mirror for kml icon exports

glennkitchellcaci:
When the primary Google icon source is unavailable, KMLExport will switch the icon source to either 
1) the fully-Location-resolved internal `/images/icons/kml` folder (e.g. https://my.url/images/icons/kml/...), or
2) a mirror website, as configured in `settings.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "copy:resources": "xargs -n 2 cp -r < .build/resources-copy-files",
     "copy": "npm-run-all -s copy:files copy:images copy:onboarding copy:views copy:config copy:resources",
     "rename": "find dist -type f | xargs perl -pi -e \"s#{APP}#$(json -a admin.brand.opensphere admin.about.application -d '\\n'< .build/settings.json | sed '/^$/d' | head -n 1)#g\"",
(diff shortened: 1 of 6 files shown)


bradh:
@glennkitchellcaci Is the use of mirror as a fallback (if configured) a hard requirement?

Would it be possible to use the mirror if configured, then fallback to the "standard" location(s)?

Or am I missing what this does?

glennkitchellcaci:
@bradh Thankye for the feedback! I'm swapping the `pushd` and `popd` to `cd` as we speak. About your questions:

> Would it be possible to use the mirror if configured, then fallback to the "standard" location(s)?

For consistency, it works the same basic way as replaceGoogleUri() -- where it will only enable the swap when the default Google Earth Icon source (maps.google.com) is unavailable.

> Is the use of mirror as a fallback (if configured) a hard requirement?

The logic is: During a KML export only and only if Google is not available and a Google icon is selected > swap to "mirror URL" (if configured) OR the "current URL" (the /images folder contains a built-in mirror as part of OpenSphere). The "mirror URL" configuration - if one is there - taking precedence over the built-in "current URL" is a hard requirement.

bcleveland1022 (approved):

PR #1169, opened 2021-02-10
fix(href): add href to a tags so hyperlinks appear correctly

lilysikes:
After JQuery update it was discovered that <a> tags without href don't appear correctly. This ticket fixes those issues, but for now we must change "attribute-allowed-values" errors to be warnings.

### DIFF
--- a/.htmlvalidate.json
+++ b/.htmlvalidate.json
@@ -8,6 +8,7 @@
   },
 
   "rules": {
+    "attribute-allowed-values": "warn",
     "element-name": "off",
     "element-required-attributes": "off",
     "no-inline-style": "warn",
--- a/views/filter/editfilters.html
+++ b/views/filter/editfilters.html
@@ -15,8 +15,8 @@
         </div>
       </div>
       <ul class="nav nav-tabs flex-shrink-0">
-        <li class="nav-item"><a class="nav-link" ng-class="{active: tab == 'basic'}" ng-click="filters.setTab('basic')">Basic</a></li>
-        <li class="nav-item"><a class="nav-link" ng-class="{active: tab == 'advanced'}" ng-click="filters.setTab('advanced')">Advanced</a></li>
+        <li class="nav-item"><a href class="nav-link" ng-class="{active: tab == 'basic'}" ng-click="filters.setTab('basic')">Basic</a></li>
+        <li class="nav-item"><a href class="nav-link" ng-class="{active: tab == 'advanced'}" ng-click="filters.setTab('advanced')">Advanced</a></li>
       </ul>
       <div ng-if="tab == 'basic'" class="flex-column d-flex flex-fill">
         <basicfilterbuilder root="filters.root" columns="columns" is-complex="filters.isComplex"></basicfilterbuilder>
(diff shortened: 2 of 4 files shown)


csgibson (approved):

csgibson:
@wallw-bits this is in line with what we are doing elsewhere but maybe not necessary for opensphere. Yay or Nay?

wallw-bits:
I am not a fan of using an anchor tag just for `ng-click`. Since these do not link anywhere, they should not be anchors in the first place. They were probably done that way because someone copy/pasted from the bootstrap examples, which intend for the `href` to have a value.

csgibson:
@wallw-bits I agree with that. I think it was the result of some of the bootstrap defaults like you mentioned. Anchor tags pull in things like 'cursor: pointer' automatically. There might be other considerations I am not aware of. Not sure if @jsalankey remembers?

We don't have to accept this PR though if it's not what we want to do. I think @schmidtk got these working after the jquery upgrade already.

schmidtk:
Bootstrap's examples use `<a href="#" ...>` for navs, which is what I used in [this commit](https://github.com/ngageoint/opensphere/commit/55dce78a0f53beb515089f1ed2950c331f1bb1e3) to fix the layers window tabs.

In doing some reading, it seems both `href="#"` and an empty `href` point to the top of the current document and should have the same behavior. I tested in Chrome, Firefox, and Edge, and both approaches had the intended behavior without unwanted side effects.

I'm fine either either approach but we should pick one and stick with it for consistency. Given the empty `href` is shorter and doesn't imply a document fragment, that might be better than Bootstrap's `href="#"`.

schmidtk:
Sorry, I take that back. Given this PR introduces lint errors which are worked around by changing `attribute-allowed-values` to a warning, we should use Bootstrap's recommended `href="#"` pattern instead.

wallw-bits:
I don't think `href="#"` will play nice with Angular. A good portion of those styles are not specific to the tag, and therefore we should try making those divs or something first.

schmidtk:
If `cursor: pointer` is the only thing we lose by switching to a div, maybe we should just add that to `.nav-link` in the Bootstrap overrides. I agree with Will, that would be preferable to using anchors on elements that aren't actually links.

.htmlvalidate.json
```diff
@@ -8,6 +8,7 @@
   },
 
   "rules": {
+    "attribute-allowed-values": "warn",
```
schmidtk, line 11: We should avoid downgrading existing errors to warnings, and instead find a solution that does not produce errors.

views/filter/editfilters.html
```diff
@@ -15,8 +15,8 @@
         </div>
       </div>
       <ul class="nav nav-tabs flex-shrink-0">
-        <li class="nav-item"><a class="nav-link" ng-class="{active: tab == 'basic'}" ng-click="filters.setTab('basic')">Basic</a></li>
-        <li class="nav-item"><a class="nav-link" ng-class="{active: tab == 'advanced'}" ng-click="filters.setTab('advanced')">Advanced</a></li>
+        <li class="nav-item"><a href class="nav-link" ng-class="{active: tab == 'basic'}" ng-click="filters.setTab('basic')">Basic</a></li>
```
schmidtk, line 18: It sounds like the only thing preventing us from using `div` here is the lack of `cursor: pointer`. Is that accurate? If so, I wouldn't be opposed to adding that to `.nav-link` in our Bootstrap overrides. I agree with @wallw-bits on not using anchor tags for things that aren't links.

schmidtk (changes requested):

PR #1170, opened 2021-02-11
fix(shapefile): fix shapefile import dialog in electron

bcleveland1022:
Currently is if user tried to import a shapefile within electron the shapefile import dialog would pop up asking for the other file dbf or shp.  The user would be allowed to browse for a file and select it but it would never populate the import dialog and never allow the user to import a shapefile.  This fix now allows the user to import the shapefile.


glennkitchellcaci (approved):
Sweet use of an old function to solve a new problem.

Issue #746, opened 2019-08-23
Units doesn't close on clicking Scale


justin-bits:
If already open, the units pop up should close on clicking the scale.  Found when `status-bar.spec.js` started failing.

justin-bits:
Workaround for test added to #735.  Be sure to update the test when fixing this.

bradh:
I think this can now be closed.

PR #1172, opened 2021-02-12
fix(scaleline): ensure menu closes on toggle action

bradh:
The report (#746) says Units doesn't close. It looks like it doesn't, but what is really happening
is the menu closes (on "click outside" processing), and then opens again (in response the click).

There is existing logic in the menu code to properly handle toggle actions, but it only works for
buttons (based on nodeName like 'button' or 'btn-group' in the classes). Changing that could have
unintended consequences outside of OpenSphere, so the approach is the allow subclasses to specify
that they can be toggled.

### DIFF
--- a/src/os/ui/menu/menu.js
+++ b/src/os/ui/menu/menu.js
@@ -328,5 +328,9 @@ os.ui.menu.Menu.prototype.dispatchEvent = function(evt) {
  * @return {boolean}
  */
 os.ui.menu.Menu.supportsToggle = function(el) {
-  return !!el && (el.nodeName.toLowerCase() == 'button' || goog.dom.classlist.contains(el, 'btn-group'));
+  return !!el
+    && (el.nodeName.toLowerCase() == 'button'
+      || goog.dom.classlist.contains(el, 'btn-group')
+      || goog.dom.classlist.contains(el, 'js-menu__toggle')
+    );
 };
--- a/src/os/ui/scaleline.js
+++ b/src/os/ui/scaleline.js
@@ -33,8 +33,9 @@ const directive = function() {
     restrict: 'E',
     replace: true,
     scope: true,
-    template: '<li class="pointer" id="scale-line" ng-click="ctrl.openMenu()" ng-right-click="ctrl.openMenu()">' +
-        '<div class="unit-group"></div></li>',
+    template: '<li class="pointer js-menu__toggle" id="scale-line" ' +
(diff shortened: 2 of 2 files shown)


src/os/ui/menu/unitmenu.js
```diff
@@ -79,6 +79,12 @@ os.ui.menu.unit.toggleUnit = function(event) {
   os.unit.UnitManager.getInstance().setSelectedSystem(event.type);
 };
 
+/**
+ * @inheritDoc
+ */
+os.ui.menu.Menu.prototype.supportsToggle = function(el) {
```
schmidtk, line 85: Replacing the `supportsToggle` function on the prototype will impact any `Menu` created after this code is loaded.

src/os/ui/menu/menu.js
```diff
@@ -131,7 +131,7 @@ os.ui.menu.Menu.prototype.onClick_ = function(e) {
       openerClicked = !!jqEl[0];
     }
 
-    if (openerClicked && os.ui.menu.Menu.supportsToggle(openerEl)) {
+    if (openerClicked && (os.ui.menu.Menu.supportsToggle(openerEl) || this.supportsToggle(openerEl))) {
```
schmidtk, line 134: I'm not certain we want the menu itself to determine toggle behavior, because the same menu could potentially be opened from different DOM elements.

It might be preferable to allow a DOM element to identify itself as a menu toggle element. For example, we could add a `js-menu__toggle` class to the element and look for that in `os.ui.menu.Menu.supportsToggle`. That would both avoid any breaking changes (nothing would reasonably have a class like this already) while introducing new functionality.

schmidtk (changes requested):

src/os/ui/menu/unitmenu.js
bradh: I had intended an override (not this cut-n-paste), but that isn't going to work here. Perhaps a function passed in as a constructor option could work better.

src/os/ui/menu/menu.js
bradh: I did think about the multiple use of the menu, but it seemed like the common case was single use, and the next most common case was the opener was button (which is already handled). If you have the `openerEl` then you could figure it out for the corner cases.

I hadn't consider adding a class to the opener element. Will give that a try.

schmidtk (approved):

PR #1180, opened 2021-02-18
Move WMTS support into OGC server

schmidtk:
WMTS was initially implemented as its own plugin/server. This was convenient for WMTS-only servers, but it would be helpful to support adding an OGC server with WFS/WMTS layer pairs.

This PR moves the WMTS code into the existing OGC server. To avoid user-facing changes, WMS will be used as the default tile layer type if both WMS and WMTS layers exist for the same identifier. A new `WMTSServer` class was also added to avoid breaking existing config for WMTS-only servers.

This also fixes a couple bugs in the existing WMTS implementation that prevented the time dimension from working properly. The regex to fix the case in the URL template was missing a backslash to escape the closing brace, and the case-sensitive time key was not being used to update the dimension on the source. WMTS servers with time support (like [NASA GIBS](https://gibs.earthdata.nasa.gov/wmts/epsg4326/best/1.0.0/WMTSCapabilities.xml)) should now work properly. Note the NASA GIBS layers are out of date, but should have tiles in the range `2015-06-07/2020-06-10`.

To test adding an OGC layer with WMTS, this demo server may be used:
https://demo.geo-solutions.it/geoserver/web/

The layers are a little hit or miss if they actually work, but adding the WMTS service first will detect as an OGC server and the WFS/WMS URL's can be added in the import UI.

Another WMTS server, without time support: https://mrdata.usgs.gov/mapcache/wmts?service=WMTS&request=GetCapabilities&version=1.0.0&


schmidtk:
I added tests for most of the new code. I also added `os.net.resetCrossOriginCache` primarily to aid tests, so they can reset anything added by previous tests. That's more module-friendly given we shouldn't be accessing private variables like `crossOriginCache_` directly outside the file.

jsalankey (approved):

PR #1182, opened 2021-02-22
feat(shp): add codepage sidecar file to shapefile export

### LINKED ISSUE #767: Add codepage sidecar file to shapefile output
(i.e. `.cpg` file containing the encoding that is used for text - `UTF-8` or similar).

bradh:
Files are marked as UTF-8.

Also fixes a problem where non-ASCII values would be corrupted on export. 

Adds a new test file for shapefile export.

Resolves #767


schmidtk (approved):
This looks good, thanks for the fix @bradh!

PR #1183, opened 2021-02-22
fix(toastui): css fix

kwilson-bits:
Fix a minor z-index problem introduced by the recent toastui version upgrade.

### DIFF
--- a/scss/os/_overrides_tuieditor.scss
+++ b/scss/os/_overrides_tuieditor.scss
@@ -367,6 +367,8 @@
   //////////// EDITOR POPUP WINDOWS //////////
   .tui-popup-wrapper,
   .tui-editor-popup {
+    z-index: $u-zindex-base;
+
     .tui-popup-body,
     button:not(.tui-popup-close-button):not(.te-ok-button):not(.te-close-button) {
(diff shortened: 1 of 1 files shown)


kwilson-bits:
The default z-index from the toastui CSS was set to 9999, whereas our popups expect the z-index to be 500.

scss/os/_overrides_tuieditor.scss
```diff
@@ -367,6 +367,8 @@
   //////////// EDITOR POPUP WINDOWS //////////
   .tui-popup-wrapper,
   .tui-editor-popup {
+    z-index: 500;
```
schmidtk, line 370: We prefer to base `z-index` values off of `$u-zindex-base`, to help enforce consistent ordering across our SCSS. That's set to 500, so you should be good dropping it in here.

schmidtk (changes requested):

scss/os/_overrides_tuieditor.scss
kwilson-bits: I changed it to use that scss variable.

schmidtk (approved):

PR #1184, opened 2021-02-22
Added Custom Select style

atsuydam:
Added a custom select style to styletype.js and and an or statement to vectorsource.js to use the custom select if defined or the default select if not.


schmidtk:
The build is failing due to the commit messages not matching our [guidelines](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit). We require a subject, and preferably a scope prior to the message. For example:
```
feat(vector): added custom select color
```
This is supposed to be checked when you commit, but perhaps the git hooks were not installed in your workspace.

atsuydam:
Duplicate

PR #1187, opened 2021-02-22
fix(drawcontrols): missing config assignments in base draw controls

jkae:
`hidesExtraControls` is accessed in `initControlMenu` called from the constructor. For any es6 subclasses, where `this` may not be called before `super()`, extra controls will always be visible even if that is not desired.

`supportsLines` was also being assigned to itself which is probably not the correct behavior.

### DIFF
--- a/src/os/ui/draw/basedrawcontrols.js
+++ b/src/os/ui/draw/basedrawcontrols.js
@@ -1,7 +1,6 @@
 goog.provide('os.ui.draw.BaseDrawControlsCtrl');
 goog.provide('os.ui.draw.baseDrawControlsDirective');
 
-goog.require('goog.Disposable');
 goog.require('goog.log');
 goog.require('goog.log.Logger');
 goog.require('ol.Feature');
@@ -54,7 +53,6 @@ os.ui.Module.directive('drawControls', [os.ui.draw.baseDrawControlsDirective]);
  *
  * @param {!angular.Scope} $scope
  * @param {!angular.JQLite} $element
- * @extends {goog.Disposable}
  * @constructor
  * @ngInject
  */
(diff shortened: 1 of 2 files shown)


src/os/ui/draw/basedrawcontrols.js
```diff
@@ -114,12 +110,34 @@ os.ui.draw.BaseDrawControlsCtrl = function($scope, $element) {
...
+ * @type {goog.log.Logger}
+ * @private
+ */
+os.ui.draw.BaseDrawControlsCtrl.LOGGER_ = goog.log.getLogger('os.ui.draw.BaseDrawControlsCtrl');
+
+
+/**
+ * Angular initialization lifecycle function. Sets up event listening and our controls menu.
+ */
+os.ui.draw.BaseDrawControlsCtrl.prototype.$onInit = function() {
```
jsalankey, line 140: To be clear about the `$onInit` function, not every directive needs to have one. Typically it's useful for stuff that you want to have happen after the directive gets added to the DOM and rendered for the first time. It's useful for timing/sizing issues, and probably not necessary for other situations. It probably won't hurt anything to have it here in this case, but unless something was broken without it, it shouldn't be necessary to add it.

jsalankey (approved):

src/os/ui/draw/basedrawcontrols.js
schmidtk: I should have left comments in this ticket instead of Slack, but this particular `$onInit` is for the case where we want code to execute after the class has been fully constructed. This allows the child constructor to influence the initialization function without overriding it. In this case, the child is changing the `supportsLines` property which impacts `initControlMenu`.

PR #1189, opened 2021-02-23
Revert WMS layer parsing behavior

schmidtk:
In #1180 I tried to avoid creating a `DescriptorNode` if it was going to later be swapped for a `SlickTreeNode` due to the "layer" actually being a folder. This broke something in an app based on OpenSphere, so I'm reverting that change.


schmidtk:
FYI @glennkitchellcaci, this was the root cause of what you fixed in #1186.

glennkitchellcaci (approved):

PR #1191, opened 2021-02-23
chore: replace http-server with light-server

bradh:
This helps with wasm loading, which currently breaks chardetng usage in dev mode.


jsalankey:
Thanks for updating this!

jsalankey (approved):

PR #1192, opened 2021-02-24
fix(im): handle CSV import from Places tab

### LINKED ISSUE #740: Better handle CSV import via Places Import
CSV imports via Layers>Places are currently handled non-intuitively.  KML is the expected format and is handled correctly.  GeoJSON and SHP are not expected, but imported easily as features.  CSV import throws a warning indicating that the file type could not be detected.  Only after selecting CSV from the drop down can the file be imported as features.

- Adjust file format detection to work with CSV (as it already does with GeoJSON and SHP)
- [optional] Notify the user that the file contents will be imported as features, where the option then exists to save those features as places, OR ...
- [optional] Give the user the ability to drop the contents in directly as places

bradh:
Resolves #740 - ensures that CSV, SHP and GeoJSON are all handled by the import dialog off Places.

However I'm not sure this is the right behaviour either. Importing things on Places works, but they may not show up on Places (and don't look like placemarks or KML tracks).

If this is the desired behaviour, then it might be possible to remove the multiple import managers, and just use the global instance.

### DIFF
--- a/src/plugin/file/csv/csvplugin.js
+++ b/src/plugin/file/csv/csvplugin.js
@@ -67,6 +67,9 @@ plugin.file.csv.CSVPlugin.prototype.init = function() {
   im.registerImportDetails('CSV', true);
   im.registerImportUI(os.file.mime.csv.TYPE, new plugin.file.csv.ui.CSVImportUI());
   im.registerParser(this.id, plugin.file.csv.CSVParser);
+  os.placesImportManager.registerImportDetails('CSV', true);
+  os.placesImportManager.registerImportUI(os.file.mime.csv.TYPE, new plugin.file.csv.ui.CSVImportUI());
+  os.placesImportManager.registerParser(this.id, plugin.file.csv.CSVParser);
 
   // register the csv exporter
   os.ui.exportManager.registerExportMethod(new plugin.file.csv.CSVExporter());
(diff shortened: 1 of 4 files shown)


jsalankey:
It is correct that we don't want to support just any kind of imports with the button on the Places window. That uses a separate import stack because we want to import any data that you pick with it directly into Places (rather than importing it as a layer and then separately saving that layer to Places). With KML in particular, we try to merge the tree into the Places tree, which is why it only supports KML currently.

The desired behavior would be to pipe the resultant features from parsing the other file types directly into a folder in the Places layer. We'd likely still want things like CSV to run through their current import UIs, but to have a special handling of the final result. This might involve extending some aspect of the import UIs, or possibly refactoring them to pass a success handler to them.

bradh:
I'm good with the first paragraph. I'm not sure I follow the second paragraph. 

Are the required formats for that special handling: CSV, GeoJSON and Shapefile? Should there be a way to filter before adding as Places?

jsalankey:
Currently, the import UIs like `plugin.file.csv.ui.CSVImportUI` are registered to the global file Import Manager for the purpose of parsing out files into separate layers. When you register them with the Places Import Manager, they still create layers on success, which might be unintuitive to the user.

The import UIs registered currently by `PlacesPlugin` are different from the standard KML/KMZ import UIs. They are registered here: https://github.com/ngageoint/opensphere/blob/master/src/plugin/places/placesplugin.js#L77-L79.

In kmlplacesimportui.js `plugin.places.KMLPlacesCtrl.prototype.onImportComplete_`, the success handler doesn't create a new layer, it adds the results to the Places tree. For CSV/GeoJSON/SHP, I was envisioning either whole new import UI classes or possibly a refactor of the existing ones to allow changing how they handle successfully parsing files that add the results to Places instead of a new layer.

I'm not sure I'm following the filter question, but I don't think we'd be looking to add anything filtering related here.

PR #1195, opened 2021-02-25
WIP: feat(specialcodeic): adds angle bracket sanitation to mist layer labels

Gamerlios123:


wallw-bits:
Is this coming from the Layer Rename popup or some other source?

Gamerlios123:
It's for a UOM ticket intended to allow angle brackets to be displayed in many fields.

src/os/data/layernode.js
```diff
@@ -341,7 +342,7 @@ os.data.LayerNode.prototype.formatLabel = function(value) {
  * @inheritDoc
  */
 os.data.LayerNode.prototype.formatValue = function(value) {
-  var s = os.data.LayerNode.base(this, 'formatValue', value);
+  var s = os.data.LayerNode.base(this, 'formatValue', core.text.Sanitize.sanitizeField(value));
```
wallw-bits, line 345: This function does not exist within OpenSphere, so you cannot call it here.

wallw-bits (changes requested):

PR #1197, opened 2021-02-26
improvement(registry): improve EventTarget events

glennkitchellcaci:
Include the full entry instead of just the key when firing add, update, remove, and clear events

BREAKING CHANGE: Callbacks for registry event handlers now must accept the full entry, not just the
key; and in the case of remove/clear, the entry() and get() functions will not return a value
anymore

### DIFF
--- a/src/os/data/registry.js
+++ b/src/os/data/registry.js
@@ -39,6 +39,8 @@ class Registry extends EventTarget {
    * @inheritDoc
    */
   dispose() {
+    this.clear();
+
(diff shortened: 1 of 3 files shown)


schmidtk (dismissed):

jsalankey:
@glennkitchellcaci this has been hanging out approved for a while, let us know if it's ready to merge.

schmidtk (approved):

schmidtk (approved):

PR #1199, opened 2021-03-02
Feat server inline guidance

ameletyan:


src/os/ui/im/importmanager.js
```diff
@@ -32,6 +32,12 @@ os.ui.im.ImportManager = function() {
    */
   this.importUIs_ = {};
 
+  /**
+   * @type {Object<string, Object>}
+   * @private
+   */
+  this.serverTypes_ = {};
```
schmidtk, line 39: When you have a well-defined object type, it's preferred to create a `@typedef` in externs (`os.externs.js`). This allows the compiler to fully type check the object, and will prevent the compiler from renaming properties on the object which can cause the compiled app to behave differently.

I would recommend creating a typedef called `osx.import.ServerType` and using it here in place of `Object`. Also update the type in any functions that reference this map.

src/os/ui/uiswitch.js
```diff
@@ -204,6 +205,7 @@ os.ui.UISwitchCtrl.prototype.update_ = function() {
 
       // put the items on the scope
       s['items'] = items;
+      s['config'] = config;
```
schmidtk, line 208: Any information that needs to be passed through the scope should be on `items`.

src/os/ui/window/arcserverhelp.js
```diff
@@ -0,0 +1,79 @@
+goog.module('os.ui.window.ArcServerHelpUI');
```
schmidtk, line 1: Plugin-specific code should be contained within the plugin. Please move this under `src/plugin/arc` and update the module name appropriately.

src/os/ui/window/arcserverhelp.js
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * A dialog with information about custom date/time formats, as implemented by Moment.js.
+ *
+ * @return {angular.Directive}
+ */
+const directive = () => {
+  return {
+    restrict: 'E',
+    replace: true,
+    templateUrl: os.ROOT + 'views/window/arcserverhelp.html',
```
schmidtk, line 17: Similarly, the view should be under `views/plugin/arc`.

src/os/ui/window/ogcserverhelp.js
```diff
@@ -0,0 +1,79 @@
+goog.module('os.ui.window.OgcServerHelpUI');
```
schmidtk, line 1: This module should be in `src/plugin/ogc/ui`, and the view in `views/plugin/ogc/ui`.

src/os/ui/window/geoserverhelp.js
```diff
@@ -0,0 +1,79 @@
+goog.module('os.ui.window.GeoServerHelpUI');
```
schmidtk, line 1: This module should be in `src/plugin/ogc/ui`, and the view in `views/plugin/ogc/ui`.

schmidtk (changes requested):
The compiled-mode issues are due to the compiler renaming properties in your server type objects. This can be fixed by adding a `@typedef` for the server type object as an extern.

The server-specific help UI's also belong under their respective plugins.

src/os/ui/im/importmanager.js
```diff
@@ -30,31 +30,37 @@ os.ui.im.ImportManager = function() {
    * @type {Object<string, os.ui.im.IImportUI>}
    * @private
    */
-  this.importUIs_ = {};
+  this['importUIs_'] = {};
```
schmidtk, line 33: Bracket notation should only be used for properties that we do not want the compiler to rename, which should never apply to private properties that are only directly referenced within the class. Please switch these back to dot notation, and if something needs to be accessed externally you can add an API call to do so.

src/os/ui/uiswitch.js
```diff
@@ -69,68 +70,68 @@ os.ui.UISwitchCtrl = function($scope, $element, $compile) {
    * @type {?angular.Scope}
    * @protected
    */
-  this.scope = $scope;
+  this['scope'] = $scope;
```
schmidtk, line 73: These properties should also be reverted back to dot notation.

schmidtk (changes requested):
There are a number of places where dot notation was switched with bracket notation, and the new help UI's also use bracket notation in places where it is not needed. Dot notation is preferred because the compiler does not type check properties with bracket notation, and also doesn't minify them resulting in less compact/obfuscated code.

Please switch these back to using dot notation (`this.element_` instead of `this['element_']`).

ameletyan:
There is currently a bug in the compiled build where the help window isn't opening anymore. The current code is not ready for merge until this is fixed.

src/os/ui/file/addserver.js
```diff
@@ -0,0 +1,262 @@
...
+    /**
+     * Available server type choices in the UI.
+     * @type {Array}
+     */
+    this.items = Object.values(ImportManager.getInstance().getServerTypes() || []);
+
+    $scope.$emit(WindowEventType.READY);
+    $scope.$on(ProviderImportLoadEventType.start, this.onFormLoadingStatusChange_.bind(this));
+    $scope.$on(ProviderImportLoadEventType.stop, this.onFormLoadingStatusChange_.bind(this));
+    $scope.$on('launchHelp', this.onLaunchHelp_.bind(this));
```
schmidtk, line 93: Currently the help UI only shows up when adding a server via the new Add Server dialog. It would be nice if it was also displayed when adding a server via the Import Data dialog (the file open button next to Add Data). To accomplish this, I think we can do the following:

- Move `launchHelp` and `closeHelpWindow` into `ProviderImportCtrl`.
- Put `helpWindowId` on the controller. This will allow overriding classes to change the id if they want.
- We can either remove the server-specific portion of the window label (so all are consistent), or also add a `helpUiLabel` property to the controller that can be defined in overriding classes.
- Add an `$onDestroy` method to `ProviderImportCtrl` and call `closeHelpWindow` there. That will handle closing the UI when the window is closed, and when the server type is changed in Add Server (the controller will be disposed and a new one created for the different server type).
- Add a `helpUi` property to `ProviderImportCtrl` that defaults to an empty string, and can be set by overriding classes like `GeoserverImportCtrl`.
- Change `singleurlform` to only show the button when `ctrl.helpUi` is defined. This combined with the `singleurlform` reuse comment will make sure the help UI is opt-in and will display in both Add Server and Import Data for each server type that supports it.

These changes also have the benefit of moving the help window management into the same UI as the button launching help. While `$emit` and `$broadcast` are useful for passing information between directives, they are sometimes overused in ways that make the overall code more complicated/confusing.

schmidtk (changes requested):

views/file/addserver.html
```diff
@@ -0,0 +1,36 @@
+<div class="flex-fill w-100 d-flex flex-column">
+  <div class="modal-body container flex-fill">
+    <form>
+      <div class="form-group m-0" ng-if="!ctrl.loading">
+        <label for="serverType">Select a server type:</label>
+        <select title="ServerType" class="custom-select w-50"
```
schmidtk, line 6: The `ng-change` and `ng-disabled` properties can be removed, given the error/loading state are no longer managed at this level.

views/file/addserver.html
```diff
@@ -0,0 +1,36 @@
...
+            ng-model="ctrl.serverType"
+            ng-change="ctrl.onServerTypeChange_()"
+            ng-options="value.label for value in ctrl.items"
+            ng-disabled="ctrl.loading">
+        </select>
+      </div>
+    </form>
+    <div>
+      <uiswitch items="ctrl.serverType" directive-function="ctrl.getUi" scope-update-function="ctrl.updateUiScope"></uiswitch>
+      <div class="alert alert-danger" ng-if="error">
```
schmidtk, line 16: This error block and the loading block below are no longer used and can be removed.

src/os/ui/providerimport.js
```diff
@@ -38,9 +38,30 @@ os.ui.ProviderImportCtrl = function($scope, $element) {
...
+
+  /**
+   * The help UI for this provider.
+   * @type {string}
+   */
+  this['helpUi'] = null;
+
   this.initialize();
 
   $scope.$emit(os.ui.WindowEventType.READY);
```
schmidtk, line 55: Our change to reuse the form directive caused some unfortunate window positioning behavior when loading a URL from the Import Data dialog. Because the new form directive is loaded asynchronously with `templateUrl`, the window content isn't fully populated when the `READY` event is fired which means the window height isn't correct. The end result is the window is no longer centered. You can reproduce by entering this URL in the Import Data dialog:

https://demo.geo-solutions.it/geoserver/ows

To fix this, please fire the `READY` event inside a `setTimeout` callback. That will allow the rest of the template to load before the event is fired.

schmidtk (changes requested):
A little bit of cleanup left and a minor window positioning issue, otherwise this is looking good!

src/os/ui/file/addserver.js
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * @type {boolean}
+     */
+    this.loading = false;
+
+    /**
+     * @type {string}
+     */
+    this.serverType = '';
```
schmidtk, line 73: Sorry one more thing. The `serverType` and `items` properties are both referenced by the template, and as such should use bracket notation to avoid being renamed. It looks like the compiler isn't renaming them because of some unrelated externs, but that shouldn't be relied on because the behavior could change.

src/os/ui/file/addserver.js
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * @type {?angular.JQLite}
+     * @private
+     */
+    this.element_ = $element;
+
+    /**
+     * @type {boolean}
+     */
+    this.loading = false;
```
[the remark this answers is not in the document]
ameletyan: I would like to keep this property to prevent users from changing the server type after they click Save. I added a line to `accept()` and a new `onError_()` function that will set this variable. Let me know what you think of the added code.

views/file/addserver.html
ameletyan: The `ng-disabled` property was originally added to prevent users from changing the server type after clicking Save. To keep that feature, I made some changes to how the loading state is set.

schmidtk (approved):

PR #1208, opened 2021-03-15
fix(window): add button role to indicate the element is interactive

schmidtk:
`html-validate` added a new rule in their latest release, [aria-label-misuse](https://html-validate.org/rules/aria-label-misuse.html), that does not allow the `aria-label` attribute in non-interactive elements. To resolve this, the elements were changed to `button` instead of `span`.

Using `<span role="button">` would also indicate the elements are interactive and resolve the error, but would create two new warnings because `button` is preferred by the `prefer-native-element` rule.

### DIFF
--- a/views/window/dockedwindow.html
+++ b/views/window/dockedwindow.html
@@ -1,47 +1,47 @@
 <div class="docked-window d-flex">
-    <div class="modal-content d-flex flex-column flex-fill" ng-class="{'bg-transparent': overlay, 'border-0': !border}">
-      <div class="c-docked-window__buttons">
-        <span class="text-truncate flex-fill" title="{{label}}">
-          <i ng-if="icon" ng-class="icon" class="mr-1"></i>
-          <span ng-if="iconHtml" ng-bind-html="iconHtml"></span>
-          {{label}}
-        </span>
+  <div class="modal-content d-flex flex-column flex-fill" ng-class="{'bg-transparent': overlay, 'border-0': !border}">
+    <div class="c-docked-window__buttons">
+      <span class="text-truncate flex-fill" title="{{label}}">
+        <i ng-if="icon" ng-class="icon" class="mr-1"></i>
+        <span ng-if="iconHtml" ng-bind-html="iconHtml"></span>
+        {{label}}
+      </span>
 
-        <span ng-if="helpContext" class="d-inline-flex">
-          <context-onboarding context="{{helpContext}}"></context-onboarding>
-        </span>
+      <span ng-if="helpContext" class="d-inline-flex">
+        <context-onboarding context="{{helpContext}}"></context-onboarding>
+      </span>
 
-        <span ng-repeat="headerBtnCfg in headerBtns | reverse"
-            ng-click="windowCtrl.onHeaderBtnClick($event, headerBtnCfg);"
-            title="{{headerBtnCfg.tooltip}}">
-          <i class="{{'c-glyph ' + headerBtnCfg.iconClass}}"></i>
-        </span>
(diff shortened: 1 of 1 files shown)


schmidtk:
I also fixed the tab level in the template (it had an extra level of indentation), which makes the diff harder to read. The change is really just `span -> button` for the close/hide buttons.

views/window/dockedwindow.html
```diff
@@ -1,47 +1,47 @@
...
+      <span ng-if="helpContext" class="d-inline-flex">
+        <context-onboarding context="{{helpContext}}"></context-onboarding>
+      </span>
 
-        <span ng-repeat="headerBtnCfg in headerBtns | reverse"
-            ng-click="windowCtrl.onHeaderBtnClick($event, headerBtnCfg);"
-            title="{{headerBtnCfg.tooltip}}">
-          <i class="{{'c-glyph ' + headerBtnCfg.iconClass}}"></i>
-        </span>
+      <span ng-repeat="headerBtnCfg in headerBtns | reverse"
```
glennkitchellcaci, line 14: Did these ones not complain too? Was the error because I had an accidental (and invalid) type="button" on that one span?

views/window/dockedwindow.html
```diff
@@ -1,47 +1,47 @@
...
+        <i class="{{'c-glyph ' + headerBtnCfg.iconClass}}"></i>
+      </span>
 
-        <span ng-if="showCollapse" 
-            ng-click="windowCtrl.toggle()" 
-            title="{{collapsed ? 'Expand' : 'Collapse'}}" 
-            class="d-inline-flex">
-          <i class="fa c-glyph window-close" ng-class="{'fa-compress': !collapsed, 'fa-expand': collapsed}"></i>
-        </span>
+      <span ng-if="showCollapse"
```
glennkitchellcaci, line 20: Ditto.

views/window/dockedwindow.html
```diff
@@ -1,47 +1,47 @@
...
-            aria-label="Hide">
-          <span class="c-glyph" aria-hidden="true">&times;</span>
-        </span>
-      </div>
-      <div class="c-docked-window__content w-100 flex-fill d-flex flex-column">
-        <div class="js-window__wrapper d-flex flex-column flex-fill" ng-transclude></div>
-      </div>
+      <button ng-if="showHide"
+          class="btn border-0 p-0 close flex-shrink-0"
+          type="button"
```
glennkitchellcaci, line 37: This might have been the source. I think type=button isn't valid for spans. But swapping to actual buttons is better overall, so I'm good with that!

glennkitchellcaci (commented):
Thank you for fixing this! I think just removing type=button from that old span would fix the errors if you want to just patch and be done with it. I can do an MR ASAP to swap to buttons if they don't just "work" with the tight styling out of the box.

views/window/dockedwindow.html
schmidtk: No, this one doesn't have `aria-label` on it, though it probably should have one that matches `title`. That would produce the same error unless it was also switched to a `button`.

views/window/dockedwindow.html
schmidtk: Even without `type="button"`, a `span` does not meet the criteria for the rule. It would need a `role` that indicates it's interactive like `role="button"`. We could disable the rule if we disagree with its behavior, it [seems the intent](https://gitlab.com/html-validate/html-validate/-/issues/110) was to avoid overuse of the attribute but this is a legitimate use for it.

jsalankey (approved):

PR #1209, opened 2021-03-16
fix(kmlnodelayer): features with multiple geoms show fill appropriately

arwatts16:
Features with multiple geometries (such as an ellipse that consists of a center point and a ring) will now show fill as an option in the layers menu, and will no longer auto-fill when the color is changed.


arwatts16:
Updating this to be from release rather than master

PR #1210, opened 2021-03-16
feat(formulas): disable input when no cell is selected

arwatts16:
Features with multiple geometries (such as an ellipse that consists of a center point and a ring) will now show fill as an option in the layers menu, and will no longer auto-fill when the color is changed.

### DIFF
--- a/src/plugin/file/kml/kmlnodelayerui.js
+++ b/src/plugin/file/kml/kmlnodelayerui.js
@@ -1092,9 +1092,13 @@ plugin.file.kml.KMLNodeLayerUICtrl.prototype.isFeatureFillable = function() {
   var features = this.getFeatures();
   var feature = features.length > 0 ? features[0] : null;
   if (feature) {
-    var geometry = feature.getGeometry();
+    let hasPolyGeom = false;
 
-    return os.geo.isGeometryPolygonal(geometry);
+    os.feature.forEachGeometry(feature, (geom) => {
+      hasPolyGeom = os.geo.isGeometryPolygonal(geom, true) || hasPolyGeom;
+    });
+
+    return hasPolyGeom;
   }
 
   return false;


jsalankey:
The build failure here looks unrelated, might need a rebase.

src/plugin/file/kml/kmlnodelayerui.js
```diff
@@ -1092,9 +1092,13 @@ plugin.file.kml.KMLNodeLayerUICtrl.prototype.isFeatureFillable = function() {
...
   if (feature) {
-    var geometry = feature.getGeometry();
+    const hasPolyGeom = [];
 
-    return os.geo.isGeometryPolygonal(geometry);
+    os.feature.forEachGeometry(feature, (geom) => {
+      hasPolyGeom.push(os.geo.isGeometryPolygonal(geom, true));
+    });
+
+    return hasPolyGeom.some((bool) => bool);
```
jsalankey, line 1101: It's probably cleaner to do this with `hasPolyGeom` as a boolean. In a normal `for` loop you could simply `break` once it becomes true, but with `forEachGeometry`, you'll have to do the whole loop no matter what, so you'll only ever want to set the flag once.

arwatts16:
> The build failure here looks unrelated, might need a rebase.

Release just got rebuilt and it's failing there too. Might have to wait to merge until that fix is made.

src/plugin/file/kml/kmlnodelayerui.js
arwatts16: Didn't even think of that, but makes sense

jsalankey (approved):

bcleveland1022 (approved):

schmidtk:
Build error is unrelated and fixed on master, merging.

PR #1212, opened 2021-03-22
Improve error messages from known services

schmidtk:
Services like OGC and Arc provide information in the error response which may be useful to someone trying to add a server. This PR adds a new API to `os.net` and `os.net.Request` to register/use default response validators on an opt-in basis. These validators will parse out known error messages so they can be passed on to a user.

This behavior has been enabled in `UrlMethod`, which is used to detect/add servers to OpenSphere. Errors should be detected for OGC and Arc servers and displayed in the alert. Below are some examples that can be used for testing.

- https://demo.geo-solutions.it/geoserver/gwc/service/wmts: Should fail and indicate the Request parameter is missing.
- https://demo.geo-solutions.it/geoserver/ows: Should succeed, detected as a Geoserver. Geoserver returns 200 OK for OWS errors, and we resolve these and detect/load the Geoserver appropriately. This is counter to the OGC specification, which lists error codes for all error situations.
- https://gis.arapahoegov.com/arcgis/rest/services/Nope: Should fail and indicate the folder was not found.
- https://gis.arapahoegov.com/arcgis/rest/servicess: Should fail and indicate the URL is invalid.
- https://gis.arapahoegov.com/arcgis/rest/services: Should succeed, detected as an Arc server.
- Invalid URL (enter random text): Should fail and indicate the resource was not found.

As part of this update, an implicit circular dependency in `plugin.arc` was found/resolved by adding the new `IArcLoader` interface.

This PR also fixes an issue in modern browsers (Chrome 60+ and others) which are following the HTTP/2 specification on XHR response headers. These browsers will lowercase response header names, which broke code expecting a specific case (ie, `Content-Type` instead of `content-type`). To resolve this, `SameDomainHandler#getResponseHeaders` will now lowercase all header names to provide consistent behavior across all browsers. This is a breaking change, and dependent apps/libraries should be updated accordingly. All impacted code in OpenSphere has been updated here.


jsalankey:
This generally looks good, but the new API for default request validators has the potential to cause us pain if we open it up to use in more places. If we ever add a loose validator function, or just get unlucky with a response in the URL method hitting one of the validators when it shouldn't, we could get some nasty side effects. For now, it's probably OK, but it's definitely something to keep an eye on for the future.

schmidtk:
I fixed the failing test. My fix for the Arc circular dependency relied on a call made in the Arc plugin, which wasn't called in tests. I updated that fix so it's no longer dependent on making an extra function call and shouldn't be breaking.

As for potential side effects, GeoServer was an immediate case I had to fix with the status code check. I can't think of other cases where we expect/ignore an error and continue to load an OGC or Arc server. I also tried to make the checks very specific to reduce the chances of a false positive matching the error content.

jsalankey (dismissed):

schmidtk:
I integrated this with #1199 to ensure OGC/Arc errors were detected and included in the Add Server error alert. I fixed a couple bugs I noticed in the process, and did some code reuse cleanup and clarified some of the new server help dialogs.

PR #1215, opened 2021-03-23
refactor(columnmanager): separate static functions into new file

csgibson:
This slightly improves the ease of transition to goog modules and allows subprojects to reference the utils before the rest of opensphere is migrated.


src/os/ui/column/columnmanagerutils.js
```diff
@@ -0,0 +1,82 @@
+goog.module('os.ui.column.ColumnManagerUtils');
```
schmidtk, line 1: Do we want to make this `os.ui.column` to avoid the breaking API change? Or stick with the more descriptive name?

csgibson: Actually we can't do this, I don't think: depending on the build order, it appears to make it shadow the directive and ctrl deeper on the namespace.

schmidtk: I thought that should be resolved by this block:
https://github.com/ngageoint/closure-webpack-plugin/blob/master/src/dependencies/goog-loader-suffix-dependency.js#L36-L41

But perhaps it's not working with the mix of `goog.provide` and `goog.module`.

csgibson: I am not sure. It's causing a failure in test in one of the minor subprojects. 

This fixes it: https://github.com/ngageoint/opensphere/pull/1217

PR #1219, opened 2021-03-26
Iterate on nav bar collapse behavior

schmidtk:
The top nav had a few issues with its collapse behavior, namely:

1. The collapsed state did not initialize properly, so items would overlap when first loading OpenSphere.
2. The left/right/center were allocated a fixed size, which caused button collapse to happen earlier than needed.
3. The time UI was shifted down to a new row too late, so it would sometimes overlap other elements.
4. Changing the size of the time UI (for example, setting duration to `custom`) would not cause buttons to collapse and could result in overlap.

Changes to address the above:

- The `punyparent` directive now detects when children are added/removed from the element and recomputes the max size needed for the expanded state, then updates the expanded state. (Fixes 1)
- The `osnavtop` template was updated so the center (time UI) takes only as much space as it needs (using `col-lg-auto` instead of `col-lg-4`), and the left/right use remaining space (`col-lg`). (Fixes 2)
- Grid breakpoints were updated to better handle shifting the time UI down, and `!default` was added so plugins/apps can override the breakpoints in their own styles. (Fixes 3)
- The `punyparent` directive now watches for size changes and will update the collapse state when this happens. (Fixes 4)

Additional changes:

- Some nav items are added as an `li` element. The nav lists wrap each item in an `li`, and nested `li` is not allowed so the wrapper was actually being created as a sibling instead of a parent. I changed these items to use `div` instead, so they can be a child of the `li` wrapper.
- The `list` directive was using `append` to move items to the correct position in the update function. We don't currently have an API to change the priority or otherwise move an item (short of removing/adding it), so the `append` call was performing an unnecessary remove/add in the DOM. This was noticed testing the new `MutationObserver` code, as extra events were being fired to remove the elements and re-add them. To resolve this, existing elements are skipped and new elements are inserted after the previous one.
- The search results panel was taking up way too much space in smaller windows. It is now limited to 50% of the screen width once the `lg` breakpoint is hit. Larger windows will use the default max width (no change in behavior).
- The time UI extension (hh/mm button) was not being positioned/sized properly. It should now always be centered below the time UI, and will no longer take up the full screen width when the time UI is shifted to its own row.
- Fixed HTML lint warnings in the new OGC help template.

This branch is deployed to surge.sh here: https://quixotic-trains.surge.sh/

### DIFF
--- a/src/os/ui/util/punyparent.js
+++ b/src/os/ui/util/punyparent.js
@@ -2,6 +2,7 @@ goog.provide('os.ui.util.PunyParentCtrl');
 goog.provide('os.ui.util.punyParentDirective');
 
 goog.require('goog.Throttle');
+goog.require('goog.async.Delay');
 goog.require('goog.dom.ViewportSizeMonitor');
 goog.require('os.ui');
 goog.require('os.ui.Module');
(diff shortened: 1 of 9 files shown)


src/os/ui/util/punyparent.js
```diff
@@ -29,95 +30,184 @@ os.ui.Module.directive('punyparent', [os.ui.util.punyParentDirective]);
...
-      var c = $(child);
-      // ignore the resize trigger since thats the parent size
-      if (!c.hasClass('resize-triggers')) {
-        accumulator += c.outerWidth(true);
-      }
-      return accumulator;
-    }, 0);
+os.ui.util.PunyParentCtrl.prototype.watchChild_ = function(child) {
+  const childEl = $(child);
+  if (!childEl.hasClass('resize-triggers') && !childEl.hasClass('resize-sensor')) {
```
jsalankey, line 163: Are these class exclusions to ensure that we don't double-watch a child somehow? Is there a concern where something might have its own internal resize listener that adds these, causing us miss watching it here?

schmidtk: The resize libraries add elements to the DOM that they use to determine when a resize happens. These elements have absolute positioning and 100% width so their width is the size of the entire nav bar, but they aren't visible and don't contribute to determining the relative size of things in the nav. We want to ignore them, and I had to add the `resize-sensor` class because it's used by the new resize library that backs `os.ui.resize`. I think `resize-triggers` was the class for the old resize library.

jsalankey: Ahh gotcha, I thought the classes were added to the existing elements being checked for resize rather than entirely new dummy elements.

jsalankey (approved):

PR #1226, opened 2021-04-06
Convert Cesium plugin to use goog.module

schmidtk:
The entire Cesium plugin has been converted to use `goog.module`. I cleaned up somethings we were doing in the plugin that resulted in compiler errors using ES classes.

- Some Cesium classes like `Cesium.TilingScheme` are really abstract classes and don't provide any functional code. The Cesium class/object is not available until the library is loaded, so we had code to defer creating extending classes until that happens. This was a little strange with modules/classes, so I changed the Cesium externs to make the original constructor an interface instead. This is much cleaner, and has no side effects given Cesium's implementations were all empty anyway.
- Some Cesium mixins also needed to load async after the library loaded, so I cleaned up those files to a more consistent and module-friendly pattern.
- We frequently used `Object.defineProperties` to add `get/set` functions to Cesium plugin classes. ES classes can use `get/set` directly, so I refactored all of those so the resulting code is much cleaner.
- We had some janky code to replace the OLCS camera class with our own, because we didn't want to use their view listener. We were using all of the other code, so I dropped the class replace and instead replaced `setView_` (the view change handler) on the OLCS class with an empty function. This has the same end result, but in a less janky manner.

### DIFF
--- a/src/plugin/cesium/cesium.js
+++ b/src/plugin/cesium/cesium.js
@@ -1,7 +1,6 @@
 goog.module('plugin.cesium');
 
 const MapContainer = goog.require('os.MapContainer');
-const ui = goog.require('os.ui');
 const settings = goog.require('os.config.Settings');
 const Promise = goog.require('goog.Promise');
 const Uri = goog.require('goog.Uri');
@@ -16,6 +15,7 @@ const net = goog.require('os.net');
 const proj = goog.require('os.proj');
 const utils = goog.require('os.query.utils');
 const osString = goog.require('os.string');
+const osWindow = goog.require('os.ui.window');
 const ConfirmUI = goog.require('os.ui.window.ConfirmUI');
 const ImageryProvider = goog.require('plugin.cesium.ImageryProvider');
 
@@ -187,7 +187,7 @@ const loadCesium = function() {
  */
 const promptForAccessToken = function() {
   return new Promise(function(resolve, reject) {
-    ui.window.launchConfirmText(/** @type {!osx.window.ConfirmTextOptions} */ ({
+    osWindow.launchConfirmText(/** @type {!osx.window.ConfirmTextOptions} */ ({
       confirm: (accessToken) => {
         settings.getInstance().set(SettingsKey.ACCESS_TOKEN, accessToken);
         resolve(accessToken);
--- a/src/plugin/cesium/cesium.js
+++ b/src/plugin/cesium/cesium.js
@@ -1,23 +1,37 @@
 goog.module('plugin.cesium');
 
-const MapContainer = goog.require('os.MapContainer');
-const settings = goog.require('os.config.Settings');
 const Promise = goog.require('goog.Promise');
 const Uri = goog.require('goog.Uri');
 const TrustedResourceUrl = goog.require('goog.html.TrustedResourceUrl');
 const jsloader = goog.require('goog.net.jsloader');
+const GeometryType = goog.require('ol.geom.GeometryType');
 const olProj = goog.require('ol.proj');
 const Tile = goog.require('ol.source.Tile');
 const core = goog.require('olcs.core');
+
+const {ROOT} = goog.require('os');
+const MapContainer = goog.require('os.MapContainer');
 const AlertManager = goog.require('os.alert.AlertManager');
 const DisplaySetting = goog.require('os.config.DisplaySetting');
+const settings = goog.require('os.config.Settings');
+const osMap = goog.require('os.map');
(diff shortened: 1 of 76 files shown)


src/plugin/cesium/cesium.js
```diff
@@ -540,68 +518,48 @@ plugin.cesium.hasWorldTerrain = function() {
...
  * Enable the Cesium World Terrain provider, if configured.
  */
-plugin.cesium.enableWorldTerrain = function() {
-  const map = os.MapContainer.getInstance();
+const enableWorldTerrain = function() {
+  const map = MapContainer.getInstance();
   const renderer = map.getWebGLRenderer();
   if (renderer) {
     const supported = renderer.getSupportedTerrainProviders();
     const worldTerrain = supported.find((p) => p.type === os.map.terrain.TerrainType.ION);
```
jsalankey, line 529: Still using a global reference for `TerrainType` here.

src/plugin/cesium/cesiumcamera.js
```diff
@@ -377,545 +426,528 @@ plugin.cesium.Camera.prototype.initExtentComputation_ = function() {
...
+   */
+  flyTo(options) {
+    if (this.enabled_) {
+      var duration = options.duration != null ? options.duration : MapContainer.FLY_ZOOM_DURATION;
+      var easingFunction = options.flightMode === FlightMode.SMOOTH ?
+        Cesium.EasingFunction.LINEAR_NONE : undefined;
+
+      var target;
+      if (options.center) {
+        var center = proj.transform(options.center, osMap.PROJECTION, os.proj.EPSG4326);
```
jsalankey, line 553: Another global reference for `os.proj`. Given how long this PR is, there are probably a few of these left. Not sure if we want to bother picking through it for them or just dealing with them later.

src/plugin/cesium/cesium.js
```diff
@@ -145,54 +145,51 @@ plugin.cesium.addTrustedServer = function(url) {
...
  *
- * @return {!goog.Promise<string>}
+ * @return {!Promise<string>}
  */
-plugin.cesium.promptForAccessToken = function() {
-  return new goog.Promise(function(resolve, reject) {
-    os.ui.window.launchConfirmText(/** @type {!osx.window.ConfirmTextOptions} */ ({
+const promptForAccessToken = function() {
+  return new Promise(function(resolve, reject) {
+    ui.window.launchConfirmText(/** @type {!osx.window.ConfirmTextOptions} */ ({
```
jsalankey, line 190: Another `ui.window` case.

src/plugin/cesium/cesium.js
schmidtk: I started converting some globals, but we should write a transform to do that for us in an intelligent manner using the `deps.js` file. Otherwise it's going to add a large amount of tedious extra work across all of OpenSphere's sources.

src/plugin/cesium/cesiumcamera.js
schmidtk: Yeah we can write a transform to do these later.

src/plugin/cesium/cesiummenu.js
```diff
@@ -51,6 +50,12 @@ plugin.cesium.menu.launchAddIonAsset = function() {
     'modal': false
   };
 
-  var template = '<importionasset></importionasset>';
-  os.ui.window.create(windowOptions, template);
+  var template = `<${importIonAssetTag}></${importIonAssetTag}>`;
+  ui.window.create(windowOptions, template);
```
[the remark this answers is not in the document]
schmidtk: These should be fixed though, because they're referencing the wrong module.

src/plugin/cesium/cesium.js
schmidtk: Thanks, fixed.

src/plugin/cesium/cesiummenu.js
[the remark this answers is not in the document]
schmidtk: Fixed.

schmidtk:
There were some functions in `src/os/ui/icons.js` that were on the `os.ui` namespace, which caused similar problems to the noted `os.ui.window` references. To resolve this and hopefully better prep for converting that file, I moved all functions to `os.ui.icons`, provided that namespace, and added deprecated references to the old function paths. I fixed all references in OpenSphere, and external projects will see a deprecation warning if they're using the old functions.

jsalankey (approved):

jsalankey (approved):

PR #1228, opened 2021-04-07
refactor: add menu.unit.MENU; deprecate menu.UNIT

jkae:
Change for better ES6 module compatibility.

### DIFF
--- a/src/os/ui/menu/unitmenu.js
+++ b/src/os/ui/menu/unitmenu.js
@@ -9,17 +9,24 @@ goog.require('os.ui.menu.MenuItemType');
 /**
  * @type {os.ui.menu.Menu<undefined>|undefined}
  */
-os.ui.menu.UNIT = new os.ui.menu.Menu(new os.ui.menu.MenuItem({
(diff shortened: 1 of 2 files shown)


src/os/ui/menu/unitmenu.js
```diff
@@ -9,17 +9,24 @@ goog.require('os.ui.menu.MenuItemType');
...
-os.ui.menu.UNIT = new os.ui.menu.Menu(new os.ui.menu.MenuItem({
+os.ui.menu.unit.MENU = new os.ui.menu.Menu(new os.ui.menu.MenuItem({
   type: os.ui.menu.MenuItemType.ROOT,
   children: []
 }));
 
 
+/**
+ * @type {os.ui.menu.Menu<undefined>|undefined}
+ * @deprecated
```
jsalankey, line 20: Nit: deprecated tags can should have a comment after them to say what to do instead, i.e. "Please use goog.require('os.ui.menu.unit.MENU') instead". These comments get printed out by the compiler and make it easier to fix later.

jsalankey (approved):

PR #1229, opened 2021-04-07
IC Code Sanitation

Gamerlios123:


src/os/mainctrl.js
```diff
@@ -1,5 +1,6 @@
 goog.provide('os.MainCtrl');
 
+goog.require('coreui.mixin.ICSanitizeMixin');
```
bradh, line 3: Where is this coming from?

Gamerlios123: bits-internal

bradh: So that probably isn't going to work in OpenSphere, unless you can move it over?

Gamerlios123: Ah, does OpenSphere need a copy or will it work in bits-internal if moved to OpenSphere?

Gamerlios123: Ok, I moved the mixin into OpenSphere from bits and updated all the requires.

Gamerlios123: Looks like there was some confusion, this will work from bits-internal because it's a mixin not an export. I reverted moving the mixin.

wallw-bits, line 3: This line should be removed.

PR #1230, opened 2021-04-08
test(cesium): replace global refs in tests

schmidtk:
I added a new transform to [opensphere-jscodeshift](https://github.com/schmidtk/opensphere-jscodeshift) that finds/replaces global references in sources/tests. For source files, the transform simply adds a legacy `goog.require` then replaces it because the replace call already updates refs. For test files, a `goog.module.get` call is either found or added for each detected reference. The call is added directly to the root `describe()` call containing the reference, so subsequent references may reuse it.

To generate the changes in this PR, the transform was executed with:

```
yarn run shift -t src/transforms/es6/replaceglobals.js /path/to/opensphere/test/plugin/cesium
```

Global reference replacement is config driven by new config in `.jscodeshift.json`. The config defines the root namespaces to start from when searching for global refs, a list of modules to ignore (like `goog` which doesn't need to be required), and a map of module name to variable name for modules we want to specifically name instead of auto-generating the name.

### DIFF
--- a/.jscodeshift.json
+++ b/.jscodeshift.json
@@ -8,7 +8,9 @@
     "os.ui.filterManager": {"require": "os.query.FilterManager", "singleton": true},
     "os.layerConfigManager": {"require": "os.layer.config.LayerConfigManager", "singleton": true},
     "os.settings": {"require": "os.config.Settings", "singleton": true},
-    "os.map.mapContainer": {"require": "os.MapContainer", "singleton": true}
+    "os.map.mapContainer": {"require": "os.MapContainer", "singleton": true},
+    "os.queryManager": {"require": "os.query.QueryManager", "singleton": true},
+    "os.ui.queryManager": {"require": "os.query.QueryManager", "singleton": true}
   },
   "globalRootNamespaces": [
     "plugin",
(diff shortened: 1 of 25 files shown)


glennkitchellcaci (approved):
`os.ui` is the most problematic since it's flat out wrong 1/3 of the time, but it still helps out the other 2/3.
So I agree with removing it from the globals list and doing `os.ui.window` separately, but if you have any clever ideas to augment this new capability, (maybe some additional configs somehow for the top 2 or 3 os.ui.x) then that'd be much appreciated!

schmidtk:
> `os.ui` is the most problematic since it's flat out wrong 1/3 of the time, but it still helps out the other 2/3.

Yeah it does seem to be the primary offender. I'm not sure we need to do anything in the transforms, because the compiler will identify issues for us if we try referencing a symbol that doesn't actually exist in `os.ui`. That won't happen until the global namespace is gone, but it will error. Any files currently defining `os.ui.<something>` will have to be refactored when converted to a module, because they will no longer be capable of adding something to the `os.ui` module.

The short story here is, I don't think this script is the appropriate place to tackle this issue because it will just complicate something that needs to be fixed in another way. If this is a major problem, we could blacklist the `os.ui` namespace and skip those references for now.

schmidtk:
The script could be augmented to read in the detected module and check that the referenced expression actually exists, but that's going to be very complicated to handle properly because you're no longer just replacing a simple `MemberExpression`.

PR #1232, opened 2021-04-08
fix(featuresanitize): change feature importer sanitize to os.ui.sanitize

Gamerlios123:


schmidtk (commented):
This change is good as-is, but we might want to get rid of `os.im.FeatureImporter.sanitize` and just use `os.ui.sanitize` in its place. That would remove some unnecessary code from `MainCtrl` and generally be nice cleanup.

schmidtk (approved):
Thanks for the update, looks good!

schmidtk (approved):

PR #1234, opened 2021-04-09
Feat layer style extension

ameletyan:


src/os/layer/image.js
```diff
@@ -422,6 +443,273 @@ os.layer.Image.prototype.getSVGSet = function() {
...
+      source.removeImageFilter(this.colorFilter_);
+    }
+  }
+};
+
+
+/**
+ * @return {?Array<!osx.ogc.ImageStyle>}
+ */
+os.layer.Image.prototype.getStyles = function() {
```
schmidtk, line 609: Styles are specific to tile layers, where servers like WMS provide preset styles that may be applied to the layer. These styles change appearance of the tiles server-side. Image layers do not have a similar specification, so we do not need to duplicate all of that code here. We're only interested in the client-side transforms for brightness, contrast, etc.

src/os/mixin/imagemixin.js
```diff
@@ -0,0 +1,23 @@
+goog.provide('os.mixin.Image');
```
schmidtk, line 1: Please implement all new files using the `goog.module` format. We will have to transform all `goog.provide` files eventually, so we would prefer all new files start in this format.

schmidtk (changes requested):
Image layers do not have "styles" like tile layers, so please remove all related code. This should only add controls for client-side image adjustments.

views/layer/image.html
```diff
@@ -0,0 +1,88 @@
...
+            <td class="w-100">
+              <select class="custom-select" ng-model="style" ng-change="ctrl.onStyleChange()" ng-options="value.label for value in styles">
+              </select>
+            </td>
+          </tr>
+        </tbody>
+      </table>
+    </div>
+  </div>
+  <div ng-if="ctrl.showRefresh" class="card">
```
schmidtk, line 67: Image layers do not currently support refresh, so this block can be removed.

schmidtk (changes requested):
Functionally this works well in 2D. I had a few minor code cleanup comments, but this is currently not working properly in 3D. We will want to listen to layer style change events and update the image material so it renders the changes in Cesium.

For opacity, I believe you'll want to set the `color` property when [creating the image material](https://cesium.com/docs/cesiumjs-ref-doc/ImageMaterialProperty.html). The color should be white (255 RGB) with the opacity set as the alpha.

Before you do this, please merge the latest from master to pick up the recent changes to `ImageSynchronizer`.

views/layer/image.html
```diff
@@ -0,0 +1,67 @@
...
+          <tr title="Sets the sharpness of the layer(s)" ng-show="ctrl.showBasicColor">
+            <td>Sharpness</td>
+            <td class="w-100"><slider min="0" max="1" step="0.05" value="sharpness" name="sharpness"></slider></td>
+            <td class="text-right pl-2">
+              <button class="btn btn-secondary" ng-click="ctrl.reset('sharpness')" title="Restore default sharpness">
+                <i class="fa fa-undo"></i>
+              </button>
+            </td>
+          </tr>
+          <tr title="Sets the style of the layer(s)" ng-show="styles && items.length == 1">
```
schmidtk, line 56: Sorry I missed this before, but this isn't needed for image layers.

schmidtk (approved):
This looks great, thanks for the updates!

PR #1240, opened 2021-04-12
Clear Layer Feature Actions before Applying a Preset

arwatts16:
This will force feature actions to do what we would usually expect them to be doing by default, by explicitly unprocessing items before processing the new ones.

### DIFF
--- a/src/os/im/action/importactionmanager.js
+++ b/src/os/im/action/importactionmanager.js
@@ -52,6 +52,13 @@ os.im.action.ImportActionManager = function() {
   this.actionEntries = {};
 
   /**
+   * The last action entries applied.
+   * @type {!Object<string, !Array<!os.im.action.FilterActionEntry<T>>>}
+   * @protected
+   */
+  this.lastActionEntries = {};
+
+  /**
    * Registered import actions.
    * @type {!Object<string, !os.im.action.IImportAction<T>>}
    * @protected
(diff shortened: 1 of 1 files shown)


src/os/command/vectorlayerpresetcmd.js
```diff
@@ -98,13 +98,30 @@ os.command.VectorLayerPreset.prototype.applyFeatureActions = function(layer) {
...
     faIds.forEach(function(id) {
       faIdMap[id] = true;
     });
   }
 
   var entries = iam.getActionEntries(type);
+
+  // Force Feature Actions to act the way we want by disabling all feature
+  // actions on the layer before applying the new ones
+  entries.forEach(function(e) {
```
glennkitchellcaci, line 113: By putting this fix into the Preset command, it doesn't work when the User manually does what Presets do:
1. Apply a feature action that affects all the features
2. Disable that feature action
3. Enable another feature action that only affects some of the features
4. Click "Apply"
5. Verify the bug (styles from the disabled feature action) is still there
You should be able to solve both cases at once by updating how `ImportActionManager.getInstance().processItems()` works/is called. It accepts an opt_items and opt_unprocess, so if it unprocessed the old items first, then processed the new ones, it should work the same as what you've got here.

glennkitchellcaci (changes requested):

src/os/command/vectorlayerpresetcmd.js
arwatts16: I see what you're saying. I've changed the way `ImportActionManager.getInstance().processItems()` works by having it explicitly unprocess items rather than relying on the `processItemsProtected()` function to do it.

glennkitchellcaci: This works, but goes against the concept of consolidating the events that `processItemsProtected()` does for you.
I was thinking more along the lines that `processItemsProtected()` in ImportActionManager/FeatureActionManager would:

1. Remember the last FA's they applied, and if not null, unprocess those first (and save the events into the config), then
2. Still do what it used to do (i.e. build the config for the events so it throws one event for each Feature)

src/os/im/action/importactionmanager.js
```diff
@@ -399,7 +399,8 @@ os.im.action.ImportActionManager.prototype.processItemsProtected = function(
 os.im.action.ImportActionManager.prototype.processItems = function(entryType, opt_items, opt_unprocess) {
   var items = opt_items || this.getEntryItems(entryType);
   if (items && items.length > 0) {
-    this.processItemsProtected(entryType, items, opt_unprocess);
+    if (opt_unprocess) this.unprocessItems(entryType, items);
```
glennkitchellcaci, line 402: This works, but goes against the concept of consolidating the events that processItemsProtected() does for you.
I was thinking more along the lines that processItemsProtected() in ImportActionManager/FeatureActionManager would:
1. Remember the last FA's they applied, and if not null, unprocess those first (and save the events into the config), then
2. Still do what it used to do (i.e. build the config for the events so it throws one event for each Feature)

glennkitchellcaci (changes requested):

glennkitchellcaci (approved):

PR #1243, opened 2021-04-15
feat(track): allow track truncation by time / age out

bradh:
`os.track` supports truncation to a specified number of track points. I need something similar, but by age of the points.

That is, old points in the track will be removed if they "age out" beyond a specified time.

The PR includes some minor refactoring and jsdoc fixes.

### DIFF
--- a/src/os/track.js
+++ b/src/os/track.js
@@ -481,9 +481,7 @@ os.track.addToTrack = function(options) {
         track.getGeometry());
 
     // merge the split line so coordinates can be added in the correct location
-    geometry.toLonLat();
-    geometry = os.geo.mergeLineGeometry(geometry);
-    geometry.osTransform();
+    os.track.mergeSplitGeometry_(geometry);
 
     var flatCoordinates = geometry.flatCoordinates;
     var stride = geometry.stride;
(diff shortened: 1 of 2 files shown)


src/os/track.js
```diff
@@ -609,25 +605,71 @@ os.track.truncate = function(track, size) {
...
+
+  if (geometry.getType() === ol.geom.GeometryType.MULTI_LINE_STRING) {
+    // merge the split line so coordinates can be counted and truncated cleanly
+    os.track.mergeSplitGeometry_(geometry);
+  }
+  var coords = geometry.getCoordinates();
+  var timeNow = Date.now();
+  var pointsToKeep = 0;
+  if (coords) {
+    coords.reverse().forEach(function(c) {
```
schmidtk, line 650: `os.track.getTimeIndex` will get the closest index to a specified timestamp using binary search, so it may be more performant to:

- Get the closest index to the age-off timestamp.
- Splice the flat coordinates array from the index to the end.
- Do the same for metadata.

This would avoid the `getCoordinates` call which reconstitutes the flat coordinates (unnecessary GC of all those arrays), and reversing/iterating over the entire array.

src/os/track.js
```diff
@@ -609,25 +605,71 @@ os.track.truncate = function(track, size) {
...
+      if (ageForThisCoord < maximumAge) {
+        pointsToKeep = pointsToKeep + 1;
+      }
+    });
+  }
+  var numCoords = pointsToKeep * geometry.stride;
+  os.track.prunePointsAt_(track, geometry, numCoords);
+};
+
+os.track.prunePointsAt_ = function(track, geometry, numCoords) {
```
schmidtk, line 662: Missing JSDoc for this new function.

schmidtk (changes requested):

schmidtk:
@bradh FYI Travis stopped building OpenSphere, so you'll have to merge this branch with master to pick up the GH Actions builds.

PR #1244, opened 2021-04-15
Upgrade Closure compiler/library to latest

schmidtk:
**Draft Requirements:**
- [x] Merge ngageoint/opensphere-build-closure-helper/pull/27
- [x] Merge ngageoint/opensphere-build-resolver/pull/78
- [x] Update versions in `package.json`

This PR upgrades the Closure compiler to 20210406.0.0 (via `opensphere-build-closure-helper`) and the library to 20210302.0.0.

**Changes**

- Updated jQuery externs to 3.3, provided by Closure.
- Updated Angular externs to 1.6, provided by Closure with our local edits merged in.
- Refactored use of the `goog.log` API to reflect changes in the Closure library.
- Refactored `goog.Uri.QueryData` API to reflect changes in the Closure library.
- Refactored `goog.net.Cookies` API to reflect changes in the Closure library.
- Replace uses of `goog.base` in tests. This has been deprecated and removed since the last library update.
- Fixed additional build errors identified by the updated compiler.
- Replaced `goog.nullFunction` (deprecated) with `os.fn.noop`.
- Replaced `goog.bind` (deprecated) with native `Function#bind`.
- Simplified Job tests to no longer load additional scripts in the worker. This will hopefully make the tests more reliable.


externs/Cesium.externs.js
```diff
@@ -523,6 +523,13 @@ Cesium.BillboardCollection.prototype.add = function(options) {};
 Cesium.BillboardCollection.prototype.contains = function(opt_bb) {};
 
 
+/**
+ * @param {number} index
+ * @return {Cesium.Billboard} .
+ */
+Cesium.BillboardCollection.prototype.get = function(index) {};
```
csgibson, line 530: Why was this needed?

package.json
```diff
@@ -46,10 +46,10 @@
       "language_in": "ECMASCRIPT_2018",
```
csgibson, line 46: Should this be bumped too?

externs/Cesium.externs.js
schmidtk: It's used by `plugin.cesium.VectorContext` but the old compiler didn't identify that it was missing from externs.

package.json
schmidtk: Yes, thank you! Fixed.

schmidtk:
Rebased to get up to date and trigger a build. The build failed once because it's using Node stable, which is now 16.0.0 and `node-sass@4.x` is not supported with Node 16. I switched the Travis build to use Node LTS, which will be 14.x until October. Second build failure seemed to be an infrequent timing issue with our Job tests.

schmidtk:
`goog.nullFunction` has been deprecated. They recommend replacing with an inline empty function, but we already have `os.fn.noop` so I replaced instances with that.

schmidtk:
`goog.bind` has also been deprecated, and should be replaced with the native `bind`. `goog.partial` is still fine to use if you don't need to change the `this` binding. Updated all `goog.bind` references to resolve those warnings.

glennkitchellcaci (approved):

PR #1245, opened 2021-04-22
Feat save modify

jsalankey:
Adds support for exporting files with Opensphere as the target to save them to. This includes adding support to all current file types (as well as Geopackage) to support export and direct reimport back into the tool. This can be done automatically for file layers or by accessing the "Save" menu option on a layer with pending changes.

Also adds a basic File Settings plugin which currently only has one option, but that we want to add to in the future for better file management.

### DIFF
--- a/src/os/data/filedescriptor.js
+++ b/src/os/data/filedescriptor.js
@@ -5,10 +5,14 @@ goog.require('os.command.LayerRemove');
 goog.require('os.data.IReimport');
 goog.require('os.data.IUrlDescriptor');
 goog.require('os.data.LayerSyncDescriptor');
+goog.require('os.ex.IExportMethod');
+goog.require('os.file.File');
 goog.require('os.file.FileStorage');
 goog.require('os.im.ImportProcess');
 goog.require('os.implements');
 goog.require('os.parse.FileParserConfig');
+goog.require('os.source');
+goog.require('os.source.Vector');
 goog.require('os.ui.file.ui.defaultFileNodeUIDirective');
 goog.require('os.ui.im.ImportEvent');
 goog.require('os.ui.im.ImportEventType');
@@ -323,6 +327,67 @@ os.data.FileDescriptor.prototype.reimport = function() {
   process.setConfig(this.getParserConfig());
   process.setSkipDuplicates(true);
   process.begin();
+};
+
+
+/**
+ * Get the exporter method associated with this file type.
+ * @return {?os.ex.IExportMethod}
+ */
+os.data.FileDescriptor.prototype.getExporter = function() {
+  return null;
+};
+
+
+/**
+ * Handles changes to the underlying layer data. Updates the file in storage.
+ * @param {os.ex.ExportOptions} options
+ */
+os.data.FileDescriptor.prototype.onDataChange = function(options) {
+  const source = /** @type {os.source.Vector} */ (options.sources[0]);
+  const exporter = this.getExporter();
+
+  if (exporter) {
+    const fields = os.source.getExportFields(source, false, exporter.supportsTime());
+    const name = this.getTitle() || 'New File';
+    exporter.setItems(options.items);
+    exporter.setFields(fields);
+    exporter.setName(name);
+
+    if (exporter.isAsync()) {
+      var et = /** @type {goog.events.EventTarget} */ (exporter);
+      et.listen(os.events.EventType.COMPLETE, (event) => {
+        const result = event.target.getOutput();
+        const file = new os.file.File();
+        file.setFileName(name);
+        file.setUrl(os.file.getLocalUrl(name));
+        file.setContent(result);
+        file.setContentType(exporter.getMimeType());
+
+        // always replace. if we got here the application should have done duplicate file detection already.
+        const fs = os.file.FileStorage.getInstance();
+        fs.storeFile(file, true);
+      }, false, this);
+      et.listen(os.events.EventType.ERROR, (event) => {
+
+      }, false, this);
+
+      exporter.process();
+    } else {
+      exporter.process();
+
+      const result = exporter.getOutput();
+      const file = new os.file.File();
+      file.setFileName(name);
+      file.setUrl(os.file.getLocalUrl(name));
+      file.setContent(result);
+      file.setContentType(exporter.getMimeType());
+
+      // always replace. if we got here the application should have done duplicate file detection already.
+      const fs = os.file.FileStorage.getInstance();
+      fs.storeFile(file, true);
+    }
+  }
 };
 
 
--- a/src/os/data/filedescriptor.js
+++ b/src/os/data/filedescriptor.js
@@ -68,11 +68,20 @@ os.data.FileDescriptor = function() {
    */
   this.date_ = null;
 
+  this.log = os.data.FileDescriptor.LOGGER_;
   this.descriptorType = 'file';
 };
 goog.inherits(os.data.FileDescriptor, os.data.LayerSyncDescriptor);
 os.implements(os.data.FileDescriptor, 'os.data.IReimport');
 os.implements(os.data.FileDescriptor, os.data.IUrlDescriptor.ID);
+
+
+/**
+ * @type {goog.log.Logger}
+ * @private
+ * @const
+ */
+os.data.FileDescriptor.LOGGER_ = goog.log.getLogger('os.data.FileDescriptor');
 
 
 /**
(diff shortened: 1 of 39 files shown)


src/os/file/filesettingsui.js
```diff
@@ -0,0 +1,93 @@
...
+
+
+/**
+ * The column mapping settings UI directive
+ * @return {angular.Directive}
+ */
+const directive = () => ({
+  restrict: 'E',
+  replace: true,
+  templateUrl: os.ROOT + 'views/file/filesettings.html',
```
schmidtk, line 14: File is missing a require for this.

src/plugin/file/shp/shpdescriptor.js
```diff
@@ -116,6 +118,25 @@ plugin.file.shp.SHPDescriptor.prototype.clearData = function() {
...
+
+
+/**
+ * @inheritDoc
+ */
+plugin.file.shp.SHPDescriptor.prototype.onDataChange = function(options) {
+  plugin.file.shp.SHPDescriptor.base(this, 'onDataChange', options);
+
+  // ensure that the URL is set correctly in case we are converting from a shp/dbf file to a zip shp file
+  this.setUrl2('');
```
schmidtk, line 136: Is this the best place to make this correction, or could we do it closer to where the shp/dbf -> zip actually happens?

schmidtk (changes requested):

src/os/data/filedescriptor.js
```diff
@@ -326,6 +340,71 @@ os.data.FileDescriptor.prototype.reimport = function() {
...
+
+  if (e instanceof os.events.PropertyChangeEvent) {
+    const layer = /** @type {os.layer.Vector} */ (e.target);
+    const p = e.getProperty() || '';
+
+    if (p == os.source.PropertyChange.HAS_MODIFICATIONS) {
+      if (layer instanceof os.layer.Vector) {
+        const source = /** @type {os.source.Vector} */ (layer.getSource());
+
+        if (os.settings.get('os.file.autoSaveFiles', true) && source instanceof os.source.Vector) {
```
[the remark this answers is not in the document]
jsalankey: Yep, will fix. Also want to put in a constant for the settings key.

src/os/data/layernode.js
```diff
@@ -351,6 +352,12 @@ os.data.LayerNode.prototype.formatValue = function(value) {
 
     if (layer instanceof os.layer.Vector) {
       s += ' <featurecount></featurecount>';
+
+      var source = layer.getSource();
+      if (source instanceof os.source.Vector && source.getHasModifications()) {
+        s = `<span title="This layer has unsaved changes. Right click to save them." 
+            style="font-weight:900;">  •  </span>${s}`;
```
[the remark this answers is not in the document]
jsalankey: Yep. There might be a better way to draw this indicator too, the font-weight just helped make it a little more noticeable. Didn't even know Bootstrap had this.

src/os/ui/file/exportmanager.js
```diff
@@ -198,27 +196,56 @@ os.ui.file.ExportManager.prototype.doExport_ = function(items, fields, title, ex
...
+      }
     }
 
-    persister.save(name, result, exporter.getMimeType());
+    if (options.persister) {
+      options.persister.save(name, result, exporter.getMimeType());
+    } else {
+      try {
+        if (result instanceof Blob) {
+          await result.arrayBuffer().then((arrayBuffer) => result = arrayBuffer);
```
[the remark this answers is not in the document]
jsalankey: I wasn't aware of that, neat. 

I needed to revisit this function anyway because `async` functions piss of the compiler when they are passed as the handler to a `listen` function like this one is above. It's technically a type error because `async` functions always return a promise, which is doesn't match the allowed function type for the listen handler.

src/plugin/file/csv/csvdescriptor.js
```diff
@@ -124,6 +127,31 @@ plugin.file.csv.CSVDescriptor.prototype.setUseHeader = function(useHeader) {
 };
 
 
+/**
+ * @inheritDoc
+ */
+plugin.file.csv.CSVDescriptor.prototype.getExporter = function() {
+  return new plugin.file.csv.CSVExporter();
```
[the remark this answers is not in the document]
jsalankey: I didn't consider that, but it might save a little bit of processing. Only concern would be if there's some level of statefulness to the exporter that could cause issues. `ExportManager` for example keeps instances around, but it reconstructs new instances when you call `getExportMethods()`.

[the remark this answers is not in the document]
schmidtk: In theory you should just need to call the `reset` function on the exporter to get back to the default state. This doesn't seem like a notable amount of overhead, just a thought if it's a relatively simple change.

src/plugin/file/shp/ui/zipshpimportui.js
```diff
@@ -63,9 +67,21 @@ plugin.file.shp.ui.ZipSHPImportUI.prototype.launchUI = function(file, opt_config
...
+    zip.createReader(new zip.ArrayBufferReader(content),
+        this.handleZipReader.bind(this), this.handleZipReaderError.bind(this));
+  } else {
+    // convert the blob to an ArrayBuffer and proceed down the normal path
+    content.arrayBuffer().then((arrayBuffer) => {
+      this.zipFile_.setContent(arrayBuffer);
+      zip.createReader(new zip.ArrayBufferReader(arrayBuffer),
+          this.handleZipReader.bind(this), this.handleZipReaderError.bind(this));
+    }, (reason) => {
+      // womp womp
```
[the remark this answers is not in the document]
jsalankey: I might no longer need this `content.arrayBuffer()` call.  think the call in `ExportManager` covers all the cases of the data being a `Blob` when it gets here. If not, I'll update this.

src/plugin/file/shp/shpdescriptor.js
jsalankey: I'm not sure that I know where that place would be. The SHP exporter isn't aware of the underlying file structure backing a layer when it creates the ZIP to export. It's only in the descriptor on this new code path that we would want to ensure that we're not still loading in a DBF file that we no longer need.

schmidtk: Ok, as long as we're certain this path will always be saving a ZIP and not blowing away the DBF URL then I don't have any issue putting it here.

src/os/ui/window/confirmtext.js
```diff
@@ -51,6 +52,10 @@ os.ui.window.ConfirmTextCtrl = function($scope, $element) {
     }
   });
 
+  $timeout(function() {
+    $element.find('.js-confirm-input').focus();
```
schmidtk, line 56: Can this be married up with the `select` bit above? The intent of that is to select text in the input, and this is focusing the input. These should probably happen in the same block, and the existing `select` bit updated to use the new selector. Do these work if you call them in `$onInit`, or is that too early?

src/plugin/file/shp/shpdescriptor.js
```diff
@@ -116,6 +119,30 @@ plugin.file.shp.SHPDescriptor.prototype.clearData = function() {
...
+ * @inheritDoc
+ */
+plugin.file.shp.SHPDescriptor.prototype.onFileChange = function(options) {
+  plugin.file.shp.SHPDescriptor.base(this, 'onFileChange', options);
+
+  // ensure that the URL is set correctly in case we are converting from a shp/dbf file to a zip shp file
+  const url2 = this.getUrl2();
+  if (url2 && os.file.isLocal(url2)) {
+    const fs = os.file.FileStorage.getInstance();
+    fs.deleteFile(url2);
```
schmidtk, line 140: Ah good call. Does the SHP already get cleaned up from storage?

src/plugin/file/shp/ui/zipshpimportui.js
```diff
@@ -63,9 +67,21 @@ plugin.file.shp.ui.ZipSHPImportUI.prototype.launchUI = function(file, opt_config
...
 
-  var content = /** @type {ArrayBuffer} */ (file.getContent());
-  zip.createReader(new zip.ArrayBufferReader(content),
-      this.handleZipReader.bind(this), this.handleZipReaderError.bind(this));
+  const content = /** @type {ArrayBuffer|Blob} */ (file.getContent());
+
+  if (content instanceof ArrayBuffer) {
+    zip.createReader(new zip.ArrayBufferReader(content),
+        this.handleZipReader.bind(this), this.handleZipReaderError.bind(this));
+  } else if (content instanceof ArrayBuffer) {
```
schmidtk, line 75: This condition is the same as the `if`.

src/os/ui/window/confirmtext.js
jsalankey: Oh yeah, didn't even notice that was using `setTimeout`, just assumed there wasn't one since `$timeout` wasn't injected.

I can try the `$onInit` also, but I probably won't spend too much time messing with Angular stuff since this was just a nice to have that I tossed in along the way.

src/plugin/file/shp/shpdescriptor.js
jsalankey: That's a good point, I'll check. In the other cases when this code is hit, the underlying filename doesn't change, so its just overwritten and uses the same local URL.

src/plugin/file/shp/ui/zipshpimportui.js
jsalankey: Woops, meant to explicitly check `Blob` here.

src/os/ui/window/confirmtext.js
schmidtk: Yeah no worries, I'm not even sure `$onInit` would work for these calls because I don't think the element is actually in the DOM yet.

jsalankey: `$onInit` works, should be cleaned up.

src/plugin/file/shp/shpdescriptor.js
jsalankey: The .shp file does get overwritten and the .dbf gets removed. Verified by watching IDB from dev tools.

PR #1249, opened 2021-04-23
feat(trackpredict): Add niceties to building and extending tracks

glennkitchellcaci:
Right-click a Track or Point, click "Predict..." > "Track to..." to build a Track from that starting location.

NOTE: This capability is toggled off in this commit. "Time" currently extends from the selected track by 1 hour per point added in the interaction; a future commit will make that configurable.

### DIFF
--- a/src/plugin/track/confirmtrack.js
+++ b/src/plugin/track/confirmtrack.js
@@ -1,16 +1,11 @@
 goog.module('plugin.track.ConfirmTrackUI');
 goog.module.declareLegacyNamespace();
 
-const os = goog.require('os');
-const WindowEventType = goog.require('os.ui.WindowEventType');
+const {ROOT} = goog.require('os');
 const KMLField = goog.require('plugin.file.kml.KMLField');
-const pluginTrack = goog.require('plugin.track');
-
-const Promise = goog.require('goog.Promise');
-const osTrack = goog.require('os.track');
 const Module = goog.require('os.ui.Module');
-const osWindow = goog.require('os.ui.window');
 const PlacesManager = goog.require('plugin.places.PlacesManager');
+const WindowEventType = goog.require('os.ui.WindowEventType');
 
 const OlFeature = goog.requireType('ol.Feature');
 
@@ -22,7 +17,7 @@ const OlFeature = goog.requireType('ol.Feature');
  */
 const directive = () => ({
   restrict: 'E',
-  templateUrl: os.ROOT + 'views/plugin/track/confirmtrack.html',
+  templateUrl: ROOT + 'views/plugin/track/confirmtrack.html',
   controller: Controller,
   controllerAs: 'confirm'
 });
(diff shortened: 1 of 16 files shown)


src/plugin/track/confirmtrack.js
```diff
@@ -1,101 +1,108 @@
...
 /**
  * Prompt the user to choose a track.
  *
- * @return {!goog.Promise}
+ * @return {!Promise}
  */
-plugin.track.promptForTrack = function() {
-  return new goog.Promise(function(resolve, reject) {
-    plugin.track.launchConfirmTrack(resolve, reject);
+pluginTrack.promptForTrack = function() {
```
schmidtk, line 92: The `plugin.track` functions in this file will need to be refactored given this is not the `plugin.track` module. They'll either need to be moved, or exported by this module and references updated.

src/plugin/track/trackplugin.js
```diff
@@ -1,32 +1,48 @@
...
+  /**
+   * @inheritDoc
+   */
+  init() {
+    const predict = (settings.get(track.PREDICT, false) === true);
+
+    menu.layerSetup(predict);
+    menu.spatialSetup(predict);
+
+    ui.metricsManager.addMetricsPlugin(new Metrics());
```
schmidtk, line 37: This global ref will need to be updated. I added this to `.jscodeshift.json` very recently, so it was missed on your transform pass.

src/plugin/track/track.js
```diff
@@ -445,88 +420,104 @@ plugin.track.getTrackPositionAt = os.track.getTrackPositionAt;
...
   return;
 };
 
+/**
+ * Creates a track and adds it to the Saved Places layer.
+ *
+ * @param {!osTrack.CreateOptions} options The options object for the track.
+ * @return {osTrack.TrackFeatureLike|undefined} The track feature.
+ */
+const createAndAdd = function(options) {
```
schmidtk, line 478: Ahh I was wondering how we should handle these functions we may share with external windows. This is straightforward enough, and best of all non-breaking. I like it! 👍

schmidtk (dismissed):
Started transforming this plugin before I realized it was already in work. Had a few comments based on things I came across running the scripts.

src/plugin/track/trackplugin.js
glennkitchellcaci: ty!

src/plugin/track/confirmtrack.js
glennkitchellcaci: Moved over to TrackManager without causing any circular dependencies. Updated for the breaking change downstream.

src/plugin/track/track.js
glennkitchellcaci: If I find more Track functions that get alternate implementations like this, I might upgrade to a `CollectionManager` or a `Registry` for the `EventTarget` powers; but this is OK for here and any downstream spots at the moment. Let me know if I should pull the trigger and upgrade to that or another pattern.

schmidtk: If anything I would say put the function on `TrackManager` and share that with external windows instead. You can leave the original function intact, and make the `TrackManager` function a wrapper in support of this behavior.

glennkitchellcaci: I'm not sure what you mean -- If I restore createAndAdd() in TrackPlugin to what it was and in TrackManager put...
```
createAndAdd(){ this.createAndAdd_(); }
setCreateAndAdd(f) { this.createAndAdd_ = (f || TrackPlugin.createAndAdd); } 
```
...wouldn't I then have to update everywhere that calls `TrackPlugin.createAndAdd()` to instead call `TrackManager.createAndAdd()`?

OR

If I keep createAndAdd, setCreateAndAdd, and createAndAdd_ in TrackPlugin, add setCreateAndAdd to TrackManager, and call the one on TrackPlugin, wouldn't I hthen ave to also move `TrackPlugin.updateTrackSource()` and then update everywhere that uses that? That's a much smaller list at least.

schmidtk: The current solution is great because it's a non-breaking change, no issue sticking with that.

If you want to make a breaking API change for the sake of a cleaner solution, then calling `TrackManager.createAndAdd()` would be my suggestion so we no longer have to export individual functions for external windows or replace the `createAndAdd` call which feels a bit hacky. We would only have to export `TrackManager` and call any functions we need to run in the main window from that.

bcleveland1022 (approved):

PR #1250, opened 2021-04-23
build(gh): move build to github actions

schmidtk:
travis-ci.org is no longer building for free, so we'll move to GitHub Actions instead.

- The Yarn cache step saves the cache based on a hash of the `yarn.lock`, so that was added to the repo.
- Commitlint is a separate action from the build, for convenience. This follows the pattern established in the build projects.
- Scripts designed for Travis were removed and replaced within `build.yml`. The primary difference is that the surge deployment only happens on the `master` build now, but I don't believe we need to keep track of deployments for anything else.


.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
+name: Build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
wallw-bits, line 7: We've had a couple of PRs to `release`, shouldn't that get handled here as well?

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,15 @@
+name: Lint Commit Messages
+
+on:
+  pull_request:
+    branches: [ master ]
```
wallw-bits, line 5: Same here.

wallw-bits (changes requested):

.github/workflows/build.yml
schmidtk: Yes, thank you. Fixed.

.github/workflows/commitlint.yml
schmidtk: Also fixed.

wallw-bits (approved):

PR #1252, opened 2021-04-24
Fix html lint warnings

schmidtk:
- Replaced inline `style` with CSS classes.
- Allowed `ng-style`. This directive is useful for dynamically changing styles based on scope/controller values.
- Replaced `<input type="submit" hidden>` with `<button type="submit" hidden></button>` per `htmlvalidate` suggestion. I also fixed some cases where Enter was not submitting a form, such as the generic import dialog and server import UI's.
- Updated `htmlvalidate` config to error on checks that have been resolved.

### DIFF
--- a/views/file/importdialog.html
+++ b/views/file/importdialog.html
@@ -1,25 +1,31 @@
 <div class="d-flex flex-column flex-fill">
-  <div ng-form="importForm" class="modal-body">
+  <form class="modal-body" name="importForm" ng-submit="importdialog.accept()">
     <div class="input-group">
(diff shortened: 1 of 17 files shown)


views/file/importdialog.html
```diff
@@ -1,25 +1,31 @@
 <div class="d-flex flex-column flex-fill">
-  <div ng-form="importForm" class="modal-body">
+  <form class="modal-body" name="importForm" ng-submit="importdialog.accept()">
```
jsalankey, line 2: Not sure if this will have any side effects, but we typically have switched to using `ng-form` in more places. It allows for Angular to support sub-forms for validation purposes, where for some reason validation tends to break when you nest standard `<form>` elements

jsalankey (approved):

PR #1254, opened 2021-04-26
Refactor plugins to use goog.module

schmidtk:
Transforms google, pelias, openpage, position, suncalc, weather, and xyz plugins to use `goog.module`.

### DIFF
--- a/src/plugin/google/places/attrcard.js
+++ b/src/plugin/google/places/attrcard.js
@@ -1,8 +1,7 @@
-goog.provide('plugin.google.places.AttrCardCtrl');
-goog.provide('plugin.google.places.attrCardDirective');
+goog.module('plugin.google.places.AttrCardUI');
 
-goog.require('os');
-goog.require('os.ui.Module');
+const {ROOT} = goog.require('os');
+const Module = goog.require('os.ui.Module');
 
 
 /**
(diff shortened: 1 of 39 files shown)


src/plugin/google/places/attrcard.js
```diff
@@ -1,48 +1,61 @@
...
+class Controller {
+  /**
+   * Constructor.
+   * @param {!angular.Scope} $scope
+   * @param {!angular.JQLite} $element
+   * @ngInject
+   */
+  constructor($scope, $element) {
+    // just plop the result HTML in
+    var attrs = /** @type {plugin.google.places.AttrResult} */ ($scope['result']).getResult();
```
jsalankey, line 46: Not sure if we added scripts for global replacement of types, but we can come back on these later.

src/plugin/google/places/resultcard.js
```diff
@@ -1,47 +1,55 @@
...
+ * @type {string}
+ */
+const directiveTag = 'googleplacesresultcard';
 
 
 /**
  * Register the directive.
  */
-os.ui.Module.directive('googleplacesresultcard', [plugin.google.places.resultCardDirective]);
+Module.directive('googleplacesresultcard', [directive]);
```
jsalankey, line 31: Nit: this should use the `directiveTag` instead of the raw string.

src/plugin/pelias/geocoder/attrcard.js
```diff
@@ -1,48 +1,62 @@
...
+ */
+const directiveTag = 'peliasgeocoderattrcard';
 
 
 /**
  * Register the directive.
  */
-os.ui.Module.directive('peliasgeocoderattrcard', [plugin.pelias.geocoder.attrCardDirective]);
-
+Module.directive('peliasgeocoderattrcard', [directive]);
```
jsalankey, line 32: Same directive tag comment as above. Might be something easy to tweak in the scripts.

src/plugin/pelias/geocoder/search.js
```diff
@@ -1,152 +1,145 @@
...
+const GeoJSON = goog.require('ol.format.GeoJSON');
+const olProj = goog.require('ol.proj');
+const MapContainer = goog.require('os.MapContainer');
+const Settings = goog.require('os.config.Settings');
+const osExtent = goog.require('os.extent');
+const osMap = goog.require('os.map');
+const osProj = goog.require('os.proj');
+const AbstractUrlSearch = goog.require('os.search.AbstractUrlSearch');
+const {ID} = goog.require('plugin.pelias.geocoder');
+const Result = goog.require('plugin.pelias.geocoder.Result');
```
jsalankey, line 13: Nit that isn't specific to here: the requires aren't being sorted by the script. We haven't been doing this consistently, but I did like it before.

src/plugin/position/copyposition.js
```diff
@@ -1,117 +1,120 @@
...
+});
 
 
 /**
- * Add the directive to the module
+ * The element tag for the directive.
+ * @type {string}
  */
-os.ui.Module.directive('copyPosition', [plugin.position.copyPositionDirective]);
+const directiveTag = 'copy-position';
```
jsalankey, line 41: Makes sense now why we aren't using this to register the directive. Woops.

src/plugin/suncalc/lightstripsettingsui.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * @param {string=} opt_new
+   * @param {string=} opt_old
+   * @protected
+   */
+  onSettingsChanged(opt_new, opt_old) {
+    if (opt_new && opt_old && opt_new !== opt_old) {
+      if (settings.getInstance()) {
```
jsalankey, line 77: Nit: this is a weird check to have now.

jsalankey (approved):

src/plugin/google/places/attrcard.js
schmidtk: Type-only references are tricky to detect because comments are simply text in AST. You can sort of do a regex detect, but that gets ugly fast with all the variations you might see in `@type` or `@param` annotations with `|`, `?`, etc. I think there is a library to parse out JSDoc comments in a way that would make this easier, but I haven't really looked into it. Currently types in comments are only replaced if the type is actually referenced in code where we can detect it with AST.

src/plugin/google/places/resultcard.js
schmidtk: The script doesn't use `directiveTag` here because that property is the element tag for the directive, which will not be the same as the `Module.directive` value if camel case is used. For example, a directive called `myUi` would have a directive tag of `my-ui`.

src/plugin/pelias/geocoder/search.js
schmidtk: I agree, I need to make a sort that works with assigned require statements. I also need to find a way to do this automatically in an editor. It's less straightforward now that the lines don't naturally sort since we want to sort by the required module name.

src/plugin/suncalc/lightstripsettingsui.js
schmidtk: Agree, I didn't do a cleanup pass on the code outside of things that didn't compile properly.

PR #1255, opened 2021-04-26
Refactor more plugins to use goog.module

schmidtk:
Transforms heatmap, nominatim, overview, params, storage, and vectortools plugins to use `goog.module`.

The `vectortools` plugin also had a slight refactor to move the UI launch methods to a new file. This was cleaner from a dependency standpoint.

### DIFF
--- a/src/plugin/heatmap/cmd/gradientcmd.js
+++ b/src/plugin/heatmap/cmd/gradientcmd.js
@@ -1,54 +1,54 @@
-goog.provide('plugin.heatmap.cmd.Gradient');
-goog.require('os.command.AbstractStyle');
+goog.module('plugin.heatmap.cmd.Gradient');
 
+const AbstractStyle = goog.require('os.command.AbstractStyle');
 
 
 /**
  * Changes the gradient of a heatmap.
- *
- * @extends {os.command.AbstractStyle}
- * @param {string} layerId
- * @param {number} value
- * @param {number=} opt_oldValue
- * @constructor
  */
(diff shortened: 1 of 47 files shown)


src/plugin/heatmap/cmd/gradientcmd.js
```diff
@@ -1,54 +1,54 @@
...
+  constructor(layerId, value, opt_oldValue) {
+    super(layerId, value, opt_oldValue);
+    this.title = 'Change heatmap gradient';
+  }
 
+  /**
+   * @inheritDoc
+   */
+  getOldValue() {
+    var layer = /** @type {plugin.heatmap.Heatmap} */ (this.getLayer());
```
jsalankey, line 25: Missing type import.

src/plugin/heatmap/heatmapsynchronizertype.js
```diff
@@ -0,0 +1,8 @@
+goog.module('plugin.heatmap.SynchronizerType');
```
jsalankey, line 1: Nit: looks like multiple provides got split into multiple files that could be consolidated into a general `plugin.heatmap` module.

jsalankey (approved):

src/plugin/heatmap/heatmapsynchronizertype.js
schmidtk: Yeah the script attempts to split out the provides in a way that will not be a breaking change. If something has a specific `provide`, we want `goog.require` to continue to work for that same name so a refactor like this would constitute a breaking change.

PR #1257, opened 2021-04-27
fix(net): proxy handling only enabled if configuration is present

jkae:
BREAKING: old proxyURL setting is no longer respected


src/os/net/net.js
```diff
@@ -138,7 +138,7 @@ os.net.addDefaultHandlers = function() {
   os.net.RequestHandlerFactory.addHandler(os.net.LocalFileHandler);
   os.net.RequestHandlerFactory.addHandler(os.net.SameDomainHandler);
   os.net.RequestHandlerFactory.addHandler(os.net.ExtDomainHandler);
-  os.net.RequestHandlerFactory.addHandler(os.net.ProxyHandler);
+  // No longer adds os.net.ProxyHandler here - depends on configuration
```
schmidtk, line 141: The `goog.require` should be removed as well. I don't think the comment is necessary. If someone is curious where the proxy handler gets loaded, they'll find it quickly via search.

src/os/ui/abstractmainctrl.js
```diff
@@ -236,11 +238,12 @@ os.ui.AbstractMainCtrl.prototype.initialize = function() {
     os.settings.set(['about', 'version'], this.scope['version']);
   }
 
-  // set the proxy url
-  os.net.ProxyHandler.PROXY_URL = /** @type {string} */ (os.settings.get(['proxyUrl'], os.net.ProxyHandler.PROXY_URL));
-
-  // set up the proxy the new way
-  os.net.ProxyHandler.configure(/** @type {?Object} */ (os.settings.get(['proxy'])));
+  // configure the proxy via settings & only add the handler if it's successful
+  if (os.net.ProxyHandler.configure(/** @type {?Object} */ (os.settings.get(['proxy'])))) {
```
schmidtk, line 242: I don't think we need to get rid of the `proxyUrl` setting. I'm not sure it's used anywhere, but no reason we can't continue to support it. I would get `proxyUrl` and `proxy` from settings, and if either is defined go ahead and add the proxy handler and call the original two lines. Then `os.net.ProxyHandler.configure` doesn't need to change, which doesn't seem necessary.

schmidtk (changes requested):

schmidtk (approved):

PR #1260, opened 2021-05-04
feat(trackpredict): initiate TrackInteraction automatically

glennkitchellcaci:
begin the TrackInteraction as soon as user clicks the MenuItem

### DIFF
--- a/src/plugin/track/trackinteraction.js
+++ b/src/plugin/track/trackinteraction.js
@@ -7,6 +7,7 @@ const MeasureInteraction = goog.require('os.interaction.Measure');
 const OlStroke = goog.require('ol.style.Stroke');
 const OlStyle = goog.require('ol.style.Style');
 
+const MapBrowserEvent = goog.requireType('ol.MapBrowserEvent');
 const OlLineString = goog.requireType('ol.geom.LineString');
 
 /**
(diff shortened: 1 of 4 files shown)


src/plugin/track/trackinteraction.js
```diff
@@ -99,6 +101,40 @@ class TrackInteraction extends MeasureInteraction {
...
+    this.setEnabled(toggle);
+    this.setActive(toggle);
+
+    if (toggle && opt_mapBrowserEvent) {
+      let event;
+
+      // prime the coordinates of the pointerdown event for the DrawInteraction part of TrackInteraction
+      event = opt_mapBrowserEvent;
+      event['type'] = 'pointerdown';
+      event['originalEvent'] = new PointerEvent('pointerdown');
```
jsalankey, line 128: Modifying these properties with bracket notation seems dangerous. Do these work in compiled mode? 

Would it be better to create a new `MapBrowserEvent` instance to pass to `handleEvent` below? Or do the modifications need to be on the original event reference for whatever else it does as part of the event handler chain?

glennkitchellcaci: Yep; you're spot on about why I did it this way -- OpenLayers didn't seem to trust the `new ol.MapBrowserEvent()` when I tried it that way. Though I did make some tweaks here that made it work better in this version that I didn't try again with the `new` event version. Namely, I stopped passing a clone of the `mapBrowserEvent.originalEvent` to the `new PointerEvent()`'s optional second parameter. That parameter gave the `new PointerEvent()` the same positions and map information as the original event, but for some reason that made the whole outer MapBrowserEvent stop working with the top-level ol.Interaction classes.

jsalankey (approved):

PR #1262, opened 2021-05-06
Feat modify combine

jsalankey:
Combines the menu options for modifying areas and geometries together into a single UI. This cleans up the UX and helps make it clearer what the menu item does. Also adds support for using the modify interaction to update geometries on non-point places, which were previously impossible to change once they were drawn.


src/os/interaction/modifyinteraction.js
```diff
@@ -225,6 +223,34 @@ class Modify extends OLModify {
...
+    const changed = value !== this.getActive();
+
+    if (changed && this.clone_) {
+      if (!value) {
+        const geometry = this.clone_.getGeometry();
+        if (geometry) {
+          olEvents.unlisten(geometry, OLEventType.CHANGE, this.onGeometryChange, this);
+        }
+
+        os.MapContainer.getInstance().removeFeature(this.clone_);
```
schmidtk, line 239: There are a couple global references to `os.MapContainer` here that should use the module import.

src/os/ui/featureedit.js
```diff
@@ -689,6 +689,8 @@ os.ui.FeatureEditCtrl.prototype.disposeInternal = function() {
     this.previewFeature = null;
   }
 
+  goog.dispose(this.interaction);
```
schmidtk, line 692: This should also remove the interaction from the map if defined. OL doesn't have any dispose logic of its own, so we could either remove the interaction here (where the logic to add it exists) or in the dispose function for our override (to ensure it's removed from the map no matter where it's being used).

I would probably lean toward putting it in the dispose function if `this.getMap()` is defined on the interaction so this case is handled everywhere the interaction is used.

src/os/ui/featureedit.js
```diff
@@ -1674,6 +1676,63 @@ os.ui.FeatureEditCtrl.prototype.onAxisChange = function() {
...
+os.ui.FeatureEditCtrl.prototype.modifyGeometry = function() {
+  if (this.interaction) {
+    os.MapContainer.getInstance().getMap().removeInteraction(this.interaction);
+    this.interaction.setActive(false);
+    goog.dispose(this.interaction);
+  }
+
+  if (this.previewFeature) {
+    const mc = os.MapContainer.getInstance();
+    this.interaction = new os.interaction.Modify(this.previewFeature);
```
schmidtk, line 1693: We should define `this.interaction` in the constructor so the compiler is aware of the expected type.

src/os/ui/featureedit.js
```diff
@@ -1674,6 +1676,63 @@ os.ui.FeatureEditCtrl.prototype.onAxisChange = function() {
...
+  os.MapContainer.getInstance().getMap().removeInteraction(this.interaction);
+};
+
+
+/**
+ * Callback handler for canceling a modify.
+ */
+os.ui.FeatureEditCtrl.prototype.onInteractionCancel = function() {
+  if (this.interaction) {
+    os.MapContainer.getInstance().getMap().removeInteraction(this.interaction);
```
schmidtk, line 1728: Do we want to incorporate all three of these lines in the interaction's dispose function?

schmidtk (changes requested):

src/os/interaction/modifyinteraction.js
jsalankey: Unfortunately this file already had some non-modularized references that were necessary due to a non-obvious circular dependency. When I first put it in, the idea was to just use the drawing layer/source directly, but for some reason that didn't play nicely with some of the drawing layer specific-stuff in `MapContainer`.

src/os/ui/featureedit.js
jsalankey: Yeah it's now only used in these 2 places (`modifyarea` and `featureedit` directives) so all of the dispose logic should be able to remove itself from the map.

src/os/interaction/modifyinteraction.js
schmidtk: Ahh, ok we can resolve that separately.

src/os/interaction/modifyinteraction.js
```diff
@@ -225,6 +229,34 @@ class Modify extends OLModify {
...
   }
 
+  /**
+   * @inheritDoc
+   */
+  setActive(value) {
+    const changed = value !== this.getActive();
+
+    if (changed && this.clone_) {
+      if (!value) {
```
schmidtk, line 239: This logic to manage the clone listener and map feature are also in the constructor and dispose function. Do we still want them there?

jsalankey: I don't think we're doing it anywhere currently, but if something wanted to deactivate the interaction without disposing the interaction, I think we'd still want this logic to remove the feature from the map.

schmidtk (approved):

PR #1263, opened 2021-05-07
Feat layers ellipse

arwatts16:
Continued work - https://github.com/arwatts16/opensphere/tree/feat-layers-followon


src/os/im/mapping/ellipsemappingmanager.js
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+
+
+  /**
+   * Execute the Mappings for the source
+   */
+  executeMappings() {
+    const sourceId = this.source_.getId();
+    const layer = os.map.mapContainer.getLayer(sourceId);
```
glennkitchellcaci, line 183: goog.require() and do the getInstance()

src/os/ui/layer/vectorlayerui.js
```diff
@@ -659,6 +713,33 @@ os.ui.layer.VectorLayerUICtrl.prototype.onLineDashChange = function(event, value
...
+    const source = /** @type {os.source.Vector} */ (layer.getSource());
+    const sourceId = source.getId();
+
+    const emm = source.getMappingManager(sourceId) || new os.im.mapping.EllipseMappingManager.MappingManager(source);
+
+    emm.setMappingOptions(value);
+    emm.createMappings(value);
+    emm.executeMappings(layer);
+
+    source.setMappingManager(sourceId, emm);
```
glennkitchellcaci, line 737: It looks like there's only one MappingManager per VectorSource, so I'm not sure you need the by-source-id source.getMappingManager(sourceId) when there's just the one.

src/os/ui/layer/vectorlayerui.js
```diff
@@ -1061,8 +1142,18 @@ os.ui.layer.VectorLayerUICtrl.prototype.getShape = function() {
  * @return {Array<string>} The available shape options
  */
 os.ui.layer.VectorLayerUICtrl.prototype.getShapes = function() {
+  return goog.object.getKeys(os.style.SHAPES);
```
glennkitchellcaci, line 1145: I've seen this line of code a few places. Would it be worth adding an `os.style.getShapes()` or `os.style.getShapesKeys()` function that everyone calls instead?

src/os/source/vectorsource.js
```diff
@@ -534,6 +540,34 @@ os.source.Vector.prototype.changed = function() {
...
+
+
+/**
+ * @param {string} key
+ * @param {*} manager
+ */
+os.source.Vector.prototype.setMappingManager = function(key, manager) {
+  this.mappingManagers[key] = manager;
+};
+
```
glennkitchellcaci, line 569: How much of this lives between refreshes? Should this.mappingManagers has some/part of itself saved during this.persist() and this.restore()?

src/os/source/vectorsource.js
```diff
@@ -416,6 +416,12 @@ os.source.Vector = function(opt_options) {
    */
   this.hasModifications = false;
 
+  /**
+   * Mappings for the source
+   * @type {Object<string, *>}
+   */
+  this.mappingManagers = {};
```
glennkitchellcaci, line 423: Do/will Mappings only work on VectorSource data? Would it be better to move this storage into MappingManagers and make it a singleton?

glennkitchellcaci: Thank you Alyssa for explaining where this is going.

So, as a follow on, please create a `SourceMappingManager` that stores MappingManagers by sourceId > MappingManagerTypeEnum > then the MappingManager, e.g. {'12345': {'ellipseMMKey': new EllipseMappingManager()}} and its own getInstance(). That way, this won't add much of anything to VectorSource, which is exported to be shared among windows.

For now, just update the above to use the MappingManagerTypeEnum (i.e. ELLIPSE_MAPPING_KEY) instead of the SourceId so it'll be a smaller change to move this out later.

src/os/im/mapping/ellipsemappingmanager.js
arwatts16: This one is going to have to stay like that for now. Using `os.MapContainer` results in a circular dependency. I went ahead and added a comment on it though.

src/os/source/vectorsource.js
arwatts16: I think we'd just need to persist `this.MappingOptions_` and the source they're associated with. This would make it easy to recreate the whole mappingManager.

schmidtk: Ideally the manager would implement `IPersistable` and take care of its own persistence internally.

src/os/source/vectorsource.js
schmidtk: I would personally prefer to keep this out of VectorSource in the initial implementation instead of adding it with the assumption it will be removed later. VectorSource is bloated enough as-is, and it doesn't seem this belongs here.

arwatts16: I've added a `SourceMappingManager` that could take care of this. The only other question is if I should create an `AbstractMappingManager` that `EllipseMappingManager` could implement. And also there might be a more ideal way to store mappingmanagers in the new file so let me know if you guys have notes on that.

PR #1265, opened 2021-05-08
refactor(arc): transform arc plugin to goog.module

schmidtk:
Transforms the Arc plugin and tests to use `goog.module`.

### DIFF
--- a/src/os/ui/query/activeentries.js
+++ b/src/os/ui/query/activeentries.js
@@ -1,0 +1,13 @@
+goog.module('os.ui.query.ActiveEntries');
+goog.module.declareLegacyNamespace();
+
+/**
+ * @typedef {{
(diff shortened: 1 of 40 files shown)


schmidtk:
The build was failing due to a type issue with `plugin.arc.ArcFeatureType#persist`. That function didn't quite match the `IPersistable` interface due to not returning anything, but even fixing the function and still using `@inheritDoc` didn't fix the issue. This might be a problem with the compiler and extending interfaces, but it can be resolved by specifying types with `@override`.

schmidtk:
Finished replacing JSDoc-only global references. These are only replaced by `opensphere-jscodeshift` if a `goog.require` statement existed. I may be able to use something like [jsdoc-parse](https://www.npmjs.com/package/jsdoc-parse) to find/fix these in comments only.

jsalankey (approved):

PR #1268, opened 2021-05-10
fix(draw): normalize polygons/lines in current projection

schmidtk:
The line/polygon draw tools were not working in projections other than EPSG:4326 because the geometry was always normalized using that projection, instead of the current map/geometry projection.

We previously normalized the geometry using `x = 0` instead of the default (first geometry coordinate), which was causing rendering issues across the antimeridian. I believe this was intended to resolve issues with servers not accepting the coordinates in the geometry we send, so that solution will need to be reviewed.

[Test Deployment](https://os-fix-polygon-normalization.surge.sh/)

- Drawing lines/polygons across the antimeridian should work in both 2D and 3D, in both EPSG:4326 and EPSG:3857.
- Drawing large polygons (spanning more than half the world) should both display properly and query properly.
- When large/crossing polygons are drawn in 3D, switching to 2D should still render the geometry correctly.
- Polygons crossing the antimeridian should query properly with GeoServer. [This demo instance](https://demo.geo-solutions.it/geoserver/ows) can be used to test.
- While these queries already worked with Arc servers, we should verify they still work.

### DIFF
--- a/src/os/geo/jsts.js
+++ b/src/os/geo/jsts.js
@@ -20,11 +20,13 @@ goog.require('ol.geom.MultiPoint');
 goog.require('ol.geom.MultiPolygon');
 goog.require('ol.geom.Point');
 goog.require('ol.geom.Polygon');
+goog.require('ol.proj');
 goog.require('ol.proj.Projection');
 goog.require('os.fn');
 goog.require('os.geo');
 goog.require('os.geo2');
 goog.require('os.geom.GeometryField');
+goog.require('os.interpolate');
 goog.require('os.map');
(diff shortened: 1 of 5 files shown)


schmidtk:
Converted to a draft because there are still issues when the polygon is drawn in 3D.

schmidtk:
The 3D issue was caused by `getGeometry` transforming the geometry to EPSG:4326 prior to normalization, which was expecting the current map projection. That transform is not necessary and has been removed.

jsalankey (approved):

PR #1269, opened 2021-05-11
feat(add-data): adds support for turning on full providers from add data

jsalankey:
Adds a new checkbox UI for layer providers/folders that allows for turning on full sets of layers. This UI will prompt the user to confirm when turning on more than a predefined number of layers with a single click for safety. This is currently only employed by the Geopackage plugin provider from opensphere-plugin-geopackage, but can be adapted to any provider we like.

### DIFF
--- a/src/os/ui/data/layercheckbox.js
+++ b/src/os/ui/data/layercheckbox.js
@@ -1,0 +1,119 @@
+goog.module('os.ui.data.LayerCheckboxUI');
+
+const ConfirmUI = goog.require('os.ui.window.ConfirmUI');
+const Module = goog.require('os.ui.Module');
+const TriState = goog.require('os.structs.TriState');
+const TriStateCheckboxCtrl = goog.require('os.ui.TriStateCheckboxCtrl');
+const triStateCheckboxDirective = goog.require('os.ui.triStateCheckboxDirective');
+
+const TriStateTreeNode = goog.requireType('os.structs.TriStateTreeNode');
+
+
+/**
+ * Tristate checkbox extension for enabling layers.
+ *
+ * @return {angular.Directive}
+ */
+const directive = () => {
+  const dir = triStateCheckboxDirective();
+  dir.controller = Controller;
+  return dir;
+};
+
+
+/**
+ * The element tag for the directive.
+ * @type {string}
+ */
+const directiveTag = 'layercheckbox';
+
+
+/**
+ * Add the directive to the module
+ */
+Module.directive(directiveTag, [directive]);
+
+
+
+/**
+ * Controller for the layer checkbox.
+ * @unrestricted
+ */
+class Controller extends TriStateCheckboxCtrl {
+  /**
+   * Constructor.
+   * @param {!angular.Scope} $scope The Angular scope.
+   * @param {!angular.JQLite} $element The root DOM element.
+   * @ngInject
+   */
+  constructor($scope, $element) {
+    super($scope, $element);
+
+    /**
+     * The minimum count of layers before prompting the user.
+     * @type {number}
+     */
+    this.minCount = $scope['minCount'] || 10;
+  }
+
+  /**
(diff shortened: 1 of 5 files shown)


src/os/ui/data/layercheckbox.js
```diff
@@ -0,0 +1,119 @@
...
+  toggle(e) {
+    if (this.scope) {
+      const item = /** @type {TriStateTreeNode} */ (this.scope['item']);
+
+      if (!this.getDisabled(item)) {
+        const children = item.getChildren();
+        const state = item.getState();
+
+        if (state == TriState.OFF) {
+          if (children && children.length > this.minCount) {
```
schmidtk, line 71: There may be some specific situations we want to consider here:

- If this node has 5 folders with 5 layers each the warning would not be triggered but 25 layers would be added.
- If this node has 5 folders with 10 layers each, we would get one warning displayed by each of the 5 folders.

We can solve the first with a recursive child count, but the second would have to prevent the warning from also being displayed by the children in their `toggle` calls. One option might be making the `confirm` function here globally disable the warning before calling `toggleInternal` and enable again after the call.

schmidtk (changes requested):

src/os/ui/data/layercheckbox.js
jsalankey: Will fix the recursive count issue.

For folders that might exceed the limit, I don't think the checkbox UI code is what gets called to activate them when the state change bubbles down the tree, it calls the `setState` function on the nodes. That's why I put this in the UI and not in the `setState` call: it should only ever be called by a click, never internal model changes. This also prevents it from being hit on application restore (which I ran into when I had it in the `setState` function initially).

src/os/ui/data/layercheckbox.js
```diff
@@ -68,16 +69,45 @@ class Controller extends TriStateCheckboxCtrl {
         const state = item.getState();
 
         if (state == TriState.OFF) {
-          if (children && children.length > this.minCount) {
+          let count = 0;
+          const countLayerChildren = (child) => {
```
schmidtk, line 73: We might want to make this a public convenience function on this class (could be static) or elsewhere. That would make it directly testable, and would avoid creating a new function instance every time this function is called. You would just need to add `count` as a parameter/return value to avoid the need for a local closure, then swap  `forEach` with `reduce` to aggregate the count.

src/os/ui/data/layercheckbox.js
```diff
@@ -0,0 +1,149 @@
...
+        if (state == TriState.OFF) {
+          let count = 0;
+          const countLayerChildren = (child) => {
+            if (child instanceof DescriptorNode) {
+              const descriptor = child.getDescriptor();
+
+              if (descriptor instanceof LayerSyncDescriptor) {
+                const layers = descriptor.getOptions();
+
+                if (Array.isArray(layers)) {
```
schmidtk, line 80: Is `Array.isArray` needed? `getOptions` should return an array or null, and on null we shouldn't adjust the count.

schmidtk (changes requested):

src/os/ui/data/layercheckbox.js
jsalankey: Yep. I just tend to find local recursive `forEach` more straightforward than `reduce`.

src/os/ui/data/layercheckbox.js
jsalankey: Woops, was originally calling `getLayerOptions()` which is protected, but returns `{Array<Object>|Object}`. Didn't realize the return profile was different here.

src/os/ui/data/layercheckbox.js
schmidtk: I do too, but if a local function closure can be avoided I tend to lean towards that.

schmidtk (approved):

PR #1271, opened 2021-05-13
Feat import xyz layers

ameletyan:


src/plugin/config/configprovider.js
```diff
@@ -114,6 +115,8 @@ plugin.config.Provider.prototype.loadConfig = function(id, config) {
     if (!descriptor) {
       descriptor = new os.data.ConfigDescriptor();
       descriptor.setId(descriptorId);
+      descriptor.setType(config['type']);
```
schmidtk, line 118: Are these two lines needed? `ConfigDescriptor` will get the type and description from the base configs set by the `setBaseConfig` call below. The `getType` and `getDescription` overrides on `ConfigDescriptor` will search the configs for the appropriate field, so you shouldn't need to call these setters.

`ConfigDescriptor#getType` does use the `layerType` field on the config, so maybe that needs to be changed/added to the layer config you're creating?

src/plugin/xyz/xyzproviderimport.js
```diff
@@ -0,0 +1,206 @@
...
+
+  /**
+   * Handle projection change
+   * @export
+   */
+  onProjectionChange() {
+    this.scope['config']['zoomOffset'] = null;
+    switch (this.scope['config']['projection'].code) {
+      case 'EPSG:3857':
+        this.scope['config']['zoomOffset'] = 0;
```
schmidtk, line 191: We might want to make the 3857 case also fall through as the default. We don't configure other projections by default anyway, but I think 0 zoom offset is a good default behavior.

views/plugin/xyz/xyzproviderhelp.html
```diff
@@ -0,0 +1,28 @@
+<div id="time-help" class="flex-fill w-100 d-flex flex-column">
+  <div class="modal-body container flex-fill">
+    Please provide a valid XYZ tile layer URL.
+    <br>
+    <br>
+    An XYZ tile layer URL can have many different formats, such as:
+    <ul>
+      <li>https://osm.gs.mil/tiles/default/{z}/{x}/{y}.png</li>
```
schmidtk, line 8: Example URL's should either be fake (ie, `https://example.com/{z}/{x}/{y}`) or a public facing service like those used in the default opensphere config:
```
https://services.arcgisonline.com/ArcGIS/rest/services/World_Street_Map/MapServer/tile/{z}/{y}/{x}
```

views/plugin/xyz/xyzproviderhelp.html
```diff
@@ -0,0 +1,28 @@
...
+      <li>https://services.arcgisonline.com/ArcGIS/rest/services/<br>World_Street_Map/MapServer/tile/{z}/{y}/{x}</li>
+    </ul>
+    Keep in mind that the positions of {x}, {y}, and {z} can vary between URLs.
+    <br>
+    <br>
+    Although the {x}, {y}, and {z} portions of each URL can be replaced with an integer value, it is not recommended to do so as this will only provide a single tile of the layer.
+    <br>
+    <br>
+    OpenSphere currently supports two projections for XYZ tile layers: EPSG:3857 and EPSG:4326. If an XYZ tile layer does not appear as expected, consider reimporting it with a different projection.
```
schmidtk, line 20: Avoid hard coding the application name, because other applications using OpenSphere may change this in their config. To get the actual app name, assign `os.config.getAppName` to a property on the controller or scope and reference that in the template.

src/plugin/xyz/xyzproviderimportform.js
```diff
@@ -0,0 +1,37 @@
+goog.module('plugin.xyz.XYZImportForm');
+goog.module.declareLegacyNamespace();
+
+const Module = goog.require('os.ui.Module');
+const {directive: xyzImportDirective} = goog.require('plugin.xyz.XYZImport');
+const {ROOT} = goog.require('os');
+
+
+/**
+ * A dialog with information about custom date/time formats, as implemented by Moment.js.
```
schmidtk, line 10: Please update the comment to reflect what this directive is used for.

schmidtk (changes requested):

src/plugin/xyz/xyzproviderimport.js
```diff
@@ -0,0 +1,206 @@
...
+
+    $scope['typeName'] = 'XYZ Provider';
+    $scope['urlExample'] = 'https://example.com/{z}/{x}/{y}';
+    $scope['config']['id'] = 'xyz';
+    $scope['config']['label'] = this.getLabel() || '';
+    $scope['config']['title'] = this.scope['config']['label'];
+    $scope['config']['type'] = 'XYZ';
+    $scope['config']['layerType'] = 'XYZ';
+    $scope['config']['url'] = '';
+    $scope['config']['minZoom'] = 1;
```
schmidtk, line 84: I would suggest using `os.map.MIN_ZOOM` and `os.map.MAX_ZOOM` for the initial values here.

schmidtk (changes requested):

views/plugin/xyz/xyzdescriptornodeui.html
```diff
@@ -0,0 +1,8 @@
+<div>
+  <span ng-if="nodeUi.show()" class="flex-shrink-0" ng-click="nodeUi.edit()">
+    <i class="fa fa-pencil-square-o fa-fw c-glyph" title="Edit this layer"></i>
```
schmidtk, line 3: We typically use [this icon](https://fontawesome.com/v5.15/icons/pencil-alt?style=solid) for edit. In FA4 the classes are `fa fa-pencil`, now that we're on FA5 we should start using `fas fa-pencil-alt`.

schmidtk (changes requested):
Functionally everything looks good, I had a few minor cleanup comments.

src/plugin/xyz/xyzdescriptornodeui.js
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Generic node UI for XYZ descriptors.
+ *
+ * @return {angular.Directive}
+ */
+const directive = () => {
+  return {
+    restrict: 'AE',
+    replace: true,
+    templateUrl: ROOT + 'views/plugin/xyz/xyzdescriptornodeui.html',
```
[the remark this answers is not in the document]
ameletyan: I changed it to an inline template and got rid of `xyzdescriptornodeui.html`.

schmidtk (approved):

PR #1272, opened 2021-05-14
feat(wfs): support changing the content type header in wfs requests

schmidtk:
Using `Content-Type: text/xml` in a WFS POST request requires sending a CORS preflight request. We encountered a GeoServer that does not handle this properly with its authentication approach. Until the remote service fixes this issue, our temporary workaround is to change the content type to `text/plain`, which constitutes a simple request and doesn't require a preflight.

This does not change any default behavior, but allows OGC servers and WFS layers to specify the value of the `Content-Type` header sent in a WFS POST request.

Depends on ngageoint/opensphere-state-schema/pull/23 being merged for tests to pass.


jsalankey (approved):

schmidtk:
@jsalankey I updated the `yarn.lock` to use the new version of `opensphere-state-schema`, so test errors are now resolved.

jsalankey (approved):

PR #1274, opened 2021-05-18
feat: added warning when /web is in a GeoServer URL

ameletyan:


src/os/ui/providerimport.js
```diff
@@ -291,3 +291,11 @@ os.ui.ProviderImportCtrl.prototype.formDiff = function() {};
...
  */
 os.ui.ProviderImportCtrl.prototype.getConfig = function() {};
+
+
+/**
+ * Handles URL changes
+ *
+ * @export
+ */
+os.ui.ProviderImportCtrl.prototype.validateUrl = function() {};
```
schmidtk, line 301: Sorry I suggested the wrong place for this. I think it makes more sense on the `SingleUrlProviderImportCtrl` class. Could you move it there please?

src/plugin/ogc/ui/geoserverimport.js
```diff
@@ -75,3 +77,17 @@ plugin.ogc.ui.GeoserverImportCtrl.prototype.getDataProvider = function() {
...
+
+
+/**
+ * @inheritDoc
+ * @export
+ */
+plugin.ogc.ui.GeoserverImportCtrl.prototype.validateUrl = function() {
+  if (/\/web\/?$/.test(this.scope['config']['url'])) {
+    this.scope['customUrlMessage'] = 'GeoServer URLs ending with \"/web\" are typically for the administration ' +
+    'interface. Consider replacing \"/web\" with \"/wfs,\" \"/wms,\" or \"/wmts\" for the OGC service APIs.';
```
schmidtk, line 89: OpenSphere only supports the `/owc` endpoint for the GeoServer import path so we shouldn't advertise these here. You can use them if you are adding the server via Add Server -> OGC Server, but here they will still produce an error.

schmidtk (changes requested):

schmidtk (approved):

PR #1275, opened 2021-05-19
Feat layers followon

arwatts16:
Related to - https://github.com/ngageoint/opensphere/pull/1263


views/layer/ellipsecolumns.html
```diff
@@ -0,0 +1,47 @@
+<div>
+
+  <div>
+    Applying these mappings will cause the layer to reload and you could lose its current state. <b>This action cannot be undone.</b>
+  </div>
+
+  <div>
+    <div class="form-inline mb-1">
+      <input class="form-check-input" type="radio" ng-value="0" ng-model="ctrl.inputType" id="circle">
+      <label class="form-check-label" for="circle">Circle</label>
```
schmidtk, line 10: It might be nice to put these radios together at the top of the form. When you change the selection, the Ellipse radio moves because of the changes to the rest of the form. As a user, this is a bit jarring and would feel better if the radios were stationary and only the form below changed.

src/os/ui/layer/vectorstylecontrols.js
```diff
@@ -42,6 +43,10 @@ os.ui.layer.vectorStyleControlsDirective = function() {
       'size': '=',
       'shape': '=',
       'shapes': '=',
+      'columns': '=?',
+      'allowEllipseConfig': '=?',
+      'ellipseMappings': '=?',
```
schmidtk, line 48: Is this still used? I don't see it being set or referenced anywhere.

src/os/ui/layer/vectorstylecontrols.js
```diff
@@ -112,6 +117,22 @@ os.ui.layer.VectorStyleControlsCtrl = function($scope, $element) {
...
+  this['ellipseMapping'] = undefined;
+
+  this['allowEllipseConfig'] = this.scope['allowEllipseConfig'] || false;
+
+  /**
+   * Help text for Ellipse configuration
+   * @type {string}
+   */
+  this['configEllipse'] = `This layer does not have columns that support Ellipse styling (Radius, or Semi Minor, Semi 
+    Major, and Orientation). To configure which columns should be mapped be used instead, use the settings cog to
```
schmidtk, line 133: Wording is a little off here, and perhaps more verbose than needed. Maybe simplify to something like:
> Required ellipse columns are not configured on this layer. To display ellipses, click the settings cog and select the requested columns.

src/os/data/filedescriptor.js
```diff
@@ -173,25 +176,37 @@ os.data.FileDescriptor.prototype.getLayerOptions = function() {
...
 os.data.FileDescriptor.prototype.setMappings = function(value) {
   this.parserConfig['mappings'] = value;
 };
 
 
+/**
+ * @inheritDoc
+ */
+os.data.FileDescriptor.prototype.update = function() {
+  this.reimport();
```
schmidtk, line 198: Can we reload the data without going through the whole reimport process? It felt confusing that I would need to go through the whole wizard again, and I imagine users will feel the same.

src/os/ui/layer/ellipsecolumns.js
```diff
@@ -0,0 +1,356 @@
...
+    yesButtonClass: 'btn-danger',
+    windowOptions: {
+      'label': 'Map Ellipse Columns',
+      'x': 'center',
+      'y': 'center',
+      'width': '330',
+      'height': 'auto',
+      'modal': 'true',
+      'show-close': 'true',
+      'headerClass': 'bg-danger u-bg-danger-text'
```
schmidtk, line 298: The red header makes it feel like I'm doing something inherently dangerous/risky, which I don't think is appropriate for this window. I would personally vote for using the standard header/button for this UI.

schmidtk (changes requested):

src/os/ui/layer/vectorlayerui.js
```diff
@@ -146,6 +146,12 @@ os.ui.layer.VectorLayerUICtrl = function($scope, $element, $timeout) {
    */
   this['uniqueId'] = null;
 
+  /**
+   * Feature Toggle
+   * @type {boolean}
+   */
+  this['allowEllipseConfig'] = os.settings.get(os.ui.layer.EllipseColumnsUI.ALLOW_ELLIPSE_CONFIG, false);
```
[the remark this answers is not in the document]
arwatts16: I included this to make the feature toggle-able so that anyone who doesn't want it can disable it in their configs.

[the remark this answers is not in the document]
schmidtk: Ok, we probably want to enable it by default. It does seem like a useful addition, so I would vote making it opt-out instead of opt-in.

src/os/ui/layer/vectorstylecontrols.js
```diff
@@ -174,6 +195,21 @@ os.ui.layer.VectorStyleControlsCtrl.prototype.hasCenter = function() {
...
+ *
+ * @return {boolean}
+ * @export
+ */
+os.ui.layer.VectorStyleControlsCtrl.prototype.hasEllipseCols = function() {
+  const layerNodes = this.scope['layerNodes'];
+  const layer = layerNodes ? layerNodes[0].getLayer() : undefined;
+  const source = layer ? layer.getSource() : undefined;
+
+  return source ? source.supportsShape(os.style.ShapeType.ELLIPSE) : false;
```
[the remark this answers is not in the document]
arwatts16: `supportsShape` is only checking if the expected ellipse columns exist. Meaning that if the user wanted it to be an ellipse but those columns didn't exist, then they could never add them.

src/os/ui/layer/vectorlayerui.js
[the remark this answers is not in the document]
arwatts16: I was planning to leave this as opt-in because then we don't have to worry about if we missed a config where the change should definitely not be, at least until the feature is done

views/plugin/kml/kmlnodelayerui.html
```diff
@@ -19,6 +19,8 @@
           line-dash="lineDash"
           shape="shape"
           shapes="shapes"
+          allow-ellipse-config="false"
+          layer-nodes="vector.layerNodes"
```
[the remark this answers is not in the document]
arwatts16: I originally was adding layerNodes because `vectorstylecontrols.js` already expects to have `layerNodes` as part of the scope. I think I can remove that entirely now though because above `layer-nodes` is `allow-ellipse-config="false"` so this ui is not supported/able to be used for kml node layers anyway.

views/wiz/geometrystep.html
```diff
@@ -133,53 +133,7 @@
...
-          <label class="col-form-label col-2">Orientation</label>
-          <select class="custom-select col-4 mr-1" name="orientation" ng-model="step.ellipse.orientation.column" ng-change="geomStep.validate()"
-              ng-required="step.showEllipse && (step.ellipse.semiMajor.column || step.ellipse.semiMinor.column)"
-              ng-options="column.name as column.name for column in geomStep.columns">
-            <option value="">-- None --</option>
-          </select>
-          <small class="form-text invalid-feedback d-inline-block col-2" ng-if="ellipseForm.orientation.$error.required">Required!</small>
-        </div>
-      </div>
+      <ellipsecolumns layer="config"></ellipsecolumns>
```
schmidtk, line 136: The `<div class="card-body">` is still needed here to provide some padding around the controls inside the card.

schmidtk (changes requested):
In addition to the inline feedback, you can't currently switch from Ellipse to Circle on a file layer. I tried this for CSV and GeoJSON, and it continues to map ellipse columns and display an ellipse after switching to Circle in the config UI. It looks like a `RadiusMapping` is added, but the ellipse column mappings are still present.

src/os/ui/layer/vectorstylecontrols.js
```diff
@@ -219,3 +255,15 @@ os.ui.layer.VectorStyleControlsCtrl.prototype.select2Formatter_ = function(item,
...
 };
 
+
+/**
+ * Launches the window to configure ellipse columns
+ *
+ * @export
+ */
+os.ui.layer.VectorStyleControlsCtrl.prototype.launchConfigureWindow = function() {
+  const layer = this.scope['layerNodes'] ? this.scope['layerNodes'][0].getLayer() : undefined;
```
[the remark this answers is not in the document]
schmidtk: The control to launch this UI is still visible when multiple vector layers are selected. Given the control will only act on one of the layers, we should hide it unless there is only a single layer selected.

arwatts16:
> In addition to the inline feedback, you can't currently switch from Ellipse to Circle on a file layer. I tried this for CSV and GeoJSON, and it continues to map ellipse columns and display an ellipse after switching to Circle in the config UI. It looks like a `RadiusMapping` is added, but the ellipse column mappings are still present.

Yeah, so the issue I have with this is that in `ellipsecolumns.js updateMappings()` I'm combining existing mappings with ones that I've created. So instead of saying "oh hey, now that we're mapping radius lets stop mapping semiMajor/semiMinor/orientation" I'm just replacing the mapping if it already exists. Is there an instance where someone would want both types of mapping (circle and ellipse) or is it probably fair to remove ellipse if you're using circle, and vice versa?

schmidtk:
@arwatts16 the mappings should be mutually exclusive because they are both rendered by the same Ellipse style selection. If the data has ellipse fields there shouldn't be a reason to switch to Circle, but if you incorrectly map ellipse fields and want to switch to Circle you're out of luck because the ellipse mappings aren't removed.

arwatts16:
@schmidtk Alright, yeah good point. I've updated the `updateMappings()` function in `ellipsecolumns.js` to reflect that

src/os/data/layersyncdescriptor.js
```diff
@@ -109,6 +120,45 @@ os.data.LayerSyncDescriptor.prototype.disposeInternal = function() {
...
+
+/**
+ * @inheritDoc
+ */
+os.data.LayerSyncDescriptor.prototype.updateMappings = function(layer) {
+  const dm = os.data.DataManager.getInstance();
+  dm.updateDescriptor(this, this);
+  dm.persistDescriptors();
+
+  // Delete the layer, then prompt the descriptor to make new layers
```
schmidtk, line 149: It's a little odd that this implementation only removes the layer, and leaves it to extending classes to add the layer back. This should be a complete implementation on its own. Can we recreate the layer here? It would be nice if descriptors like OGC and Arc didn't need to add their own mapping logic and just got it for free from `LayerSyncDescriptor`.

src/os/data/layersyncdescriptor.js
```diff
@@ -109,6 +120,45 @@ os.data.LayerSyncDescriptor.prototype.disposeInternal = function() {
...
+  return this.mappings;
+};
+
+
+/**
+ * @inheritDoc
+ */
+os.data.LayerSyncDescriptor.prototype.setMappings = function(value) {
+  this.mappings = value;
+  const dm = os.data.DataManager.getInstance();
```
schmidtk, line 136: Can you use `this.saveDescriptor()` here instead?

src/os/data/filedescriptor.js
```diff
@@ -173,25 +177,41 @@ os.data.FileDescriptor.prototype.getLayerOptions = function() {
...
 
 
 /**
- * Set the column mappings to apply to imported data.
- *
- * @param {Array.<os.im.mapping.IMapping>} value
+ * @inheritDoc
  */
 os.data.FileDescriptor.prototype.setMappings = function(value) {
   this.parserConfig['mappings'] = value;
```
schmidtk, line 191: `setMappings` was previously only called when the descriptor was created. Now that we're also calling it after creation, we need to make sure the descriptor gets saved to settings. Currently if you change ellipse mappings on an active file layer and refresh, your changes are lost.

schmidtk (changes requested):

src/os/data/filedescriptor.js
arwatts16: So that's what I have `updateMappings()` for, actually saving the descriptor for refresh (not sure if that's the same as saving to settings). I wanted to keep this function as simple as possible since it's also used in `geometrystep.js` for import and then call `updateMappings()` anytime someone adds a new mapping after import.

schmidtk: Ok, that makes sense. We should make sure the descriptor gets saved in `updateMappings` then.

schmidtk (approved):

PR #1277, opened 2021-05-21
Perform JSTS validation on individual components of MultiPolygon

jkae:
This resolves an issue where overlapping polygons are merged together, losing outlines and intended styling.

Also resolves an issue where a single overlapping edge will invalidate an entire MultiPolygon rather than just the overlap.

These two things are generally intended behavior, but prevent loading of KML that Opensphere itself is able to parse & render.


schmidtk:
A MultiPolygon is invalid if its polygons overlap, even on an edge. I don't think we should change this in our validation because sending this polygon to a server may cause an error. It seems the core issue is that the KML specification uses `MultiGeometry` to represent a MultiPolygon, but that does not have the same restrictions. A MultiGeometry may have different geometry types, and those geometry types may overlap one another.

It seems the problem here is that the OpenLayers KML parser (specifically `ol.format.KML.readMultiGeometry_`) uses a Multi* geometry type if all geometries in the collection are of the same type. This doesn't seem like the correct behavior, given the difference in geometry validation between a GeometryCollection and a MultiPolygon.

@wallw-bits what are your thoughts on this?

wallw-bits:
I do not think the renderer should be restricted on that rule. If servers do not accept those, then that should be checked before being sent to the server.

I was under the impression that KML switched to a `MultiX` type from `MultiGeometry` for performance reasons, but you could certainly try that out.

schmidtk:
Ah that's right. @jkae we should verify Arc and OGC servers (GeoServer) support MultiPolygons validated with this approach, or if we need to do any further validation of those before using them in a query.

jkae:
When querying using an imported KML polygon it gets saved to areas, so it's re-validated as a full MultiPolygon. It seemed that coincident polygons could be re-polygonized into a valid query, but ones with overlapping edges were unable to do that: it would fail and return the original MultiPolygon geometry, which threw an error in the OGC and ArcGIS servers I tested.

My recent commit updates the JSTS validation to unionize any 100% invalid MultiPolygon when saving to Areas, which I hope is an acceptable solution.

schmidtk (approved):
I created a KML to test various `MultiGeometry` instances. It has geometries with a single intersecting vertex, an intersecting edge, partial overlap, full overlap, and no overlap. All cases rendered properly as KML, and loaded as Areas with Arc and GeoServer.

This is my test file, rename to `.kml` to load in OpenSphere:
[MultiPolygon_Overlap.txt](https://github.com/ngageoint/opensphere/files/6649130/MultiPolygon_Overlap.txt)

src/os/feature/feature.js
```diff
@@ -1380,6 +1381,43 @@ const validateGeometries = function(feature, opt_quiet) {
...
+ *
+ * MultiPolygon geometries have their internal polygons individually validated, and removed if they fail the validation.
+ * Otherwise the structure of the MultiPolygon is not affected.
+ *
+ * @param {Geometry} geometry The polygon-type geometry to validate
+ * @param {boolean=} opt_quiet If alerts should be suppressed
+ * @return {Geometry|undefined} a valid polygon, or undefined if invalid
+ * @private
+ */
+const validatePolygonType_ = function(geometry, opt_quiet) {
```
schmidtk, line 1395: Module-local symbols should not use `@private` (this is assumed unless the symbol is exported), or a trailing underscore.

https://google.github.io/styleguide/jsguide.html#file-goog-module-exports

src/os/feature/feature.js
```diff
@@ -29,6 +29,7 @@ const DynamicFeature = goog.require('os.feature.DynamicFeature');
 const fields = goog.require('os.fields');
 const {reduceExtentFromGeometries} = goog.require('os.fn');
 const geo = goog.require('os.geo');
+const {validate: jstsValidate} = goog.require('os.geo.jsts');
```
schmidtk, line 32: To encourage consistent naming, named exports should not be renamed on import. Either use the original name, or import the full module exports using a name based on the import path (ie, `osGeoJsts`).

https://google.github.io/styleguide/jsguide.html#es-module-imports

schmidtk (changes requested):
I have a few minor requests based on the JS style guide.

schmidtk (approved):

PR #1278, opened 2021-05-21
fix: trash icon now appears consistently in the Add Data window

ameletyan:
`views/data/descriptornodeui.html` was deleted because it was no longer referenced and the HTML within was converted into a string in the last place that it was referenced


schmidtk (approved):
Nice catch, looks good!

PR #1290, opened 2021-06-01
Transform feature action stack to goog.module

schmidtk:
Transforms all import/feature action code to use `goog.module`. This includes the `os.im.action`, `os.ui.im.action`, and `plugin.im.action` namespaces.

### Change Notes

- The feature action plugin was requiring action config UI's as an indirect way of loading them with Angular. I reworked this to make the dependencies more concrete. The plugin imports the actions and adds them to the managers, and the actions import their respective config UI to reference the directive's tag name.
- There were a number of convenience functions defined on a different namespace than their file was providing. These were moved to the correct module. This includes `os.im.action`, `os.ui.im.action`, and `plugin.im.action.feature.ui`.
- Similar to the above, `plugin.im.action.feature.ui.launchForLayer` was created to isolate that function. This could not be included in the `plugin.im.action.feature.ui` module because a circular dependency would have been introduced.
- `FeatureActionsCtrl` had a "no this before super" situation to resolve. The parent constructor (`FilterActionsCtrl`) was calling an init function, so the child had to set some things up prior to calling it. To resolve this, I added the `$onInit` lifecycle function to `FilterActionsCtrl` which does the initialization after all constructors have finished executing.
- Resolving implicit dependencies caused a circular dependency with `ImportActionManager`. I used a similar solution as `PlacesManager` to resolve this, adding functions to get/set the global import action manager in the `os.im.action` module. This was a bit more complicated than places because `ImportActionManager` is extended by the feature action manager, which replaces the global instance when the feature action plugin is loaded. It might be beneficial to replace all `ImportActionManager.getInstance()` calls with `getImportActionManager()`, but I only replaced instances that were causing circular dependencies.
- The compiler was not recognizing the `IImportAction` function signatures on `AbstractImportAction`. This is caused by google/closure-compiler/issues/3583, and the workaround is to `goog.require` all interfaces instead of using `goog.requireType`. The hard dependency causes the implemented/extended interface to be loaded first by the compiler which resolves the type confusion. This requires suppressing eslint errors, because those requires are only referenced in JSDoc.


src/os/im/action/abstractimportaction.js
```diff
@@ -3,7 +3,9 @@ goog.module.declareLegacyNamespace();
 
 const osXml = goog.require('os.xml');
 
-const IImportAction = goog.requireType('os.im.action.IImportAction');
+// The compiler has to process this first or @inheritDoc will not work properly on implementing classes.
+// @see https://github.com/google/closure-compiler/issues/3583
+const IImportAction = goog.require('os.im.action.IImportAction'); // eslint-disable-line opensphere/no-unused-vars
```
jsalankey, line 8: Is importing this and assigning the return value with a lint disable a better solution than just a naked require? I know we can't use naked requires with ES6 modules, but we aren't on those yet and disabling the linter is kinda ugly.

src/os/im/action/filteractionentry.js
```diff
@@ -14,6 +14,7 @@ const IImportAction = goog.requireType('os.im.action.IImportAction');
  * Filter entry that performs actions on matched data.
  *
  * @implements {IComparable<FilterActionEntry>}
+ * @unrestricted
```
jsalankey, line 17: We typically add this for bracket access, is that being used on this class somewhere?

src/os/im/action/abstractimportaction.js
schmidtk: In this particular case, you need an assignment for the `const IImportAction` referenced in the `@implements` tag so you can't just use a bare `require`.

For the broader issue like requires for directives only referenced in the template, I've gone back and forth on what to do. You can still call `goog.require` in an ES module, but Closure will warn you if you're doing that with another ES module. We will have to use an `import` statement with an `eslint-disable-line` comment at that point, so I don't think it would hurt to start doing that now. I'm fine with either approach though for goog modules.

src/os/ui/im/action/filteractionimport.js
```diff
@@ -1,165 +1,179 @@
...
-goog.require('os.im.action.cmd.FilterActionAdd');
-goog.require('os.layer.Drawing');
-goog.require('os.ui.Module');
-goog.require('os.ui.filter');
-goog.require('os.ui.filter.im.filterImportDirective');
-goog.require('os.ui.im.action.FilterActionImporter');
+goog.module('os.ui.im.action.FilterActionImport');
+goog.module.declareLegacyNamespace();
+
+const mapContainer = goog.require('os.MapContainer');
```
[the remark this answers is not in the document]
schmidtk: Thanks, replaced this and a reference in `ImportActionManager` with `os.map.instance`.

schmidtk:
Resolved the merge conflict from the command stack refactor.

jsalankey (approved):

PR #1294, opened 2021-06-08
feat(auth): adds basic api/hooks for servers with special auth support

jsalankey:
Adds an API for registering servers that support additional authentication methodologies besides PKI/basic auth. It supports registration of a pattern to match servers as well as ways to link out and message to the user that they are connecting to a server that may provide more data than it appears on the surface.

### DIFF
--- a/src/os/mixin/urltilemixin.js
+++ b/src/os/mixin/urltilemixin.js
@@ -288,14 +288,6 @@ ol.source.UrlTile.prototype.loadingDelay_ = null;
 
 
 /**
- * Flag for whether this source has alerted an error due to missing authentication.
- * @type {boolean}
- * @private
- */
-ol.source.UrlTile.prototype.authAlerted_ = false;
-
-
-/**
  * @inheritDoc
  */
 ol.source.UrlTile.prototype.disposeInternal = function() {
@@ -425,12 +417,11 @@ ol.source.UrlTile.prototype.onImageLoadOrError_ = function(evt) {
   } else {
     this.errorCount_++;
 
-    if (!this.authAlerted_) {
-      const urls = this.getUrls();
-      urls.forEach((url) => {
-        os.auth.alertAuth(url);
-      });
-    }
+    // request failed, check if it's potentially due to a missing authentication with the server
+    const urls = this.getUrls();
+    urls.forEach((url) => {
+      os.auth.alertAuth(url);
+    });
   }
 
   this.decrementLoading();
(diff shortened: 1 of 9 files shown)


src/os/mixin/urltilemixin.js
```diff
@@ -414,6 +424,13 @@ ol.source.UrlTile.prototype.onImageLoadOrError_ = function(evt) {
     this.loadCount_++;
   } else {
     this.errorCount_++;
+
+    if (!this.authAlerted_) {
```
schmidtk, line 428: It looks like `alertAuth` handles deduplication of alerts, but should `authAlerted_` also be set to true here to avoid trying to alert again?

src/os/ui/data/baseprovider.js
```diff
@@ -207,3 +220,33 @@ os.ui.data.BaseProvider.prototype.getUniqueId = function() {
...
+
+/**
+ * @inheritDoc
+ */
+os.ui.data.BaseProvider.prototype.formatIcons = function() {
+  let icons = '';
+
+  const auth = this.getAuth();
+  if (auth) {
+    icons = `<i class="fa fa-sign-in" title="${auth.tooltip}"></i>`;
```
schmidtk, line 243: I'm not certain we'll ever refactor all the old FA4 classes so we can stop using the shim, but ideally new code should use [FA5 classes](https://fontawesome.com/v5.15/icons/sign-in-alt?style=solid).

src/os/ui/server/abstractloadingserver.js
```diff
@@ -124,11 +125,13 @@ os.ui.server.AbstractLoadingServer.prototype.finish = function() {
  */
 os.ui.server.AbstractLoadingServer.prototype.formatIcons = function() {
   var icons = os.ui.server.AbstractLoadingServer.base(this, 'formatIcons');
+
   if (this.getError()) {
     var message = 'Server failed to load. See the log/alerts window for details.';
-    icons += '<i class="fa fa-warning text-warning" title="' + message + '"></i>';
+    icons += `<i class="fa fa-warning text-warning" title="${message}"></i>`;
```
schmidtk, line 131: While we're updating this line, here's the [FA5 icon](https://fontawesome.com/v5.15/icons/exclamation-triangle?style=solid).

schmidtk (changes requested):
I had a few minor comments, otherwise looks good.

src/os/mixin/urltilemixin.js
jsalankey: Yep, meant to take this out when I moved that to the auth module.

schmidtk (approved):

schmidtk (approved):

PR #1299, opened 2021-06-14
Support user settings in Electron app

schmidtk:
This adds a UI for loading/managing user config files in the Electron app. Support for this on the Electron side is implemented in [this opensphere-electron PR](https://github.com/ngageoint/opensphere-electron/pull/39).

- Settings management UI can be opened via Support > Customize Settings. This should only appear in the Electron app.
- Settings files my be imported via the Import File button in the UI, or drag/drop.
- Settings may be loaded by URL, and the app will download the file each time it's loaded. A cache defeater is appended to the URL to ensure changes are picked up on each load.
- If a duplicate file name is detected, you should be prompted to either replace the existing file or save a new one (or close the window to do nothing).
- The default application file should be displayed as "Default" with an icon indicating it's a default settings file. This file can be disabled but not edited or removed.
- User-loaded files can be enabled/disabled, edited (change the label), or removed.
- Settings files are loaded top-down in the order displayed in the UI, and drag/drop can be used to reorder them.
- When the load order or enabled state is changed, this should be indicated in the UI and you should be able to click "Restart & Apply" to relaunch the application and reload settings.

### Other Changes

- We had custom SCSS for `i.fa` and similar selectors that were not working with new Font Awesome 5 icons using `far`, `fas`, etc. These styles have been updated to include the new FA classes.
- Added `AbstractSettingsInitializer#setFileUri` to allow changing the path to the base settings file. Electron uses this to load the managed overrides file.
- `SettingsFile#onGetSuccess_` would error if the content could not be parsed as JSON. This error is now caught, and `onGetFail_` called to handle the error appropriately.
- The `confirmtext` directive has been transformed to a module so the `launchConfirmText` can be loaded appropriately as a module import.
- The `urldragdrop` directive can now optionally be allowed when a modal backdrop is present. This is opt-in, and the default behavior remains the same.
- The `ImportManager` can now be assigned a default import UI. `ImportProcess#importFile` will use the default UI if one isn't registered for the detected file type. If a default UI isn't set, `ImportProcess` will fall back on `AnyTypeImportUI` as it did before.


scripts/electron/preload.js
```diff
@@ -15,22 +17,81 @@ let certHandler;
...
+  COOKIE_UPDATE: 'update-cookies',
+
+  SETTINGS_ADD: 'add-settings',
+  SETTINGS_GET_FILES: 'get-settings-files',
+  SETTINGS_GET_BASE_FILE: 'get-base-settings-file',
+  SETTINGS_GET_USER_DIR: 'get-user-settings-dir',
+  SETTINGS_REMOVE: 'remove-settings',
+  SETTINGS_SET: 'set-settings',
+  SETTINGS_SUPPORTED: 'user-settings-supported',
+  SETTINGS_UPDATE: 'update-settings'
```
jsalankey, line 77: Are these event types something that can be exported from the Electron context/otherwise shared rather than redeclaring identical strings?

src/plugin/electron/customizesettingsui.js
```diff
@@ -0,0 +1,230 @@
...
+ * The element tag for the directive.
+ * @type {string}
+ */
+export const directiveTag = 'customizesettings';
+
+
+/**
+ * Add the directive to the module.
+ */
+Module.directive('customizesettings', [directive]);
```
jsalankey, line 53: Nit: can we use the `directiveTag` variable above here?

src/plugin/electron/customizesettingswindow.js
```diff
@@ -0,0 +1,128 @@
...
+ * The element tag for the directive.
+ * @type {string}
+ */
+export const directiveTag = 'customizesettingswindow';
+
+
+/**
+ * Add the directive to the module.
+ */
+Module.directive('customizesettingswindow', [directive]);
```
jsalankey, line 35: Same question on the `directiveTag`

src/plugin/electron/settingsimportui.js
```diff
@@ -0,0 +1,212 @@
...
+    const content = /** @type {string} */ (file.getContent());
+    const url = file.getUrl();
+
+    // Internal settings files are prefixed with a '.', so remove those from the filename if present to avoid conflicts.
+    const fileName = file.getFileName().replace(/^\.+/, '');
+
+    if (content && url) {
+      try {
+        const parsed = JSON.parse(content);
+        if (parsed && jsonMime.detect(null, file, parsed)) {
```
jsalankey, line 153: Is there a reason we need to check the file type here again? Shouldn't that be done by the `ImportManager` stack before ever launching this UI?

views/plugin/electron/customizesettingswindow.html
```diff
@@ -0,0 +1,28 @@
+<div class="flex-fill w-100 d-flex flex-column">
+  <div class="d-flex modal-body">
+    <customizesettings changed="ctrl.changed"></customizesettings>
```
jsalankey, line 3: Can we combine the window with the "form" directive here rather than having two separate directives? I don't think we'll ever want to put `customizesettings` anywhere not in the window.

src/plugin/electron/electronplugin.js
```diff
@@ -93,6 +103,22 @@ if (isElectron()) {
...
   ElectronOS.updateCookies();
+
+  //
+  // Load Electron's settings file (see appsettings.js in opensphere-electron).
+  //
+  // This is currently only supported in the main window.
+  //
+  if (ElectronOS.supportsUserSettings()) {
+    // Enable mixed content so http/https settings files may be loaded.
+    ExtDomainHandler.MIXED_CONTENT_ENABLED = true;
```
jsalankey, line 114: Is this a safe thing to set globally within Electron? I don't know how general mixed content behaves there vs. in the browser.

schmidtk: Yes, and we do this in the default Electron-specific settings file. The problem I ran into was this needs to be set before settings are loaded in order to load settings files from an http/https URL, because the app is loaded from a `file://` URL.

views/plugin/electron/customizesettingswindow.html
schmidtk: I can, this is an artifact of initially having this UI in the Settings window as a settings plugin then moving it to a window later. I can combine them again, either way is fine by me.

src/plugin/electron/settingsimportui.js
schmidtk: I want to say there was a path where the content wasn't checked yet so I was verifying it appeared to be a JSON settings file. I'll have to take a look.

scripts/electron/preload.js
schmidtk: Yeah that was bothering me and I never fixed it. Moved the cookie/settings event types that had duplicated enums to their own modules, that are now used by the app and the preload script.

src/plugin/electron/settingsimportui.js
schmidtk: Ah right, this was for drag/drop which doesn't go through the `ImportManager`. The drop is handled by the UI directly via the `url-drag-drop` directive.

jsalankey: Gotcha, yeah the drag drop directive design is a bit old and jank, might be something to look at in the future.

jsalankey (approved):

PR #1302, opened 2021-06-16
Transform time/timeline to goog modules

schmidtk:
This transforms `os.time` and all timeline UI's to use `goog.module`.

### Other Changes

- `os.time.UNBOUNDED` was moved to `TimeRange` to resolve a circular dependency.
- The `VariableReplacer` initialization for `now` and `time` was moved to `os.time.replacers#init`. This function must be called to add the replacers (previously they were added on file load), but this is called in `AbstractMainCtrl` which should be extended pretty universally. For use in tests, it was added to the global `beforeEach` in `os.mock.js`. This was all done to resolve some circular deps.
- The `timeOffset` and `timeOffsetLabel` globals on `os.time` do not work well with modules. We could avoid a breaking change with ES modules, but this would have to be converted after all dependent files. To resolve this issue, I added getters and replaced all references. The `timeOffset` assignment in tests was a workaround to an invalid format for `getStart/End` spies (they should return a number, not a string), so I fixed that.
- There were a number of compiler errors due to `ITimelineItem` missing functions that were being called on objects typed with that interface. I added these functions as well as basic implementations in `BaseItem`.
- Fixed some typing issues with `os.ui.timeline` UI's that were previously missed by the compiler.
- The timeline UI `AXIS_HEIGHT` and related properties were causing a circular dependency. I moved those to the `os.ui.timeline` module and created getters/setters to allow overriding them.


src/os/ui/abstractmainctrl.js
```diff
@@ -259,6 +260,9 @@ os.ui.AbstractMainCtrl.prototype.initialize = function() {
   // set up trusted URI's
   os.net.loadTrustedUris();
 
+  // initialize variable replacers for time values in URI's
+  os.time.replacers.init();
```
jsalankey, line 264: Hopefully any extending applications are still calling this `initialize` function, but in case they fully override it, this could cause some annoying breaks for them.

test/os/os.mock.js
```diff
@@ -127,6 +128,8 @@ beforeEach(function() {
     if (!os.settingsManager) {
       os.settingsManager = os.ui.config.SettingsManager.getInstance();
     }
+
+    os.time.replacers.init();
```
jsalankey, line 132: This feels like an odd place to also include and call this. Is there some reason we need it here and in `MainCtrl`?

src/os/ui/abstractmainctrl.js
schmidtk: All apps I'm aware of extend this base main controller, but if they don't they'll lose these URL replacers. We can make importing the file add the replacers because the circular dependency was related to the file that required it. Adding the require somewhere else would still work, but in general I don't like the idea of changing app behavior on import alone. An unintentional import could have unwanted side effects to an application.

test/os/os.mock.js
schmidtk: Tests don't load `MainCtrl`, so this file sets up managers and such that would normally be done there.

jsalankey: That might be more appropriate in `os.mock.js` then if its for tests.

jsalankey (approved):

PR #1303, opened 2021-06-16
chore(fa): updates to checkbox icon after fa5 upgrade

briedinger:
Fixing bugs with checkbox icon in other repos, updating checkbox to use proper new modifier and for consistency sake.

### DIFF
--- a/scss/os/_overrides_bootstrap.scss
+++ b/scss/os/_overrides_bootstrap.scss
@@ -49,7 +49,10 @@ $form-control-disabled-border: $gray-600 !default;
 .dropdown-item {
   cursor: pointer;
 
-  i.fa {
+  i.fa,
+  i.fab,
+  i.far,
+  i.fas {
     margin-right: ($spacer * .25);
   }
 }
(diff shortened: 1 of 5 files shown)


scss/os/_slicktree.scss
```diff
@@ -62,7 +62,9 @@
 
 // Node Icons
 .c-node-icons {
```
schmidtk, line 64: There is one more `.fa` specific style on line 12 of this file we'll want to update.

scss/os/_overrides_bootstrap.scss
```diff
@@ -49,7 +49,9 @@ $form-control-disabled-border: $gray-600 !default;
 .dropdown-item {
   cursor: pointer;
 
-  i.fa {
+  i.fa,
```
schmidtk, line 52: We should also add `.fab` to support the free brand icons.

schmidtk (changes requested):
Thanks for updating these Brian! I had some of these pending in #1299 but I should have isolated those from the greater change in that PR. Two quick requests, when you have a moment.

schmidtk (approved):

PR #1304, opened 2021-06-18
feat: added style controls to KML ground overlay

ameletyan:


ameletyan:
still need to deal with the following:
- failing to persist, `LayerSyncDescriptor.onLayerChange` is not being called on change
- visual bug where the screen goes white for a split second when altering the style sliders

src/os/ui/layer/abstractlayerui.js
```diff
@@ -159,6 +159,11 @@ os.ui.layer.AbstractLayerUICtrl.prototype.getLayerNodes = function() {
   var nodes = [];
 
   var items = this.scope['items'] || [];
+  for (var i = 0; i < items.length; i++) {
```
schmidtk, line 162: This is too specific for an abstract class. I would recommend making a new layer UI that extends `os.ui.layer.ImageLayerUI` (something like `plugin.file.kml.ui.KMLImageNodeUI`) and provides its own `getLayerNodes` function. When the image layer is created by the KML parser, use `setLayerUI` to override the default UI for the layer.

src/plugin/file/kml/ui/kmlnode.js
```diff
@@ -400,6 +408,16 @@ class KMLNode extends SlickTreeNode {
    */
   setImage(image) {
     this.image_ = image;
+    if (this.image_) {
+      // Give the image style control options.
+      var options = {};
+      options[ControlType.COLOR] = ColorControlType.PICKER_RESET;
+      this.image_.setLayerOptions(options);
```
schmidtk, line 415: This might be more appropriate in `readGroundOverlay_` in the KML parser, where the layer is created.

src/plugin/file/kml/ui/kmlnode.js
```diff
@@ -400,6 +408,16 @@ class KMLNode extends SlickTreeNode {
    */
   setImage(image) {
     this.image_ = image;
```
schmidtk, line 410: Layer nodes attach listeners to the layer, and are themselves event targets. We'll want to make sure any existing node gets disposed, or these listeners will prevent the object from being garbage collected and it will be leaked.

Prior to setting the image layer reference, check if `this.layerNode` is defined and if so dispose it. You can also safely do this without checking via:
```
goog.dispose(this.layerNode);
this.layerNode = null;
```
This will also ensure the node is cleaned up in `KMLNode#disposeInternal` when the image is set to null.

schmidtk (changes requested):

schmidtk (approved):

PR #1307, opened 2021-06-21
Transform alert/action/array/bearing/buffer/column/control/easing/events to modules

schmidtk:
Transforms the following under `src/os`:

- alert
- action
- array
- bearing
- buffer
- column
- control
- easing
- events

### Additional Changes

- The alert stack was mostly transformed, but we have a temporary file that makes the `os.alertManager` global available. This can be module with `declareLegacyNamespace` without breaking anything, so I converted it.
- Added `os.bearing.isGeomagLoaded` to check if the geomag library has finished loading. This avoids the need to check a private property in test initialization.
- `os.buffer.createFromConfig` is sometimes shared with an external window context. To do this, we place the reference in the external window. I added a setter to update this function, following the approach used in places/tracks/etc.
- Updated some API issues the compiler found when using `PropertyChangeEvent`.

### DIFF
--- a/src/os/control/mapmodecontrol.js
+++ b/src/os/control/mapmodecontrol.js
@@ -1,134 +1,139 @@
-goog.provide('os.control.MapMode');
+goog.module('os.control.MapMode');
+goog.module.declareLegacyNamespace();
 
-goog.require('goog.dom');
-goog.require('goog.dom.classlist');
-goog.require('ol');
-goog.require('ol.control.Control');
-goog.require('ol.css');
-goog.require('ol.events');
-goog.require('ol.events.EventType');
+const dom = goog.require('goog.dom');
+const Control = goog.require('ol.control.Control');
+const css = goog.require('ol.css');
+const events = goog.require('ol.events');
+const EventType = goog.require('ol.events.EventType');
+const dispatcher = goog.require('os.Dispatcher');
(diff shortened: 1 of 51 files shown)


src/os/control/mapmodecontrol.js
```diff
@@ -1,134 +1,140 @@
...
+const dom = goog.require('goog.dom');
+const GoogEventType = goog.require('goog.events.EventType');
+const Control = goog.require('ol.control.Control');
+const css = goog.require('ol.css');
+const events = goog.require('ol.events');
+const EventType = goog.require('ol.events.EventType');
+const dispatcher = goog.require('os.Dispatcher');
+const MapChange = goog.require('os.MapChange');
+const osActionEventType = goog.require('os.action.EventType');
+const mapInstance = goog.require('os.map.instance');
```
jsalankey, line 13: Nit: we've generally preferred to destructure requires when only calling 1 or 2 functions off of them. Doesn't make any functional difference, but if we set up the jscodeshift script to do these replaces it would be a bit more consistent to just get the `getMapContainer` function.

schmidtk: Updated and resolved the merge conflict.

jsalankey (approved):

PR #1308, opened 2021-06-23
build(linthtml): disable html linting for release branch

smithtb:
Temporary fix only for release builds


kwilson-bits (approved):

schmidtk:
If you pin the linter version this shouldn't be necessary. We should make sure this doesn't get merged back to master.

PR #1309, opened 2021-06-24
Add UI to directly compare layers

schmidtk:
This adds a new UI that compares two layers side-by-side with a slider control in the middle. To get to the UI, select two layers in the Layers window, right-click (or hit the new Selected Actions button), and choose Compare Layers.

### Other Changes

- Changed the default width of the Layers window to avoid collapsing the new/updated buttons by default. Layers window defaults can also be configured via the `os.layers.defaults` settings key.
- The Toggle Tiles button now correctly sets the enabled state (checkbox) instead of the layer visibility state (eye icon, which isn't displayed on tile layers).

### DIFF
--- a/src/os/ui/menu/defaultwindowsmenu.js
+++ b/src/os/ui/menu/defaultwindowsmenu.js
@@ -1,6 +1,7 @@
 goog.provide('os.ui.menu.windows.default');
 
 goog.require('os.config.ServerSettings');
+goog.require('os.config.Settings');
 goog.require('os.metrics.keys');
 goog.require('os.ui.events.UIEvent');
 goog.require('os.ui.events.UIEventType');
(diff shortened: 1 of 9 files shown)


src/os/ui/menu/defaultwindowsmenu.js
```diff
@@ -33,20 +45,23 @@ os.ui.menu.windows.default.setup = function() {
...
-    'show-close': 'true',
+    'x': layersDefaults.x != null ? layersDefaults.x : 12,
+    'y': layersDefaults.y != null ? layersDefaults.y : 96,
+    'width': layersDefaults.width != null ? layersDefaults.width : 400,
+    'height': layersDefaults.height != null ? layersDefaults.height : 550,
+    'min-width': layersDefaults.minWidth != null ? layersDefaults.minWidth : 300,
+    'max-width': layersDefaults.maxWidth != null ? layersDefaults.maxWidth : 1000,
+    'min-height': layersDefaults.minHeight != null ? layersDefaults.minHeight : 350,
+    'max-height': layersDefaults.maxHeight != null ? layersDefaults.maxHeight : 2000,
+    'show-close': layersDefaults.showClose != null ? layersDefaults.showClose : true,
```
jsalankey, line 64: Might be a bit cleaner to have a default object with all the hardcoded values in it and just merge the object from settings into it.

schmidtk: Yeah that's a better idea. Initially I had a few values here, and fell into the slippery slope of adding more things. I'll merge instead.

schmidtk: Fixed that and rebased to resolve a merge conflict.

jsalankey (approved):

jsalankey (approved):

PR #1310, opened 2021-06-24
build(lint): re-enable html linting and lint in parallel

schmidtk:
This was disabled on the release branch and merged into master.

I also made the lint steps run in parallel because they're all independent and this should speed up the build a bit.


csgibson (approved):

csgibson (changes requested):
html-validate should be moved to 4.14 to avoid a bug that breaks builds both local and CI

schmidtk:
It sounds like the correct solution is to upgrade `html-validate` and fix new errors. A dependency of `html-validate` must be causing the CI errors we're seeing given versions that used to work no longer work.

PR #1314, opened 2021-06-29
Transform the data stack to goog modules

schmidtk:
This transforms `os.data` and related code to goog modules.

### Breaking Changes

- `os.data.DataManager` and `os.data.OSDataManager` have been merged. Historically the distinction was made to separate out layer/source management and to provide a manager that didn't have a dependency on `os.MapContainer`. In practice there were a ton of places referencing the base class but expecting `OSDataManager`. This resulted in a large number of compiler errors when the type was interpreted correctly. Now that we have a way to reference the map container without a direct dependency, merging the two classes seemed appropriate. Applications that do not have a map/layers/sources will not be impacted, as that part of the manager will remain unused.
- `os.osDataManager` no longer exists. Applications should reference `os.dataManager` or preferably `os.data.DataManager.getInstance()`.
- The data manager now has a `setMapContainer` function that must be called to support managing sources. This should be called in the main controller when the map/data manager is initialized.
- Added `os.state.instance` to provide access to the state manager without a direct dependency. This resolves circular dependencies that came up in the module transition. Applications will have to call `setStateManager` in their main controller and state tests to ensure the global instance is available.

### Other Changes

- Several classes outside `os.data` were also transformed because they extended classes that had been transformed.
- Transforming these files identified a lot of places where JSDoc types were incorrect or not specific enough, resulting in a lot of new compiler errors. These have all been resolved by providing the expected type for the compiler.
- The `areafilteradd` directive was unused and expected a descriptor type that is not available in OpenSphere. It has been moved to an external library.


src/plugin/xyz/xyzdescriptornodeui.js
```diff
@@ -90,6 +92,14 @@ class Controller extends DescriptorNodeUICtrl {
     };
     uiWindow.create(windowOptions, editUi, undefined, undefined, undefined, scopeOptions);
   }
+
+  /**
+   * @override
+   * @return {ConfigDescriptor}
+   */
+  getDescriptor() {
+    return /** @type {ConfigDescriptor} */ (super.getDescriptor());
```
jsalankey, line 101: Bit of a weird override, is this cleaner than casting the return value where its called in here?

schmidtk: I can't recall why I thought this was a good idea. Replaced with a cast on the call.

jsalankey (approved):

PR #1318, opened 2021-07-01
fix(mappings): deconflict user and exec mappings

arwatts16:


src/plugin/ogc/wfs/wfslayerconfig.js
```diff
@@ -347,10 +351,22 @@ class WFSLayerConfig extends AbstractDataSourceLayerConfig {
...
 
+
+  /**
+   * Returns if a mapping is elliptical
+   * @param {IMapping} mapping
+   * @return {boolean}
+   */
+  isEllipticalMapping(mapping) {
+    const field = mapping.toField;
+    return RadiusMapping.REGEX.test(field) ||
```
schmidtk, line 366: Can you compare `mapping.getId()` to the `ID` for each of these mappings? Running a regex on each seems more expensive than necessary. Similar to the `updateMappings` function in the ellipse columns controller.

src/plugin/ogc/wfs/wfslayerconfig.js
```diff
@@ -333,11 +333,15 @@ class WFSLayerConfig extends AbstractDataSourceLayerConfig {
       new SemiMinorMapping()
     ];
 
+    const hasEllipse = options['mappings'].some((temp) => this.isEllipticalMapping(temp));
```
schmidtk, line 336: We should probably sanity check that `options['mappings']` is actually defined. Our WFS descriptor always sets it, but this layer config could be used elsewhere without a similar change.

schmidtk (changes requested):

schmidtk (approved):

PR #1319, opened 2021-07-02
Transform map and various utilities to goog modules

schmidtk:
This transforms most of `src/os/*.js` to modules and a handful of other utility type namespaces (geom/map/proj/object/track). This included `MapContainer` which was shockingly uneventful, and most of `src/os/map`. I deferred on some more complicated files (`os.map`), and the core os files like `os.js`, `main.js`, and `mainctrl.js`.

### Change Notes

- `os.instanceOf` and `os.registerClass` are both provided but share a private map. I moved this code into a `os.classRegistry` module, and converted the old provides to separate modules that pass through to the new module. I did not flag these as deprecated at this point because I didn't want to add to the build noise, but at some point we should deprecate/replace/remove the old functions.
- When converting the projection switch code, I ran into a cycle between `os.proj.switch` and `SwitchProjection`. This was solely due to `SwitchProjection` reading the enable raster reprojection flag, so I moved that bit to `os.proj.reprojection` and refactored references to that flag. That value is set by `MapContainer` as well, so I created a setter to avoid issues when converted to an ES module.
- `os.track` has a few functions that are used in external window contexts in other apps, so I created setters for those following the pattern in `plugin.track`.
- I transformed most of `src/os/map`, but `os.map` has a number of properties like `PROJECTION` that are modified, and I didn't want to dig into resolving those just yet.


schmidtk:
Resolved the merge conflict from #1309.

jsalankey (approved):

PR #1321, opened 2021-07-06
fix(electron): set maximum memory in preload script and fix type

schmidtk:
Changes to support upgrading Electron to the latest (currently 13.1.6).

This PR supplements ngageoint/opensphere-electron/pull/42.

- Loads maximum memory config in the preload script.
- Electron 13.x limits types that can be passed across the context bridge, and numbers are not allowed. Converted the memory value to a string before sending to the main process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -282,7 +282,7 @@
     "npm-run-all": "^4.1.5",
     "opensphere-build-closure-helper": "^7.0.0",
     "opensphere-build-index": "^4.0.0",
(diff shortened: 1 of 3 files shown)


jsalankey (approved):

schmidtk:
Upgraded the resolver to 9.0.0 to pick up the Electron related changes, and rebased against master.

jsalankey (approved):

PR #1324, opened 2021-07-07
Transform config stack to goog modules

schmidtk:
This transforms all config/settings/storage code to modules.

### Notable Changes

- Added a `os.config.instance` module to provide access to the global settings instance without a direct dependency. This instance is managed directly by the Settings module `getInstance` and `setInstance` calls, so external apps/projects do not need to call this.
- Use of `os.config.ThemeSettings.Keys` was causing a circular dependency between `ThemeSettings` and the UI. I moved the `Keys` enum to `os.config.theme` instead.
- UI shims were not created, so those module names changed.

For all breaking changes, all known downstream projects were updated accordingly.


src/os/config/configinstance.js
```diff
@@ -0,0 +1,35 @@
+goog.module('os.config.instance');
```
jsalankey, line 1: Is this singleton pattern necessary for `Settings` (and by extension, all singletons, even if they don't pose major circular dependency issues?) With settings in particular we use it EVERYWHERE, and switching them all over just seems kind of annoying.

schmidtk: I'm only creating these in places where a circular dependency exists, so it's helpful to have a way to reference the singleton without a hard dependency. Using it isn't expected/required, unless it resolves a circular dep.

jsalankey: Gotcha. Wouldn't have expected any circular dependencies with settings since stuff usually just requires it in and it sits at the top of any dependency tree and doesn't require too much stuff itself.

jsalankey (approved):

src/os/config/configinstance.js
schmidtk: The cycle was between `os.config` and `os.config.Settings` (uses appNs/coreNs). I didn't want to move anything in `os.config` because it's used all over the place, so I created the instance module instead.

PR #1327, opened 2021-07-13
Transform annotation/ex/feature and UI's (datetime through file) to goog modules

schmidtk:
This branch is based off #1326 due to related changes. Leaving this as a draft until that is merged.

### Other Changes

- The annotation classes were calling the parent constructor last due to the `initialize` methods in the abstract classes. That does not work with ES classes due to the "no this before super" rule, so I refactored those classes to initialize properly. The UI classes now use Angular's `$onInit`, and `FeatureAnnotation` now calls `createUI` instead of the abstract class.
- The drawing layer was causing a circular dependency with files that needed to reference its ID. I created `os.layer.LayerId` (similar to the `LayerClass` and `SourceClass` modules) to resolve this.
- `os.ui.file.kml` had some properties that were set outside the module. This does not work with a directly exported property, so I made get/set methods for the properties (`isGoogleMapsAccessible` and `mirror`), and replaced all references.
- The feature info UI had a circular dependency due to the Angular event types it fires. I moved the events to `FeatureInfoEvent` to resolve this.
- The mock async command had an arbitrary 1 second timeout to mimic an async command. This was adding up to 10+ seconds of unnecessary delays running tests. I replaced this with an immediate timeout to defer execute/revert until after the call stack cleared (still async), and added an optional parameter to that class to allow manually calling the execute/revert functions for tests that perform checks between execution states.


schmidtk:
Rebased to resolve the merge conflict.

jsalankey (approved):

PR #1329, opened 2021-07-13
feat(importer): alert the user when a mapping a bad column

arwatts16:
Send the user an alert when the mapping results in an undefined column. It should tell the user what mapping failed, and how many out of the total failed. A warning message appears is some but not all features fail to map, an error message appears if all features fail to map. 

~~This will only show for mappings that use the `toField` attribute (anything extending `os.im.mapping.RenameMapping`)~~
This will now only show for ellipse mappings (radius, semi major, semi minor, and orientation).

I have hit a snag though with CSV files and being able to get the total number of results from the importer that I'm not sure how to work around just yet.

### DIFF
--- a/src/os/im/importer.js
+++ b/src/os/im/importer.js
@@ -67,11 +67,11 @@ os.im.Importer = function(parser) {
   this.mappings = null;
 
   /**
-   * Holds count of how many mappings failed in each mapping
+   * Holds count of how many mappings failed in each mapping that uses the toField attribute
    * @type {!Object<string, number>}
    * @protected
    */
-  this.failedMappings = {};
+  this.mappingFailCount = {};
 
   /**
    * the list of mappings for use with autodetection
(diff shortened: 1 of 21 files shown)


src/os/im/importer.js
```diff
@@ -532,18 +615,42 @@ os.im.Importer.prototype.executeMapping = function(item) {
...
+
+
 /**
  * @inheritDoc
  */
 os.im.Importer.prototype.reset = function() {
   this.stop();
   this.parser.cleanup();
   this.data_ = [];
+  this.totalCount = 0;
```
bcleveland1022, line 653: does the mappingFailCount need to be cleared?

bcleveland1022 (changes requested):

src/os/im/importer.js
arwatts16: For the reset function, probably not. I was just putting it (almost) everywhere the data was reset to avoid double-counting any features.

bcleveland1022 (approved):

schmidtk (changes requested):
Discussed externally, but this ellipse logic is far too specific for `os.im.Importer` and should be moved to a more appropriate place.

PR #1330, opened 2021-07-14
Range Ring Export

jkae:
- Enable exporting range ring features from Saved Places to polygons (in KML)
- BREAKING: `plugin.places.menu.EventType` moved to `plugin.places.EventType` module


src/plugin/places/placeseventtype.js
```diff
@@ -0,0 +1,24 @@
+goog.module('plugin.places.EventType');
```
bcleveland1022, line 1: can this match filename

bcleveland1022 (changes requested):

bcleveland1022 (approved):

PR #1332, opened 2021-07-16
feat: test ellipse configuration columns for config-based suggestions

jkae:
- feat: "Suggested Columns" optgroup for ellipse column selections (csv import, etc.)
- feat: created a directive to help with suggested columns / other situations that call for dynamic optgroups from separate lists
- fix: resolve logic / naming issues in ellipse column directive
- refactor: goog.module conversions for some import mapping types

### DIFF
--- a/config/settings.json
+++ b/config/settings.json
@@ -34,7 +34,6 @@
         }
       ]
     },
-    "allowEllipseConfiguration": true,
     "metrics": {
       "enabled": true
     },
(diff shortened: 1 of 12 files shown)


arwatts16:
Are there related MRs for the highside configs to move `allowEllipseConfiguration`?

bcleveland1022 (dismissed):

schmidtk (changes requested):

schmidtk (approved):

PR #1334, opened 2021-07-20
fix(places): places now save their uninterpolated geometries

jsalankey:
Places can now be modified properly based on their originally drawn geometries rather than their interpolated ones. This should not affect exports of KML geometries in other contexts, only when we write them to Places.

### DIFF
--- a/src/os/interpolate.js
+++ b/src/os/interpolate.js
@@ -34,6 +34,7 @@ const SettingsKey = {
 let TransformSet;
 
 /**
+ * Local interpolate config.
  * @type {!Config}
  */
 const interpolateConfig = {
@@ -68,8 +69,7 @@ let overrideMethod = null;
  * @return {boolean} Whether or not all the values needed for interpolation are present
  */
 const getEnabled = function() {
-  var config = interpolateConfig;
-  return !!(interpolateEnabled && config.distance > 1000);
+  return !!(interpolateEnabled && interpolateConfig.distance > 1000);
 };
 
 /**
@@ -94,7 +94,6 @@ const getConfig = function() {
  */
 const setConfig = function(config) {
   if (config) {
-    var interpolateConfig = interpolateConfig;
     interpolateConfig.method = /** @type {Method} */ (config['method']) || interpolateConfig.method;
     interpolateConfig.distance = /** @type {number} */ (config['distance']) || interpolateConfig.thresholdPercent;
   }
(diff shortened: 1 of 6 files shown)


src/os/interpolate.js
```diff
@@ -34,9 +34,10 @@ const SettingsKey = {
 let TransformSet;
 
 /**
+ * Local interpolate config.
  * @type {!Config}
  */
-const interpolateConfig = {
+const interpolateConfig_ = {
```
schmidtk, line 40: Historically we used underscores for private properties so it would be very clear if that property was being referenced outside the file/class/etc (ie, `obj.someProperty_` would be an immediate red flag). For properties local to a module I don't think the underscore is necessary because the property is truly private unless exported.

src/plugin/file/kml/kmltreeexporter.js
```diff
@@ -321,7 +322,7 @@ class KMLTreeExporter extends AbstractKMLExporter {
       var geomAltitudeMode;
       var featAltitudeMode = feature.get(RecordField.ALTITUDE_MODE);
 
-      geometry = feature.getGeometry();
+      geometry = /** @type {Geometry} */ (feature.get(os.interpolate.ORIGINAL_GEOM_FIELD)) || feature.getGeometry();
```
schmidtk, line 325: Global reference to `os.interpolate`.

src/plugin/places/placesmenu.js
```diff
@@ -772,7 +772,7 @@ const saveSpatialToPlaces = function(event, opt_annotation) {
     if (features.length === 1) {
       name = osFeature.getTitle(features[0]);
 
-      var featureGeom = features[0].getGeometry();
+      var featureGeom = features[0].get(os.interpolate.ORIGINAL_GEOM_FIELD) || features[0].getGeometry();
```
schmidtk, line 775: Another global ref here.

src/plugin/places/ui/quickaddplaces.js
```diff
@@ -156,6 +157,12 @@ class Controller extends Disposable {
         startTime: Date.now()
       }));
 
+      // re-interpolate the feature now to ensure that it has the original geometry and correct interpolation method
+      const method = /** @type {interpolate.Method} */ (geometry.get(os.interpolate.METHOD_FIELD));
```
schmidtk, line 161: Global ref.

schmidtk (changes requested):

src/os/interpolate.js
jsalankey: I went with the underscore to follow some of the style I saw you doing with set functions in other modules. This case was actually causing an exception on interpolation change due to the collision in variable names in the `setConfig` function. I can give it another name, but I think indicating that it is private to this module is useful.

schmidtk: Ahh yeah, those could have been named better. We started that pattern because those are the exact same function, one is exported and the other is a reference that can be changed. We have always followed Google's JS style guide though, which explicitly states to not use `@private` or end the name in an underscore.

https://google.github.io/styleguide/jsguide.html#file-goog-module-exports

schmidtk (approved):

PR #1335, opened 2021-07-20
Transform os.fields through os.histo to goog modules

schmidtk:
This branch is based off of #1327 so extra commits will be present until that is merged.


src/os/debug/fancierwindow.js
```diff
@@ -5,9 +5,10 @@ goog.require('goog.debug.FancyWindow');
 goog.require('goog.html.SafeStyleSheet');
 goog.require('goog.string.Const');
 goog.require('os.alert.AlertManager');
-goog.require('os.file.persist.FilePersistence');
 goog.require('os.ui');
```
jsalankey, line 8: This file still has a bunch of naked requires in it. Doesn't look like it was explicitly converted as part of this ticket, but we'll need to hit it eventually

schmidtk: Good call, this probably got transformed before I created the `replaceglobals` script. It's updated now.

jsalankey (approved):

PR #1336, opened 2021-07-20
Ability to export/import true Range Ring features

jkae:
- feat(export): ring options exported to KML if present on feature
- fix(rings): move `opt_replace` check inside `if (feature)` to prevent potential runtime error
- fix(rings): get `LON` and `LAT` from data if geometry is not a point -- fixes issue where importing an arc polygon, with ring options, will reposition and be off-model

### DIFF
--- a/src/os/feature/feature.js
+++ b/src/os/feature/feature.js
@@ -528,19 +528,28 @@ const createLineOfBearing = function(feature, opt_replace, opt_lobOpts) {
 const createRings = function(feature, opt_replace) {
   var ringGeom;
 
-  if (!opt_replace) {
-    ringGeom = /** @type {GeometryCollection} */ (feature.values_[RecordField.RING]);
-    if (ringGeom instanceof GeometryCollection) {
-      return ringGeom;
-    }
-  }
-
   if (feature) {
+    if (!opt_replace) {
+      ringGeom = /** @type {GeometryCollection} */ (feature.values_[RecordField.RING]);
+      if (ringGeom instanceof GeometryCollection) {
+        return ringGeom;
+      }
+    }
+
     cleanRingGeoms(feature);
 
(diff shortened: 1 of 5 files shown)


src/os/feature/feature.js
```diff
@@ -528,19 +528,28 @@ const createLineOfBearing = function(feature, opt_replace, opt_lobOpts) {
...
-  if (feature) {
     cleanRingGeoms(feature);
 
     var options = /** @type {osx.feature.RingOptions} */ (feature.get(RecordField.RING_OPTIONS));
     var geometry = feature.getGeometry();
-    var center = geometry ? toLonLat(olExtent.getCenter(geometry.getExtent()), osMap.PROJECTION) : null;
+    var center = null;
+
+    if (geometry && geometry instanceof Point) {
+      center = toLonLat(olExtent.getCenter(geometry.getExtent()), osMap.PROJECTION);
```
schmidtk, line 546: For a Point feature, `geometry.getCoordinates()` should be the same as getting the extent center.

src/os/ui/file/kml/abstractkmlexporter.js
```diff
@@ -606,6 +607,8 @@ class AbstractKMLExporter extends ZipExporter {
     this.addGeometryNode(item, element);
 
     var fields = this.getFields(item);
+    fields.push(RecordField.RING_OPTIONS);
```
schmidtk, line 610: Can this be included in the `getFields` result? It feels like a random one-off to add in here.

For the standard Export path, we have never exported internal fields like this and only include them if you choose Export Places, which is intended to preserve all OpenSphere-specific metadata. The normal Export path should export geometries/styles as currently rendered on the map, and no additional metadata because it's intended to be consumed by external apps/services like Google Earth.

schmidtk (changes requested):

schmidtk (approved):

PR #1337, opened 2021-07-20
fix(style): sets opacity properly on circle styles

jsalankey:
The Cesium synchronizer reads point feature opacity off of the associated circle style directly with `getOpacity()`, but the opacity was never being set on the style. This sets that opacity directly so it can be used by the synchronizer and anything else that might want it.

### DIFF
--- a/src/os/style/circlereader.js
+++ b/src/os/style/circlereader.js
@@ -1,6 +1,8 @@
 goog.provide('os.style.CircleReader');
 
+goog.require('goog.string');
 goog.require('ol.style.Circle');
+goog.require('os.color');
 goog.require('os.style');
 goog.require('os.style.AbstractReader');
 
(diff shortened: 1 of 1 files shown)


src/os/style/circlereader.js
```diff
@@ -42,14 +45,32 @@ os.style.CircleReader.prototype.getOrCreateStyle = function(config) {
...
+    var style = new ol.style.Circle({
       radius: radius,
       fill: fill,
       stroke: stroke
     });
 
-    this.cache[hash]['id'] = hash;
+    if (opacity != null) {
+      // set the opacity if it's defined, this is used by the WebGL synchronizers and is not passed by the options
+      // up to the parent style class for some reason (???)
```
schmidtk, line 67: From OL's point of view, the Image style's opacity value isn't needed for the RegularShape style because it has separate fill/stroke colors each with their own opacity. The Icon style does use the value despite having a color property, but icon color is optional to render something where a shape has to have a fill or stroke to be visible.

schmidtk (approved):

PR #1338, opened 2021-07-21
fix: feature action ui keeps order and collapsed states on edit

ameletyan:


src/os/ui/im/action/filteraction.js
```diff
@@ -165,24 +165,25 @@ const onEditComplete = function(original, entry) {
...
-
       var parent = original.getParent();
       if (parent) {
         parentId = parent.getId();
+        searchArray = parent.getChildren();
+      } else {
+        searchArray = iam.getActionEntries(entry.getType());
       }
 
+      insertIndex = olArray.findIndex(searchArray, function(entry) {
```
jsalankey, line 183: Nit: `findIndex` is now a natively supported function on arrays, so the preferred way to do this would be `searchArray.findIndex(...)` rather than using the OL utility function. We prefer to clean up old uses of functions like this in favor of native functions whenever possible these days.

jsalankey (approved):

PR #1340, opened 2021-07-22
Transform os.im through os.metrics to goog modules

schmidtk:
- Types for `IMapping#field` and the `field` param to `getItemField` and `setItemField` conflicted, causing compiler errors. I updated the two convenience functions to handle null/undefined values so the mapping field doesn't need to be tested before calling these.
- The compiler found a number of typing issues in the code. All of these have been resolved with appropriate `@type` annotations.
- The `anytype` import was adding a `title` property to importers, from the `getTitle` function call. This was solely so the Angular template could reference `importer.title`. Properties should not be added to classes after they're created, so I added a `getTitle` function on the controller to get the value from the importer.
- Several sets of metrics keys were defined on `os.metrics`, but in the `os.metrics.keys` module. I deprecated the old namespace, and preferring the `os.metrics.keys` enum instead. All references to these have been replaced.


src/os/data/layertreesearch.js
```diff
@@ -57,7 +57,7 @@ class LayerTreeSearch extends AbstractGroupByTreeSearch {
     var layer = /** @type {!ILayer} */ (item);
     var node;
 
-    if (layer.getTreeNode) {
+    if (typeof /** @type {!ITreeNodeSupplier} */ (layer).getTreeNode === 'function') {
```
jsalankey, line 60: Might be annoying to add all the cases as implementers of `ITreeNodeSupplier`, but it would be less gross to do an `implements` check here.

schmidtk: Yeah I didn't want to add a breaking change to fix a typing issue from the original code. The `typeof` check just made the original check a little more specific, but also shouldn't be necessary.

jsalankey (approved):

PR #1348, opened 2021-08-18
fix(export): updates export API to support statusing

jsalankey:
Adds a UI and API for export progress statusing. This is currently not used in the core OpenSphere exporters, but is used by the GeoPackage plugin.

BREAKING CHANGE: adds new method to IExportMethod


src/os/ui/file/exportstatus.js
```diff
@@ -0,0 +1,182 @@
...
+ * Launches the export progress window for a given exporter.
+ * @param {IExportMethod} exporter
+ */
+export const launch = (exporter) => {
+  var scopeOptions = {
+    'exporter': exporter
+  };
+
+  var container = angular.element(windowSelector.CONTAINER);
+  var x = container.width() - 350;
```
schmidtk, line 166: Windows treat negative `x` values as offsets from the right edge of the screen, so you could do `'x': -50` here.

schmidtk (changes requested):
This looks good. I added a minor suggestion (optional), but we have a merge conflict to resolve from my last batch of transforms.

src/os/ui/file/exportstatus.js
jsalankey: Huh, always thought that did the offset from the left side of the window. Didn't think it accounted for the width properly.

PR #1350, opened 2021-08-18
Fix interpolation breaking places

ameletyan:


ameletyan:
Pulled it out of review, found an issue in the compiled build.

jsalankey (approved):

PR #1352, opened 2021-08-19
Feat nav bar labels

ameletyan:


src/os/ui/draw/drawcontrols.js
```diff
@@ -23,7 +23,8 @@ const directive = () => ({
   replace: true,
   scope: {
     'menu': '=?',
-    'olMap': '=?'
+    'olMap': '=?',
+    'showLabel': '='
```
jsalankey, line 27: The `'=?' notation is meant to indicate that a parameter is optional, so we prefer that for one like this where the directive is used elsewhere without it.

Angular used to throw an exception if you instantiated a directive without a required parameter (hence why we use the ? for the optional menu and map here). I think they may have changed that, but still a good bit of habit for these directives.

jsalankey (approved):

PR #1354, opened 2021-08-25
feat: added warning when comparing layers with more than 50k features

ameletyan:


src/os/ui/layer/compare/layercompareui.js
```diff
@@ -462,6 +466,12 @@ const windowId = 'compare-layers';
  * @param {!LayerCompareOptions} options The layer compare options.
  */
 const launchLayerCompare = (options) => {
+  const featureCount = countFeatures(options.left) + countFeatures(options.right);
+  if (featureCount > getMaxFeatures('2d')) {
+    AlertManager.getInstance().sendAlert('Switching to 2D mode with the current data volume may degrade performance ' +
```
jsalankey, line 471: Wording this as "switching to 2D mode" doesn't make sense for the layer compare since you aren't switching modes, its simply the mode we support for it. This should be something along the lines of "Using Layer Compare with the current data volume..."

src/os/ui/layer/compare/layercompareui.js
```diff
@@ -493,6 +503,27 @@ const launchLayerCompare = (options) => {
...
+ * @param {Array<Layer>=} layerArray An array of layers.
+ * @return {number} The total number of features in the layer array.
+ */
+const countFeatures = (layerArray) => {
+  if (layerArray && layerArray.length > 0) {
+    var featureCount = 0;
+    layerArray.forEach((layer) => {
+      var source = /** @type {ISource} */ (layer.getSource());
+      if (source) {
+        featureCount += source.getFeatures().length;
```
schmidtk, line 518: You should first verify the source is an `ISource` or this line will break, for example when comparing tile layers.

src/os/ui/layer/compare/layercompareui.js
```diff
@@ -462,6 +469,12 @@ const windowId = 'compare-layers';
  * @param {!LayerCompareOptions} options The layer compare options.
  */
 const launchLayerCompare = (options) => {
+  const featureCount = countFeatures(options.left) + countFeatures(options.right);
+  if (featureCount > getMaxFeatures('2d')) {
+    AlertManager.getInstance().sendAlert('Using Layer Compare with the current data volume may degrade performance ' +
```
schmidtk, line 474: An alert is informative for the user, but doesn't prevent the bad behavior from happening. If a user opens compare with 150k features on a slow machine, they might crash their browser.

I think it would be better to treat this like our 3D -> 2D switch in the same circumstance where we first pop up a warning confirmation to make sure they want to continue. You can look at `os.webgl.launch2DPerformanceDialog` for how that is currently done. We'll want the dialog's text to be specific to Layer Compare, but otherwise it should be largely the same. If they confirm, launch the dialog. If they cancel, do nothing.

schmidtk (changes requested):
I think we should be a bit more invasive to protect the user from crashing their browser or killing performance. This will also be consistent with what happens if you try switching to 2D mode with too much data loaded. It also looks like there are merge conflicts to resolve with the master branch.

src/os/webgl/launchlayercompareperfdialog.js
```diff
@@ -0,0 +1,37 @@
...
+const Promise = goog.require('goog.Promise');
+const {launchConfirm} = goog.require('os.ui.window.ConfirmUI');
+
+
+/**
+ * Launch a dialog warning users of the risks in using 2D with lots of data.
+ *
+ * @return {!Promise}
+ */
+const launch2DPerformanceDialog = function() {
```
schmidtk, line 12: Minor nit. I would suggest putting this in `layercompareui.js` because it's not particularly useful outside of the `launchLayerCompare` function.

src/os/ui/layer/compare/layercompareui.js
```diff
@@ -541,6 +547,45 @@ const launchLayerCompare = (options) => {
...
+ * Launch the layer compare.
+ * @param {!LayerCompareOptions} options The layer compare options.
+ */
+const launchLayerCompare = (options) => {
+  const featureCount = countFeatures(options.left) + countFeatures(options.right);
+  if (featureCount > getMaxFeatures('2d')) {
+    launchLayerComparePerformanceDialog().then(() => {
+      launchLayerCompareWindow(options);
+    }, () => {
+      // This empty block prevents an undefined error from appearing
```
schmidtk, line 561: Very true, but doesn't explain "why". Promises will throw an error if they're rejected unless the rejection is handled, because rejection is by default considered to be an error. The "why" here is that we want to silently ignore the rejection because it was due to the user choosing not to open the window.

schmidtk (changes requested):
This is looking good! I have one minor cleanup nit to consolidate the extra perf module.

schmidtk (approved):

PR #1360, opened 2021-09-22
Button style and consolidation

jsalankey:
Updates the top navbar buttons by consolidating the Draw and Measure buttons into a single button/menu and the Save and State buttons in a similar way. Functionality should all be retained in merging these buttons, but lessens the redundancies in the behaviors present in the top nav. This also reworks some of the button colors, getting rid of brighter colors that were unnecessary and lent to calling out certain buttons more than they needed to be.


src/os/ui/draw/drawcontrols.js
```diff
@@ -106,8 +178,80 @@ class Controller extends BaseDrawControlsCtrl {
...
+    this.toggleMeasure(true);
+  }
 }
 
+/**
+ * Helper function for changiing the icons on measure options.
+ * @this {MenuItem}
+ */
+const updateIcons = function() {
+  if (this.eventType === Measure.method) {
```
schmidtk, line 248: Do these still mean anything in context of the Draw menu? In the Measure menu they represented the "active" measure action we would use if you clicked the measure button, but we may want to use a measure type icon (ruler or drafting compass) now.

src/os/ui/draw/drawcontrols.js
```diff
@@ -106,8 +178,80 @@ class Controller extends BaseDrawControlsCtrl {
...
+   */
+  onMeasureTypeChange_(evt) {
+    Measure.method = /** @type {Method} */ (evt.type);
+    Settings.getInstance().set(this.key_, evt.type);
+    this.toggleMeasure(true);
+  }
 }
 
+/**
+ * Helper function for changiing the icons on measure options.
```
schmidtk, line 244: Nit: `changiing`

src/os/ui/draw/drawcontrols.js
```diff
@@ -57,16 +68,77 @@ class Controller extends BaseDrawControlsCtrl {
...
+
+  /**
+   * @inheritDoc
+   */
+  initControlMenu() {
+    super.initControlMenu();
+
+    var mi = this.controlMenu.getRoot();
+    mi.addChild({
+      type: MenuItemType.SEPARATOR,
```
schmidtk, line 111: Should these use `MenuItemType.GROUP`? I'm not sure what the difference is here between that and using a separator with a label.

src/os/ui/menu/menuitem.js
```diff
@@ -187,15 +187,19 @@ class MenuItem {
...
 
+    var classes = [];
+    var isItem = type !== types.SUBMENU && type !== types.GROUP && type !== types.SEPARATOR;
+
     html += '<li';
     if (type === types.SEPARATOR) {
-      html += '>-</li>';
-      return html;
+      if (this.label) {
+        classes.push('dropdown-header');
```
schmidtk, line 196: The `SEPARATOR` type is for rendering a line between items, and this seems to be duplicative of the `GROUP` type. Does this add anything over using `GROUP`?

schmidtk (changes requested):

src/os/ui/draw/drawcontrols.js
jsalankey: Separators were originally only rendered as a horizontal line. I changed the rendering to support showing them as a sorted header with no underlying hierarchical structure. This was more convenient for cases where I didn't want to refactor the menu to add stuff to a child group, I just wanted a separator with some text on it.

src/os/ui/menu/menuitem.js
jsalankey: Switched these all over to use `GROUP` and added a breaking change commit.

src/os/mapcontainer.js
```diff
@@ -776,6 +776,8 @@ class MapContainer extends EventTarget {
     // Map
     controls.addControl(genMapGrp, 1, 'Draw Geometry',
         [KeyCodes.SHIFT, '+'], [Controls.MOUSE.LEFT_MOUSE, '+', Controls.FONT.ALL]);
+    controls.addControl(genMapGrp, 1, 'Measure',
+        [platformModifier, '+', KeyCodes.SHIFT, '+'], [Controls.MOUSE.LEFT_MOUSE, '+', Controls.FONT.ALL]);
```
schmidtk, line 780: Can we remove the drag part? If you mouse down and drag the interaction doesn't do anything, unless we fix that in the polygon interaction. We have drag on Draw Geometry for the box/circle interaction, but it doesn't work the same with Polygon and Measure.

src/os/ui/draw/drawcontrols.js
```diff
@@ -57,16 +68,94 @@ class Controller extends BaseDrawControlsCtrl {
...
+    // the main draw control UI should contribute the measure controls group if it isn't there already
+    var mi = this.controlMenu.getRoot();
+    const measureGroup = mi.find('Measure');
+
+    if (!measureGroup) {
+      mi.addChild({
+        type: MenuItemType.GROUP,
+        label: 'Measure',
+        tooltip: 'Options for measuring distances',
+        shortcut: 'Ctrl+Shift+click',
```
schmidtk, line 137: Can we display the platform modifier key here as well?

schmidtk (changes requested):
This is looking good. Two minor comments on the keyboard shortcuts.

src/os/mapcontainer.js
jsalankey: Yeah I could see this going either way. I see that as "move the mouse around", not necessarily just drag. Some sort of an "click and hold" indicator would suggest dragging to me.

schmidtk (approved):

PR #1364, opened 2021-09-28
Transform all UI's to ES modules

schmidtk:
This transforms all UI's to ES modules. All module transforms and dependency updates were done with `opensphere-jscodeshift` with no manual changes in the initial commit.

**Other Changes**

- Several tests were spying on `os.ui.window` functions like create/close to test if those functions were called. Those functions are now direct exports, which are sealed and cannot be spied on. To resolve these issues, `os.mock.js` now creates the default window container and bootstraps it with Angular. This allows windows to be actually opened/tested. I added several vendor libraries to the Karma config to support various UI's that are now being fully compiled in tests.
- ES modules now warn when importing deprecated API's. This introduced several lint warnings in the `os.ui.action` stack, which has been deprecated for awhile and will soon be removed. For now, I suppressed warnings within `os.ui.action` and left warnings in non-deprecated modules (slickgrid).


schmidtk:
Replaced by #1368.

PR #1366, opened 2021-09-30
fix: resolved issue when editing geometry from first vertex of a polygon

ameletyan:


src/os/interaction/modifyinteraction.js
```diff
@@ -409,6 +410,21 @@ class Modify extends OLModify {
...
+        // coordinates. Since it is not intended for XY coordinates to have a Z coordinate, these
+        // new values can potentially be undefined and cause problems with Cesium. To prevent this,
+        // set all undefined values to zero.
+        if (originalGeom instanceof Polygon) {
+          var flatCoordinates = originalGeom.getFlatCoordinates();
+          const isUndefined = (item) => item === undefined;
+          if (flatCoordinates && flatCoordinates.some(isUndefined)) {
+            while (flatCoordinates.some(isUndefined)) {
+              flatCoordinates[flatCoordinates.findIndex(isUndefined)] = 0;
+            }
```
ameletyan, line 423: This loop used to be simple:
`while (flatCoordinates.includes(undefined)) {
  flatCoordinates[flatCoordinates.indexOf(undefined)] = 0;
}`

However, this led to some issues during compile:
`ERROR in ./src/os/interaction/modifyinteraction.js:420 from closure-compiler: actual parameter 1 of Array.prototype.includes does not match formal parameter
found   : undefined
required: number
            while (flatCoordinates.includes(undefined)) {
                                                                 ^^^^^^^^^

ERROR in ./src/os/interaction/modifyinteraction.js:421 from closure-compiler: actual parameter 1 of Array.prototype.indexOf does not match formal parameter
found   : undefined
required: number
              flatCoordinates[flatCoordinates.indexOf(undefined)] = 0;
                                                                                  ^^^^^^^^^`

Since I was trying to find values that were `undefined`, I resorted to these less straight forward functions. Let me know if there is a more elegant way to replace `undefined` values.

schmidtk: This loop could be much more efficient. Both `some` and `findIndex` will iterate over the array until the condition is true, which means you're searching for undefined values from the start of the array for each undefined value. Given we have a flat array of coordinates, we at most should need to traverse the array once to replace any undefined values we encounter. We also _only_ want to replace undefined altitude values. We in theory shouldn't have undefined lat/lon, but we also don't want to replace those with 0's because then we're defining a coordinate that's probably not correct.

The fastest solution here would be to:
- Verify the geometry stride is >= 3 so we know it has altitude. The layout could be XYZ or XYZM (has time), but we only need to check that it's at least 3 to have the Z component.
- If stride is >= 3, iterate over the flat coords with a for loop that increments the iterator with `i += stride`, and sets `arr[i + 2] = 0` if the value is not already defined.

This will run in `O(n / stride)` time.

src/os/interaction/modifyinteraction.js
```diff
@@ -409,6 +410,21 @@ class Modify extends OLModify {
     if (this.clone_) {
       const originalGeom = this.clone_.getGeometry();
       if (originalGeom) {
+        // While handling events, OpenLayers may add a new Z (altitude) coordinate to a set of XY
+        // coordinates. Since it is not intended for XY coordinates to have a Z coordinate, these
+        // new values can potentially be undefined and cause problems with Cesium. To prevent this,
+        // set all undefined values to zero.
+        if (originalGeom instanceof Polygon) {
```
schmidtk, line 417: This is currently called each time the geometry is updated, which happens frequently during the modify interaction. We should at most need to call it once, because once the altitude values are fixed they will no longer be undefined.

We could do that when we clone the feature and add altitude values to the coordinate array, knowing OL will change the geometry layout to XYZ. However, the reason that happens is the event coordinate has a Z component in 3D mode which conflicts with OL's assumption that the event coordinate layout will match the modify geometry's layout. We can fix that in our `handleDownEvent_` override by checking if `evt.coordinate.length > stride` and if so, set the length to the stride to crop the coordinate. We will also want to override the `handleDragEvent_` function to make the same change. Between those, we can ensure the event coordinate matches the geometry we're modifying, and we won't have to change the geometry's layout at all.

schmidtk (changes requested):

schmidtk (approved):

PR #1371, opened 2021-10-12
feat(layer-compare): adds support for comparing multiple layers

jsalankey:
Allows loading multiple layers into the compare window as well as moving them left/right. Also adds basic layer controls.


src/os/ui/layer/compare/layercomparenode.js
```diff
@@ -0,0 +1,262 @@
...
+      if (config) {
+        const color = getConfigColor(config, true);
+        if (color) {
+          return createIconSet(id, icons, [], color);
+        }
+      }
+    } else if (this.layer_ instanceof Tile) {
+      const icons = [IconsSVG.TILES];
+      const layerColor = /** {Tile} */ (this.layer_).getColor();
+      let color = [256, 256, 256, 1];
```
schmidtk, line 155: 255's for these?

PR #1393, opened 2021-11-08
Feat layer compare improvements

jsalankey:
Makes the following improvements to the layer compare window:
- Adds the ability to zoom to layers in it with the right click menu.
- Allows adding the same layer to both sides of the compare for better comparisons against static data.
- Fixes issues with having the same layer on the same side more than once.
- Handles layers that are removed from the main map and disposed.

### DIFF
--- a/src/os/ui/layer/compare/layercompareui.js
+++ b/src/os/ui/layer/compare/layercompareui.js
@@ -1,5 +1,6 @@
 goog.declareModuleId('os.ui.layer.compare.LayerCompareUI');
 
+import EventType from '../../../action/eventtype.js';
 import * as capture from '../../../capture/capture.js';
 import LayerEventType from '../../../events/layereventtype.js';
 import {normalizeToCenter} from '../../../extent.js';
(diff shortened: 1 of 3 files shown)


src/os/ui/layer/compare/layercompareui.js
```diff
@@ -549,6 +609,77 @@ export class Controller {
...
 
+  /**
+   * Removes a layer from the compare.
+   * @param {Layer|ILayer} layer The layer to remove.
+   */
+  remove(layer) {
+    // please the compiler and our terrible layer typing
+    layer = /** @type {Layer} */ (layer);
+
+    const rightLayerArr = this.rightLayers.getArray().slice();
```
schmidtk, line 620: Nit, but as long as we're always creating a new array we could combine a few lines here with:
```
this.rightLayers.getArray().filter((item) => item !== layer);
```

schmidtk (changes requested):

schmidtk (approved):

PR #1398, opened 2021-11-09
Fix issues with Saved Places field display

### LINKED ISSUE #1080: Places mappings are not applied until loaded from storage
When creating new Saved Places, columns that are normally populated by mappings are empty until the application is refreshed and places are loaded from storage. This includes `TIME` and unit-normalized ellipse fields.

To reproduce:

- Right click the map and choose Create Place.
- Set time and ellipse fields and click OK.
- Double click the feature on the map. The time/ellipse fields will not be populated.
- Reload the application and double click the feature again. Fields are now populated, due to mappings executing on import.

schmidtk:
This fixes a couple of issues with fields displayed for Saved Places:

- The `TIME` column was being used for time, which doesn't exist until the place is exported/imported (the KML exporter saves the field). Instead, the internal `recordTime` column is now used by the source. This fixes time not displaying in Feature Info for new places.
- Altitude will now be displayed in Feature Info for places.
- Ellipse fields (semi major/minor and orientation) were not displaying for newly created places, but would display after app refresh. This is due to mappings being detected/executed by the KML importer on refresh. This was fixed by detecting/executing mappings when a place is created/edited, so fields should always be accurately displayed in Feature Info.

Fixes #1080.

### DIFF
--- a/src/plugin/places/places.js
+++ b/src/plugin/places/places.js
@@ -3,6 +3,7 @@ goog.declareModuleId('plugin.places');
 import * as annotation from '../../os/annotation/annotation.js';
 import CommandProcessor from '../../os/command/commandprocessor.js';
 import SequenceCommand from '../../os/command/sequencecommand.js';
+import ColumnDefinition from '../../os/data/columndefinition.js';
 import RecordField from '../../os/data/recordfield.js';
 import * as osFeature from '../../os/feature/feature.js';
 import Fields from '../../os/fields/fields.js';
@@ -110,12 +111,13 @@ export const SourceFields = [
   Fields.LAT_DMS,
   Fields.LON_DMS,
   Fields.MGRS,
+  Fields.ALT,
   Fields.SEMI_MAJOR,
   Fields.SEMI_MINOR,
   Fields.SEMI_MAJOR_UNITS,
   Fields.SEMI_MINOR_UNITS,
-  Fields.TIME,
-  Fields.ORIENTATION
+  Fields.ORIENTATION,
+  new ColumnDefinition(Fields.TIME, RecordField.TIME)
 ];
 
 /**
--- a/src/plugin/places/places.js
+++ b/src/plugin/places/places.js
@@ -96,9 +96,9 @@ export const ExportFields = [
 ];
 
 /**
- * Fields that should be displayed on the Places source.
+ * Fields that should be displayed on the Places source. These will be set on the source via ISource#setColumns.
  *
- * @type {!Array<string>}
+ * @type {!Array<!(ColumnDefinition|string)>}
  */
 export const SourceFields = [
   KMLField.NAME,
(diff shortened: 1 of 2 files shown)


jsalankey (dismissed):

src/plugin/places/places.js
```diff
@@ -110,12 +110,13 @@ export const SourceFields = [
...
   Fields.MGRS,
+  Fields.ALT,
   Fields.SEMI_MAJOR,
   Fields.SEMI_MINOR,
   Fields.SEMI_MAJOR_UNITS,
   Fields.SEMI_MINOR_UNITS,
-  Fields.TIME,
-  Fields.ORIENTATION
+  Fields.ORIENTATION,
+  RecordField.TIME
```
jsalankey, line 119: Was using the wrong time field before causing some issue?

schmidtk: The `TIME` column doesn't exist on Saved Places until app refresh, because the KML exporter saves it here:
https://github.com/ngageoint/opensphere/blob/0cbf7c7a2e00f8e20913abbdbd2cc13055c179b2/src/os/ui/file/kml/abstractkmlexporter.js#L600-L602

If you create a new Place and open Feature Info for it, the `TIME` value will be blank because that column doesn't exist. Feature Info will display the `recordTime` field as `TIME` so using that column here fixes the Feature Info display issue.

It looks like the Show Features grid does not do that mapping however. I wonder if we should run a time mapping to add the `TIME` column instead of this change.

schmidtk: I updated the column to use `RecordField.TIME` as the `field` and `Fields.TIME` as the `name`. This is consistent with how the time column is typically treated on vector sources (see `os.source#addDefaults`), and with the field `featureedit.js` is actually setting for the time.

src/plugin/places/places.js
```diff
@@ -110,12 +111,13 @@ export const SourceFields = [
...
   Fields.MGRS,
+  Fields.ALT,
   Fields.SEMI_MAJOR,
   Fields.SEMI_MINOR,
   Fields.SEMI_MAJOR_UNITS,
   Fields.SEMI_MINOR_UNITS,
-  Fields.TIME,
-  Fields.ORIENTATION
+  Fields.ORIENTATION,
+  new ColumnDefinition(Fields.TIME, RecordField.TIME)
```
jsalankey, line 120: Surprised the compiler didn't complain about this, but the type here is declared as `!Array<string>`. Seems like we just pass this later on as columns to the source, which handles both, but just to be sure we have it clear here.

schmidtk: Good call. I updated the type to match `ISource#setColumns`, which is where these fields are being used in `AbstractDataSourceLayerConfig`.

jsalankey (approved):

PR #1399, opened 2021-11-11
feat(shp): add multipoint and multipolygon import/export support

### LINKED ISSUE #433: Shapefile export splits MultiPolygons
When exporting to Shapefile, MultiPolygon geometries are split into individual polygons. They should be exported as a PolygonM, so when imported they are still a MultiPolygon.

schmidtk:
The SHP exporter previously split both `MultiPoint` and `MultiPolygon` geometries into individual geometries. The SHP specification supports both of these via the `MultiPoint/MultiPointZ` shape and `Polygon/PolygonZ` with `parts` for each polygon/ring. This support was implemented in the exporter. I also consolidated and documented some of the logic that was duplicated between allocating space for a geometry and appending the geometry to the buffer.

The SHP parser did not implement support for `MultiPoint` shapes. This is now implemented so OpenSphere can import these geometries.

Resolves #433.

### DIFF
--- a/src/plugin/file/shp/shp.js
+++ b/src/plugin/file/shp/shp.js
@@ -1,6 +1,12 @@
 goog.declareModuleId('plugin.file.shp');
 
+const GeometryLayout = goog.require('ol.geom.GeometryLayout');
+const GeometryType = goog.require('ol.geom.GeometryType');
+
+const SimpleGeometry = goog.requireType('ol.geom.SimpleGeometry');
+
 /**
+ * SHP shape types.
  * @enum {number}
  */
 export const TYPE = {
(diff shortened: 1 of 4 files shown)


src/plugin/file/shp/shp.js
```diff
@@ -67,3 +73,123 @@ export const isSHPFileType = function(content) {
...
+      case GeometryType.POLYGON:
+        // Each part is a ring, all set.
+        partCoordinates = coordinates;
+        break;
+      case GeometryType.MULTI_POLYGON:
+        // Each part is a ring, flatten by one level.
+        partCoordinates = coordinates.flat(1);
+        break;
+      default:
+        // Unsupported geometry type.
```
jsalankey, line 142: I know we don't expect this to be hit, but it would probably be good to log a warning or an error just in case something weird happens.

jsalankey (approved):

src/plugin/file/shp/shp.js
schmidtk: For the purpose of this function I think it's fine not to log anything, but the caller may want to log if an empty array is returned.

PR #1401, opened 2021-11-11
fix: only one draw/measure interaction will be active at a time

### LINKED ISSUE #155: Draw Controls and Measure tool can both be toggled at the same time
Some might consider this to be a feature? 

A user can click on the draw controls and the measure tool and both appear to be toggled. If they then try to draw on the map the measure tool always goes first, then the draw tools. Not sure if anyone is using this 'feature' to cut down on trips back to the menu, but it seems possible.

ameletyan:
Fixes #155.

### DIFF
--- a/src/os/ui/draw/basedrawcontrols.js
+++ b/src/os/ui/draw/basedrawcontrols.js
@@ -365,11 +365,6 @@ export class Controller {
    * @export
    */
   activateControl(type) {
-    // Cancel any active measure interaction
-    if (this.measureInteraction && this.measureInteraction.getEnabled()) {
-      this.measureInteraction.setEnabled(false);
-    }
-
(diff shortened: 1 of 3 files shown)


src/os/ui/draw/basedrawcontrols.js
```diff
@@ -365,6 +365,11 @@ export class Controller {
    * @export
    */
   activateControl(type) {
+    // Cancel any active measure interaction
+    if (this.measureInteraction && this.measureInteraction.getEnabled()) {
```
schmidtk, line 369: `this.measureInteraction` isn't defined in the base class. The compiler should have caught this, but we should extend `activateControl` in `drawcontrols.js` for this behavior.

schmidtk (changes requested):

schmidtk (approved):

PR #1404, opened 2021-11-15
Auto import KML to Saved Places and hit detect labels in Saved Places

schmidtk:
This changes the following behavior when importing KML to Saved Places:

- KML's are now imported automatically to Saved Places. The UI that was being displayed didn't actually change anything with the import, and was not needed.
- The KML tree will now be preserved when importing to Saved Places.
- The Places tab in the Layers window now supports drag/drop to import, similar to the Areas/Filters tabs.

This also changes how hit detection works with Saved Places. Previously we disabled layer-level hit detection because the Always Show Labels option for each feature used the same field (`StyleField.SHOW_LABELS`) to show the label as the hit detection logic. That control now toggles a separate field, `RecordField.FORCE_SHOW_LABEL`, which enables feature-level label hit detection that precedes the layer-level detection. These labels will still be hit detected, but _only_ against other features with the flag set. This does require two passes of features in the label hit detection logic, but incurs minimal additional processing time. The end result is hit detection of labels in the Saved Places layer by default, and the ability to prioritize display of individual feature labels if needed.


src/os/command/feature/featureiconcmd.js
```diff
@@ -54,16 +53,4 @@ export default class FeatureIcon extends AbstractFeatureStyle {
 
     super.applyValue(configs, value);
   }
-
-  /**
-   * @inheritDoc
-   */
-  finish(configs) {
```
jsalankey, line 61: Do we need to dispatch some other event for icon changes? Just want to be sure the old event wasn't papering over some other need to notify an icon change.

schmidtk: The old comment was copy/pasted from the layer level command. This dispatch was only handled by `KMLNode`, but was lumped in with a color change which did more than needed. A color change needs to propagate the `icons` event up the tree so the node icon color will be updated. An icon change only needs to fire an event to trigger saving the KML (for Saved Places), which is what the style event is doing in the abstract command.

jsalankey (approved):

PR #1407, opened 2021-11-17
Fix issues with Go To

schmidtk:
This fixes/changes Go To behavior for Image, Tile, and VectorTile layers.

- Image layers were not setting the layer extent, so Go To would not work.
- Tile and VectorTile layers would not show Go To if the extent was 80% of the current projection's extent. This behavior has been removed, so the option will display as long as the extent is not empty.
- Arc layers were not handling the extent projection. The extent will now be parsed into an appropriate projection, if the extent's source projection is supported by OpenSphere.
- When zooming to a large extent, the 50% buffer that was applied would sometimes exceed the world extent, which would cause the 2D view to zoom a world over. This has been fixed.

Other Changes

- Fixed test JSON for Arc, which put `wkid` directly under `extent`. This should be under `spatialReference`, per the [specification](https://developers.arcgis.com/web-map-specification/objects/extent/).
- Fixed issues computing the extent for features with multiple geometries in EPSG:3857 (or non-degree projections). The `getThinnestExtent` function used a threshold that was based on degree units, which did not work well when the projection is in meters. That function now takes an optional epsilon value, and the `getFunctionalExtent` functions will now pass a projection-appropriate value.

### DIFF
--- a/src/plugin/arc/node/arcservicenode.js
+++ b/src/plugin/arc/node/arcservicenode.js
@@ -4,6 +4,7 @@ import ConfigDescriptor from '../../../os/data/configdescriptor.js';
 import DataManager from '../../../os/data/datamanager.js';
 import LayerType from '../../../os/layer/layertype.js';
 import Request from '../../../os/net/request.js';
+import {EPSG4326} from '../../../os/proj/proj.js';
 import ColorControlType from '../../../os/ui/colorcontroltype.js';
 import BaseProvider from '../../../os/ui/data/baseprovider.js';
 import DescriptorNode from '../../../os/ui/data/descriptornode.js';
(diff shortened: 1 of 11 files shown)


src/plugin/arc/node/arcservicenode.js
```diff
@@ -185,17 +186,16 @@ class ArcServiceNode extends LoadingNode {
    */
   addImageLayer_(json) {
     const id = this.server_.getId() + BaseProvider.ID_DELIMITER + /** @type {string} */ (json['name']);
-    const extent = /** @type {Object<string, number>} */ (json['extent']);
-    const wkid = /** @type {number} */ (extent['spatialReference']['latestWkid']);
+    const extent = arc.readEsriExtent(/** @type {Object} */ (json['extent']), EPSG4326);
```
jsalankey, line 189: Do we always intend to read the extent into 4326, or should we be using the current map projection? Does that get transformed later down the line?

schmidtk: This is called when the server is loaded, and the projection may change by the time the layer is loaded. I decided to always read the projection as 4326, and `ArcImageLayerConfig` will handle transforming it to the current projection. This is similar to the approach in `AbstractTileLayerConfig`, which assumes 4326 unless specified in `extentProjection`.

jsalankey (approved):

PR #1411, opened 2021-12-03
refactor(compare): abstracts flyto logic into static function

jsalankey:
Allows use of `flyTo` on any OL Map instance with our options. Adds plugin hook for the layer compare window to utilize this behavior to fly to a point with rotation (rather than just an extent).


src/os/ol/ol.js
```diff
@@ -3,5 +3,60 @@
...
+ * @type {number}
+ */
+export const FLY_ZOOM_DURATION = 1000;
+
+/**
+ *
+ * @param {!osx.map.FlyToOptions} options
+ * @param {OLMap} map
+ */
+export const flyTo = function(options, map) {
```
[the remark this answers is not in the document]
jsalankey: Yeah that would probably be better to just leave this for mixins. Looked around a bit for a better place and didn't love any of the options, but `os.map` makes more sense.

src/os/map/map.js
```diff
@@ -1,16 +1,19 @@
 goog.declareModuleId('os.map');
 
 import '../ol/ol.js';
+import FlightMode from '../map/flightmode.js';
```
schmidtk, line 4: We should look into a lint rule to check for the shortest path to a module. From this module, you can use `./flightmode.js`.

schmidtk (dismissed):
Commented on a module path nit that we should eventually add an eslint rule to handle, but otherwise this looks good.

src/os/map/map.js
jsalankey: Ahh, heh yeah I just copied it in from the other file and didn't even think about it. I'll clean this up.

schmidtk (approved):

PR #1413, opened 2021-12-06
feat: improved handling for duplicate alerts

ameletyan:


src/os/ui/alert/alertpopup.js
```diff
@@ -127,35 +101,33 @@ export class Controller {
...
-      } else {
-        this.resetDupes_();
+    var display = true;
+
+    if (this.lastEvent_ && event.getMessage() == this.lastEvent_.getMessage()) {
+      // See if a popup for the current alert is showing
+      var msg = event.getMessage();
+      var idx = '';
+      if (msg) {
+        idx = this['alertPopups'].findIndex((popupObj) => popupObj['msg'] == msg);
```
schmidtk, line 111: It looks like the index is only being used to get the item, so you might want to use `find` instead of `findIndex`, which will return the item directly if found.

src/os/alert/alertmanager.js
```diff
@@ -150,28 +157,34 @@ export default class AlertManager extends EventTarget {
...
-          break;
-        case AlertEventSeverity.WARNING:
-          log.warning(opt_logger, alert);
-          break;
-        case AlertEventSeverity.INFO:
-        case AlertEventSeverity.SUCCESS:
-        default:
-          log.info(opt_logger, alert);
-          break;
+    if (this.mostRecentAlert_ && this.mostRecentAlert_.getMessage() == alert) {
```
schmidtk, line 160: I don't think the alert manager needs to handle deduplication beyond what it's already doing with the throttle. The throttle consolidates alerts received within a short window (default 500ms) before they're even dispatched, and from there UI's can be responsible for how they handle further duplicates. Once an event is fired, it shouldn't be changed because UI's would have to know how to respond to that change. It's simpler if they only have to respond to new events and can assume every new event is for new messages.

src/os/alert/alertevent.js
```diff
@@ -58,6 +58,18 @@ export default class AlertEvent extends GoogEvent {
      */
     this.dismissDispatcher_ = opt_dismissDispatcher || null;
 
+    /**
+     * @type {?number}
+     * @private
+     */
+    this.handlerTimeoutId_ = null;
```
schmidtk, line 65: The handler/timeout are specific to the alert popup UI, so I don't think it makes sense to track them on the alert event. The event could be handled by any number of UI's, so its API should not include UI-specific properties. These can be tracked on the alert popup object instead.

src/os/ui/alert/alertpopup.js
```diff
@@ -127,35 +101,33 @@ export class Controller {
...
+      }
+      var alertPopup = this['alertPopups'][idx];
+
+      if (alertPopup) {
+        // Update the count of the popup
+        alertPopup.count = event.getCount();
```
schmidtk, line 117: This should use `alertPopup['count']`, because the property is quoted on the object. It's possible the compiler could rename the property when using dot notation. With the suggested change to not reuse events in the alert manager, this would also change to add the event count to the current value.

schmidtk (changes requested):

src/os/ui/alert/alertpopup.js
```diff
@@ -127,35 +101,33 @@ export class Controller {
...
+        alertPopup = this['alertPopups'].find((popupObj) => popupObj['msg'] == newMessage);
+      }
+
+      if (alertPopup) {
+        // Update the count of the popup
+        alertPopup['count'] += event.getCount();
+
+        // Reset the timeout on the dismiss handler so the popup remains
+        clearTimeout(alertPopup['handlerTimeoutId']);
+        setTimeout(alertPopup['handler'], Controller.ALERT_TIMER);
```
schmidtk, line 121: You'll want to update `handlerTimeoutId` from the return value of this `setTimeout` call to ensure the updated timeout can also be cleared.

src/os/alert/alertevent.js
```diff
@@ -114,4 +114,13 @@ export default class AlertEvent extends GoogEvent {
   getCount() {
     return this.count_;
   }
+
+  /**
+   * @param {?number|undefined} count
+   */
+  increaseCount(count) {
```
schmidtk, line 121: Can this be removed now?

test/os/alert/alertmanager.test.js
```diff
@@ -134,7 +134,7 @@ describe('os.alert.AlertManager', function() {
       const promises = [];
 
       for (const key in AlertEventSeverity) {
-        promises.push(am.sendAlert('alert', AlertEventSeverity[key], logger));
+        promises.push(am.sendAlert(`alert ${key}`, AlertEventSeverity[key], logger));
```
schmidtk, line 137: Does the original test work again, now that we aren't changing AlertManager?

schmidtk (changes requested):

src/os/ui/alert/alertpopup.js
```diff
@@ -127,35 +101,33 @@ export class Controller {
...
+    if (newMessage.length > 0 && newMessage == this.lastMessage_) {
+      // See if a popup for the current alert is showing
+      var alertPopup = null;
+      if (newMessage) {
+        alertPopup = this['alertPopups'].find((popupObj) => popupObj['msg'] == newMessage);
+      }
+
+      if (alertPopup) {
+        // Update the count of the popup
+        alertPopup['count'] += event.getCount();
```
schmidtk, line 117: Sometimes this count update isn't reflected in the UI because the Angular scope isn't applied if `displayAlert_` isn't called. We should call `apply(this.scope_);` at the end of this block to make sure the new count is shown in the UI.

schmidtk: This was a quick change, so I added a commit for it. All good now!

schmidtk (approved):

PR #1449, opened 2024-05-10
30 - containerized local development

sethahrenbach:
- remove `&` that backgrounds `npm run container-start-server` for containerized running
- use `0.0.0.0` instead of `localhost` for container execution
- Improve README with better instructions


package.json
```diff
@@ -144,7 +144,7 @@
     "package:update": "if git diff --name-only ORIG_HEAD HEAD | grep --quiet package.json; then echo 'UPDATE: package.json was updated, consider running yarn in your workspace root'; fi",
     "perms": "chmod -R u+rwX,go+rX,go-w .",
     "semantic-release": "semantic-release --dry-run",
-    "start-server": "light-server --serve ../../ --bind localhost --port 8282 --no-reload --open --quiet &",
+    "start-server": "light-server --serve ../../ --bind 0.0.0.0 --port 8282 --no-reload --open --quiet",
     "stop-server": "run-script-os",
```
bradh, line 148: I guess this means the server is exposed on all networks in a non-container environment. That seems like it should at least be explicit in the documentation.

sethahrenbach: I made a new script for running in a container and left the old one the same. good feedback

JoshNels (approved):

